PR #11, opened 2022-05-17
fix(token): remove double clone()ing from CPI calls

sohrab-:
_I'm not sure if you are accepting PRs from outside the project but I noticed this and I thought I raise it._

[`to_account_info()`](https://github.com/project-serum/anchor/blob/5194de04b1d2446730d42f75e875f3b2facca312/lang/src/lib.rs#L133) helpfully (?) performs a `clone()` already so the second cloning is not necessary.

`cargo test` runs fine but can't seem to get `anchor test` going. Rolled back to Anchor 0.20.1 and added some missing dependencies like `@orca-so/sdk` but no luck. I was hoping there would be a GitHub Action for it but doesn't look like it.

### DIFF
--- a/programs/whirlpool/src/util/token.rs
+++ b/programs/whirlpool/src/util/token.rs
@@ -14,11 +14,11 @@ pub fn transfer_from_owner_to_vault<'info>(
 ) -> Result<(), ProgramError> {
     token::transfer(
         CpiContext::new(
-            token_program.to_account_info().clone(),
+            token_program.to_account_info(),
             Transfer {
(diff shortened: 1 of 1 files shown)


odcheung (approved):
Confirmed on our setup that the unit tests and integration tests passes. Thank you for making this change!

odcheung:
Good call out on a better integration workflow here. I've created a task on our side to work on a better workflow to verify patches coming from other sources (Github Actions, like you suggested)

PR #12, opened 2022-05-17
Add token estimates for liquidity quotes, group token resolution

philcchen:
- Add tokenEst(imates) to the increase/decrease liquidity quotes
- Add group token resolution to minimize RPC calls when resolving ATAs

Token estimates - What we can expect the resulting token to be at this given time / on-chain data
Token min - The minimum expected amount of token received / sent given the estimate and slippage


odcheung (approved):
looks good to me!

PR #14, opened 2022-05-23
Fix incorrect fetch of mintA instead of mintB

everlastingsong:
* Fix incorrect fetch of mintA instead of mintB


odcheung (approved):
awesome, I'll merge this in. really appreciate your contribution!

PR #16, opened 2022-05-25
[v0.2.0] Improve increase/decrease liquidity functions for SDK

odcheung:
- Add support for auto-resolving ATA accounts prior for decrease liquidity ix
- Removed sourceWallet support for increase_liquidity ix
- Add separate funder/payer support for Whirlpool/Position increase/decrease liquidity functions
- Add tests to verify functions work as intended

Dependency on new version of common-sdk:
https://github.com/orca-so/orca-sdks/pull/6

NOTE:
- WhirlpoolClient API change. Pointing this to a 0.2.0 branch
- Noticed a limitation where sourceWallet and positionWallet has to be the same for increase_liquidity on the contract. Created a smart contract ticket for it and adjusted the Whirlpools SDK to reflect that

### DIFF
--- a/sdk/src/impl/position-impl.ts
+++ b/sdk/src/impl/position-impl.ts
@@ -45,8 +45,7 @@ export class PositionImpl implements Position {
   async increaseLiquidity(
     liquidityInput: IncreaseLiquidityInput,
     sourceWallet?: Address,
-    positionWallet?: Address,
-    payer?: PublicKey
+    positionWallet?: Address
   ) {
     const sourceWalletKey = sourceWallet
       ? AddressUtil.toPubKey(sourceWallet)
@@ -54,7 +53,6 @@ export class PositionImpl implements Position {
     const positionWalletKey = positionWallet
       ? AddressUtil.toPubKey(positionWallet)
       : this.ctx.wallet.publicKey;
-    const payerKey = payer ? payer : this.ctx.wallet.publicKey;
 
     const whirlpool = await this.fetcher.getPool(this.data.whirlpool, true);
     if (!whirlpool) {
@@ -96,7 +94,7 @@ export class PositionImpl implements Position {
     sourceWallet?: Address,
     positionWallet?: Address,
     resolveATA?: boolean,
-    payer?: PublicKey
+    ataPayer?: Address
   ) {
     const sourceWalletKey = sourceWallet
       ? AddressUtil.toPubKey(sourceWallet)
@@ -104,7 +102,7 @@ export class PositionImpl implements Position {
     const positionWalletKey = positionWallet
       ? AddressUtil.toPubKey(positionWallet)
       : this.ctx.wallet.publicKey;
-    const payerKey = payer ? payer : this.ctx.wallet.publicKey;
+    const ataPayerKey = ataPayer ? AddressUtil.toPubKey(ataPayer) : this.ctx.wallet.publicKey;
     const whirlpool = await this.fetcher.getPool(this.data.whirlpool, true);
 
     if (!whirlpool) {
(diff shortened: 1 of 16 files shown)


sdk/src/impl/position-impl.ts
```diff
@@ -93,30 +112,41 @@ export class PositionImpl implements Position {
...
-        ).publicKey,
-        tickArrayUpper: PDAUtil.getTickArray(
-          this.ctx.program.programId,
-          this.data.whirlpool,
-          TickUtil.getStartTickIndex(this.data.tickUpperIndex, whirlpool.tickSpacing)
-        ).publicKey,
-        positionAuthority: positionWalletKey,
-      })
+    const txBuilder = new TransactionBuilder(this.ctx.provider);
+    const [ataA, ataB] = await resolveOrCreateATAs(
```
0xscuba, line 116: i think one issue with using `resolveOrCreateATAs` is that it makes a network call, and thus makes it harder for clients who want to handle this on their own in a batched manner more difficult

odcheung: Makes sense. Thanks for flagging this. I was talking to @philcchen the other day and I believe after he continues working on transaction packing, he'll work on allowing `resolveOrCreateATAs` to take in an ATA cache. So once that's in, we can expose the cache as a parameter for these types of `WhirlpoolClient` calls.

As of right now, WhirlpoolClient is mainly targeted for simpler projects. From some comments on integration i think it makes more sense to include this for now then add optionality to skip later.  So I suggest we make the trade off here to add the resolve tx while we follow up with the ataMap down the road. Meanwhile, more advanced use cases will just construct their own `decreaseLiquidityIx`

What do you guys think?

philcchen: I'm personally okay with the trade-off for now, and revisiting the batching of calls in the future;

That said, I don't think it would be too difficult to take in a boolean on whether to optionally add the create ATA instructions as a compromise in-between now and then?

odcheung: I've adjusted the methods to allow for optional ATA resolve for decrease-liq only (increase-liq doesn't make sense since you have to have the tokens ready prior to calling).

I also took the chance to fix up the public APIs to support different funder/payer wallet, sourceWallet, destination Wallet for these methods in the following fixes.

sdk/src/impl/position-impl.ts
```diff
@@ -40,87 +45,118 @@ export class PositionImpl implements Position {
...
+    positionWallet?: Address,
+    payer?: PublicKey
   ) {
     const sourceWalletKey = sourceWallet
       ? AddressUtil.toPubKey(sourceWallet)
       : this.ctx.wallet.publicKey;
     const positionWalletKey = positionWallet
       ? AddressUtil.toPubKey(positionWallet)
       : this.ctx.wallet.publicKey;
+    const payerKey = payer ? payer : this.ctx.wallet.publicKey;
```
philcchen, line 57: nit: can use the null coaelsecing operating, `payer ?? this.ctx.wallet.publicKey`

sdk/src/impl/position-impl.ts
```diff
@@ -40,87 +45,118 @@ export class PositionImpl implements Position {
   async increaseLiquidity(
     liquidityInput: IncreaseLiquidityInput,
     sourceWallet?: Address,
-    positionWallet?: Address
+    positionWallet?: Address,
+    payer?: PublicKey
```
philcchen, line 49: nit: I think this might be a little ambiguous here, since it's only used for paying for the ATAs and not for the any other parts of the transaction?

philcchen (approved):

sdk/src/impl/position-impl.ts
odcheung: ah. good catch. increase_liquidity has no use for payer. I forgot to remove it when i removed resolveATA from increase-liq

odcheung: For decrease-liquidity, i've renamed it to ataPayer for better clarity

PR #17, opened 2022-05-25
Add optional refresh parameter for whirlpoolclient

odcheung:
- Add an option to refresh/not-refresh accounts for WhirlpoolClient


0xscuba (approved):
lgtm

PR #19, opened 2022-05-26
Fix build errors for increase/decrease liquidity Ix

odcheung:
- Previous change to Increase/DecreaseLiquidityInput broke the Increase/DecreaseLiquidityParams
- Add new Increase/DecreaseLiquidityQuote type to add a layer of separation between Params and Quote


philcchen (approved):
👍  I waffled a bit on whether or not I wanted to separate out the tokenEst since it isn't strictly necessary for the creation of the quotes, but since the return of the quote included the tokenEst, I didn't subtype. don't have strong opinions here

PR #20, opened 2022-05-26
[v0.2.0] Refactor swapQuote functions

odcheung:
Previous swapQuote function is pretty difficult to use. Making some modifications to help users call swap quotes. This is a breaking change to the swapQuote's parameters. When we release, we have to bump to v0.2.0

- `swapQuoteByInputToken` as a convenience method to help users fetch the necessary information to perform the swap quote
- `swapQuoteWithParams` for the more advanced users who have previously fetched the data they need to perform the swap quote

### DIFF
--- a/sdk/src/quotes/public/swap-quote.ts
+++ b/sdk/src/quotes/public/swap-quote.ts
@@ -22,15 +22,16 @@ import {
 } from "../../types/public";
 import { AddressUtil, MathUtil, Percentage, ZERO } from "@orca-so/common-sdk";
 import { PriceMath, TickArrayUtil, TickUtil } from "../../utils/public";
+import { Whirlpool } from "../../whirlpool-client";
+import { AccountFetcher } from "../../network/public";
 
 /**
  * @category Quotes
  */
 export type SwapQuoteParam = {
-  whirlpoolAddress: Address;
-  swapTokenMint: Address;
   whirlpoolData: WhirlpoolData;
   tokenAmount: u64;
+  aToB: boolean;
   amountSpecifiedIsInput: boolean;
   slippageTolerance: Percentage;
   tickArrayAddresses: PublicKey[];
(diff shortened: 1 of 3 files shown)


sdk/src/quotes/public/swap-quote.ts
```diff
@@ -45,6 +46,62 @@ export type SwapQuote = {
...
+    swapMintKey.equals(whirlpoolData.tokenMintA) === amountSpecifiedIsInput ? true : false;
+
+  const tickArrayAddresses = PoolUtil.getTickArrayPublicKeysForSwap(
+    whirlpoolData.tickCurrentIndex,
+    whirlpoolData.tickSpacing,
+    aToB,
+    AddressUtil.toPubKey(programId),
+    whirlpool.getAddress()
+  );
+  const tickArrays = await fetcher.listTickArrays(tickArrayAddresses, true);
```
philcchen, line 75: if this is the publicly exposed method, it might be good to have the refresh here be optional?

sdk/src/quotes/public/swap-quote.ts
```diff
@@ -45,6 +46,62 @@ export type SwapQuote = {
...
+
+  // Check if all the tick arrays have been initialized.
+  let emptyArrayIndex = -1;
+  const arrayNeedInit = tickArrays.some((value, index) => {
```
philcchen, line 79: nit: would be better to return all arrays that are uninitialized instead of just the first one?

odcheung: yeah good idea. i've updated the code to throw an error on all the offending addresses

sdk/src/quotes/public/swap-quote.ts
```diff
@@ -45,6 +46,56 @@ export type SwapQuote = {
...
+    aToB,
+    AddressUtil.toPubKey(programId),
+    whirlpool.getAddress()
+  );
+  const tickArrays = await fetcher.listTickArrays(tickArrayAddresses, refresh);
+
+  // Check if all the tick arrays have been initialized.
+  const uninitializedIndices = TickArrayUtil.getUninitializedArrays(tickArrays);
+  if (uninitializedIndices.length > 0) {
+    const uninitializedArrays = uninitializedIndices.reduce<string>(
```
philcchen, line 81: nit: I think it could be cleaner to do a map then join?

```
unitializedIndices.map(index => tickArrayAddresses[unInitIndex].toBase58()).join(" ");
```

philcchen (approved):

sdk/src/quotes/public/swap-quote.ts
odcheung: you learn something new everyday!

PR #21, opened 2022-06-03
Adding toBaseQuoteOrder method to pool-utils

terranmoccasin:
**Background**
The Whirlpools smart contract enforces a strict ordering of pool token mints for `token_a` and `token_b` to guarantee unique pools. This ordering does not represent a `base / quote` currency pair.

**Changes**
This PR adds a helper method to determine a `base / quote` currency ordering that prioritizes USD stable coins followed by SOL and its variants.

### DIFF
--- a/sdk/src/utils/public/pool-utils.ts
+++ b/sdk/src/utils/public/pool-utils.ts
@@ -209,6 +209,45 @@ export function toTokenAmount(a: number, b: number): TokenAmounts {
   };
 }
 
+/**
+ * Given an arbitrary pair of token mints, this function returns an ordering of the token mints
+ * in the format [base, quote]. USD based stable coins are prioritized as the quote currency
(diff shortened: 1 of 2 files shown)


sdk/src/utils/public/pool-utils.ts
```diff
@@ -209,6 +209,45 @@ export function toTokenAmount(a: number, b: number): TokenAmounts {
   };
 }
 
+/**
+ * Given an arbitrary pair of token mints, this function returns an ordering of the token mints
+ * in the format [base, quote]. USD based stable coins are prioritized as the quote currency
+ * followed by variants of SOL.
+ */
+export function toBaseQuoteOrder(
```
odcheung, line 217: Can you move this into the PoolUtil namespace?

PR #25, opened 2022-06-09
[v0.2.0] Refactor swapQuote functions (#20)

odcheung:
- Refactor swapQuoteByInputToken as a convenience method to help users fetch the necessary information to perform the swap quote
- Add swapQuoteWithParams for the more advanced users who have previously fetched the data they need to perform the swap quote


odcheung:
same commit as https://github.com/orca-so/whirlpools/pull/20. push commit to main

PR #26, opened 2022-06-13
Add PoolUtil swap direction & token type util functions

odcheung:
- Add util functions to PoolUtil to help identify swap direction & token A/B for token mints

### DIFF
--- a/sdk/src/utils/public/pool-utils.ts
+++ b/sdk/src/utils/public/pool-utils.ts
@@ -1,12 +1,17 @@
 import { AddressUtil, MathUtil, Percentage } from "@orca-so/common-sdk";
 import { Address, BN } from "@project-serum/anchor";
-import { u64 } from "@solana/spl-token";
+import { Token, u64 } from "@solana/spl-token";
 import { PublicKey } from "@solana/web3.js";
 import Decimal from "decimal.js";
-import { MAX_TICK_ARRAY_CROSSINGS, WhirlpoolRewardInfoData } from "../../types/public";
+import {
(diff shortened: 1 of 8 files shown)


sdk/src/utils/public/types.ts
```diff
@@ -0,0 +1,9 @@
+export enum SwapDirection {
+  AtoB = "aToB",
+  BtoA = "bToA",
+}
+
+export enum TokenType {
+  TokenA = 1,
+  TokenB,
```
terranmoccasin, line 8: Shoudl this be `TokenB = 2,` ?

sdk/src/utils/public/pool-utils.ts
```diff
@@ -20,6 +25,44 @@ export class PoolUtil {
...
   }
 
+  /**
+   * Return the corresponding token type (TokenA/B) for this mint key for a Whirlpool.
+   *
+   * @param pool The Whirlpool to evaluate the mint against
+   * @param mint The token mint PublicKey
+   * @returns The match result in the form of TokenType enum. undefined if the token mint is not part of the trade pair of the pool.
+   */
+  public static getTokenType(pool: WhirlpoolData, mint: PublicKey) {
```
terranmoccasin, line 35: nit: add method return types for `getTokenType` and `getSwapDirection`

terranmoccasin (approved):
lgtm - just had two small comments

sdk/src/utils/public/types.ts
odcheung: Typescript will automatically fill that in if you specify the starting number

PR #32, opened 2022-06-29
Change param ordering for Position's resolveATA 

odcheung:
- Change parameter ordering for resolveATA as it is the most important parameter
- Default resolveATA to true. Power users who do not want to ping RPC constantly can manually turn it off
- Fix up some comments


terranmoccasin (approved):
:shipit:

PR #36, opened 2022-07-11
Fix incorrect TickArrayPublicKeysForSwap generation at WhirlpoolImpl

everlastingsong:
- add test for swap method of WhirlpoolImpl
- remove getTickArrayPublicKeysForSwap method from WhirlpoolImpl
- use tickarrays passed as SwapQuote

getTickArrayPublicKeysForSwap method of WhirlpoolImpl generate incorrect public keys for tickarrays when tickCurrentIndex is multiple of tickSpacing * TICK_ARRAY_SIZE(88).
Because it uses TickUtil.getPrevInitializableTickIndex, it generate public key of tickarray which doesn't contain tickCurrentIndex as first tickarray and this occurs InvalidTickArraySequence(0x1787) error.

The quote parameter contains tickarray0, tickarray1 and tickarray2, so this fix simply use them.

### DIFF
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -376,6 +376,9 @@ export class WhirlpoolImpl implements Whirlpool {
       estimatedAmountOut,
       aToB,
       amountSpecifiedIsInput,
+      tickArray0,
+      tickArray1,
+      tickArray2,
     } = quote;
     const whirlpool = this.data;
     const txBuilder = new TransactionBuilder(this.ctx.provider);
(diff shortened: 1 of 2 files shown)


odcheung:
hey @everlastingsong, good catch on this one. 
We followed up on your call out in discord and we did a big revamp to the swap-quote. Patch - https://github.com/orca-so/whirlpools/pull/30
That fix will also do what you are doing here and with more bug fixes related to traversal. We are aiming to get it out in the next few days.

everlastingsong:
Hi @odcheung, thanks for your review and comment.
I see. [SDK] Refactoring swap-quote #30 is great!

PR #42, opened 2022-08-06
Implement batch getPools method

terranmoccasin:
Implementing a `getPools` method on the WhirlpoolClient for batch fetching pools.

This change also adds fetching onchain data for token accounts for vaults (including reward vaults).

These fields were provided by the old SDK, and it feels like they also belong in the new SDK rather than interacting with the web3.js client after retrieving whirlpools.

### DIFF
--- a/sdk/src/impl/whirlpool-client-impl.ts
+++ b/sdk/src/impl/whirlpool-client-impl.ts
@@ -1,17 +1,12 @@
 import { AddressUtil } from "@orca-so/common-sdk";
 import { Address } from "@project-serum/anchor";
-import { BN } from "bn.js";
 import { WhirlpoolContext } from "../context";
 import { AccountFetcher } from "../network/public";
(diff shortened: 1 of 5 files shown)


sdk/src/impl/whirlpool-client-impl.ts
```diff
@@ -23,17 +30,69 @@ export class WhirlpoolClientImpl implements WhirlpoolClient {
     if (!account) {
       throw new Error(`Unable to fetch Whirlpool at address at ${poolAddress}`);
     }
-    const tokenInfos = await getTokenInfos(this.ctx.fetcher, account, false);
```
terranmoccasin, line 26: I noticed this `refresh` was set to false originally. Since parts of the token mint info can change on subsequent requests, I changed the value to use the `refresh` param.

Let me know if there's something I'm missing here.

PR #43, opened 2022-08-25
Bug Fix: Wrong arg order for method call

coudron:
I'm experiencing a bug with `decreaseLiquidityQuoteByLiquidity` is returning the incorrect value.

Local testing points to this possibly being the bug.


odcheung:
hey @coudron, thanks for submitting this.

we have been seeing discrepancies between the SDK quote value and the actual result returned. We have that tracking internally and will address it soon.

This patch won't fix it because the getTokenAFromLiquidity method always reorders the tokens prior to computing them, so changing the order here wouldn't yield a different result.

PR #52, opened 2022-11-16
Add getPositions function to WhirlpoolClient

odcheung:
- Provide a new method to fetch many positions in a single RPC call

### Testing
- On my test branch in Orca UI, verified that this method will return all positions


yugure-orca (approved):
@odcheung 
could you correct the comment in sdk/src/whirlpool-client.ts that is off to the left by 1 space?

matt-allan (approved):

odcheung:
> @odcheung could you correct the comment in sdk/src/whirlpool-client.ts that is off to the left by 1 space?

done!

PR #53, opened 2022-11-16
[SDK] Add `WhirlpoolClient.createPool()`

atamari-orca:
* This PR adds a `createPool` function to the `WhirlpoolClient` to support whirlpool creation.
* Also adds `TickUtil.invertTick()`, `PriceMath.invertPrice`, and `PriceMath.invertSqrtPriceX64`.

### DIFF
--- a/sdk/tests/sdk/whirlpools/whirlpool-client-impl.test.ts
+++ b/sdk/tests/sdk/whirlpools/whirlpool-client-impl.test.ts
@@ -110,4 +110,25 @@ describe("whirlpool-client-impl", () => {
     assert.ok(tickArrayAccountAfter.ticks.length > 0);
     assert.ok(tickArrayAccountAfter.whirlpool.equals(expectedPda.publicKey));
   });
+
+  it("throws an error when token order is incorrect", async () => {
+    const initalTick = TickUtil.getInitializableTickIndex(
+      PriceMath.sqrtPriceX64ToTickIndex(poolInitInfo.initSqrtPrice),
+      poolInitInfo.tickSpacing
+    );
+
+    const invInitialTick = TickUtil.invertTick(initalTick);
+
+    await assert.rejects(
+      client.createPool(
+        poolInitInfo.whirlpoolsConfig,
+        poolInitInfo.tokenMintB,
+        poolInitInfo.tokenMintA,
+        poolInitInfo.tickSpacing,
+        invInitialTick,
+        funderKeypair.publicKey
+      ),
+      /Token order needs to be flipped to match the canonical ordering \(i.e. sorted on the byte repr. of the mint pubkeys\)/
+    );
+  });
 });


sdk/src/whirlpool-client.ts
```diff
@@ -51,6 +52,25 @@ export interface WhirlpoolClient {
...
+
+  /**
+   * Create a Whirlpool account for a group of token A, token B and tick spacing
+   * @param whirlpoolConfig the address of the whirlpool config
+   * @param tokenMintA the address of the token A
+   * @param tokenMintB the address of the token B
+   * @param tickSpacing the space between two ticks in the tick array
+   * @param initialTick the initial tick that the pool is set to (derived from initial price)
+   * @param funder the account to debit SOL from to fund the creation of the account(s)
+   * @return `pubkey`: The public key of the newly created whirlpool account. `tx`: The transaction containing instructions for the on-chain operations.
```
odcheung, line 64: nit: returns poolKey instead of pubkey for more context

sdk/src/impl/whirlpool-client-impl.ts
```diff
@@ -100,4 +104,87 @@ export class WhirlpoolClientImpl implements WhirlpoolClient {
...
+  ): Promise<{ pubkey: PublicKey; tx: TransactionBuilder }> {
+    invariant(TickUtil.checkTickInBounds(initialTick), "initialTick is out of bounds.");
+    invariant(
+      TickUtil.isTickInitializable(initialTick, tickSpacing),
+      `initial tick ${initialTick} is not an initializable tick for tick-spacing ${tickSpacing}`
+    );
+
+    [tokenMintA, tokenMintB] = PoolUtil.orderMints(tokenMintA, tokenMintB).map(
+      (addr) => new PublicKey(addr)
+    );
```
odcheung, line 125: if the tokenMint order is switched here, we would have to invert the initialTick..

odcheung (commented):
lgtm! we need to handle the inverted case for the token order as well. then i think we are good to go!

sdk/src/impl/whirlpool-client-impl.ts
odcheung: true. agree on it failing so that the caller is aware that the price inverted.
As part of this change, can we also
1. provide a util function so callers can easily invert the price/tick
2. add documentation for this method on this failure case. ex. @throws error for inverted case and point them towards the invert price util fn.

To invert, I believe you simply negate the tick value. @philcchen please double check me on this.
> let i = tick
> P(i) = 1.0001^i
> Pb/Pa = 1.0001^i
> 1/ (Pb/Pa) = 1/ (1.0001^i)
> Pa/Pb = 1.0001^-i

sdk/src/impl/whirlpool-client-impl.ts
```diff
@@ -101,22 +105,116 @@ export class WhirlpoolClientImpl implements WhirlpoolClient {
     );
   }
 
-  public async getPositions(positionAddresses: Address[], refresh = false): Promise<Record<string, Position | null>> {
+  public async getPositions(
+    positionAddresses: Address[],
+    refresh = false
+  ): Promise<Record<string, Position | null>> {
```
odcheung, line 111: hm. i think we are missing a pre-commit hook on this repo's git to do the yarn prettier

sdk/tests/utils/init-utils.ts
```diff
@@ -7,19 +7,30 @@ import {
   createMint,
   mintToByAuthority,
   TickSpacing,
-  ZERO_BN
+  ZERO_BN,
```
odcheung, line 10: did we change anything in this file other than formatting?

sdk/tests/sdk/whirlpools/whirlpool-client-impl.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        TickSpacing.Standard,
+        3000,
+        PriceMath.priceToSqrtPriceX64(new Decimal(100), 6, 6),
+        funderKeypair.publicKey,
+        false
+      )
+    ).poolInitInfo;
+  });
+
+  it("successfully creates a new whirpool account and initial tick array account", async () => {
```
odcheung, line 39: can we also add test cases for the failure cases when token order requires inversion?

odcheung (approved):
approve but would like to add 1 test for the invert case

PR #54, opened 2022-11-17
[SDK] Add `Position.collectFees()`

atamari-orca:
## Description
Add a `collectFees()` wrapper method to the `Position` type

## Blocked TODOs:
- [x] @odcheung please answer the question here as well as on discord

## Pre-merge TODOs:
- [ ] Resolve in-line TODOs
- [ ] Add tests
- [ ] Add JSDoc

### DIFF
--- a/sdk/src/impl/position-impl.ts
+++ b/sdk/src/impl/position-impl.ts
@@ -11,6 +11,7 @@ import {
   DecreaseLiquidityInput,
   increaseLiquidityIx,
   decreaseLiquidityIx,
+  collectFeesIx,
 } from "../instructions";
 import { PositionData } from "../types/public";
 import { Position } from "../whirlpool-client";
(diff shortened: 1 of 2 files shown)


sdk/src/impl/position-impl.ts
```diff
@@ -189,6 +190,75 @@ export class PositionImpl implements Position {
...
+      ataPayer ?? this.ctx.wallet.publicKey,
+    ]);
+
+    const whirlpool = await this.fetcher.getPool(this.data.whirlpool, true);
+    if (!whirlpool) {
+      throw new Error("Unable to fetch whirlpool for this position.");
+    }
+
+    // Question @meep: Can either fee_owed_a or fee_owed_b be 0 while the other is non-zero?
+    //                 If so, will the collect_fees IX fail since it just blindly calls a token transfer CPI?
```
odcheung, line 211: collectfee/rewards won't fail if there's no fee to collect. It'll just be a no-op.

> Can either fee_owed_a or fee_owed_b be 0 while the other is non-zero?

Yes! Fee comes from the swap-input-token, so if a pool is newly created and it only got swapped in 1 direction, you would end up with that scenario. But as explained above, it's not something the client would have to explicitly handle.

sdk/src/impl/position-impl.ts
```diff
@@ -189,6 +190,75 @@ export class PositionImpl implements Position {
...
+
+    // TODO(atamari): Resolve the following questions:
+    // 1. Question mentioned above
+    // 2. Question asked on discord re: stale on-chain values of fee_owed_a and fee_owed_b preventing valid fee collection
```
odcheung, line 215: for the most updated value, client would have to call `updateFeesAndRewardsIx` to update reward/fee values on-chain.

If not, the collect call would only fetch the outstanding amount in the account. Any 'technically owed but not updated' amount will be reflected the next time the user adjusts liquidity or call updateFee/rewards.

sdk/src/impl/position-impl.ts
```diff
@@ -189,6 +190,75 @@ export class PositionImpl implements Position {
...
+    positionWallet?: Address,
+    ataPayer?: Address
+  ): Promise<TransactionBuilder> {
+    const [destinationWalletKey, positionWalletKey, ataPayerKey] = AddressUtil.toPubKeys([
+      destinationWallet ?? this.ctx.wallet.publicKey,
+      positionWallet ?? this.ctx.wallet.publicKey,
+      ataPayer ?? this.ctx.wallet.publicKey,
+    ]);
+
+    const whirlpool = await this.fetcher.getPool(this.data.whirlpool, true);
```
odcheung, line 205: i think refresh should be false here. We only care about the tokenMint, which will never change. If the cache already has a hit on this whirlpool account, we should leverage it.

PR #55, opened 2022-11-19
fix: strict initialized checking at swapQuoteWithParams

yugure-orca:
## related issue
https://app.asana.com/0/1202413563699470/1203257944757025/f

## changes
- removed checkIfAllTickArraysInitialized from swapQuoteWithParams
- added check to ensure that all TickArrays are initialized at getSwapTx
- In a TickArraySequence, uninitialized TickArrays are truncated.
- fix sqrtPriceLimit range check
- fix commitment issue in test code
- remove test "swap with a manual quote with dev-fee of 200%" (the code does not throw the exception)

## note
passed all tests (integration & sdk)
- needed to remove yarn.lock and increase solana/web3.js version to get TokenUtil.createSendTokensToWalletInstruction working (BigInt related)
- In jest, "before" needed to be "beforeAll".


sdk/src/quotes/swap/tick-array-sequence.ts
```diff
@@ -3,32 +3,48 @@ import {
...
     readonly aToB: boolean
   ) {
     if (!tickArrays[0] || !tickArrays[0].data) {
       throw new Error("TickArray index 0 must be initialized");
     }
 
-    this.touchedArrays = [...Array<boolean>(tickArrays.length).fill(false)];
+    // If an uninitialized TickArray appears, truncate all TickArrays after it (inclusive).
+    this.tickArrays = [];
+    for (let i=0; i < tickArrays.length && !!tickArrays[i] && !!tickArrays[i].data; i++) {
```
odcheung, line 38: style nit ideas to get rid of the ! hard cast:

1. do the null check in the for loop closure instead of the for loop conditions. early break when we detect a null (most straightforward)
2. keep the current for loop but do the null check again so typescript recognizes it's not null. 
3. make a copy of input tickArray, do a reduce. splice the array when we get to a null to early exit a reduce (peak typescript ridiculous-ness)

sdk/tests/sdk/whirlpools/whirlpool-impl.test.ts
```diff
@@ -23,7 +23,10 @@ import { initTestPool } from "../../utils/init-utils";
 import { mintTokensToTestAccount } from "../../utils/test-builders";
 
 describe("whirlpool-impl", () => {
-  const provider = anchor.AnchorProvider.local();
+  // The default commitment of AnchorProvider is "processed".
+  // But commitment of some Token operations is based on “confirmed”, and preflight simulation sometimes fail.
+  // So use "confirmed" consistently.
+  const provider = anchor.AnchorProvider.local(undefined, {commitment: "confirmed", preflightCommitment: "confirmed"});
```
odcheung, line 29: ah nice. was it only causing issues for this set of tests? should we make the change to all of the other ones?

odcheung (commented):
thanks for making this change! just some small nits

sdk/src/quotes/swap/tick-array-sequence.ts
yugure-orca: @odcheung 
Thanks for the comment & ideas.🙂
I have refactored the code using the 1st straightforward way.

odcheung (approved):

PR #56, opened 2022-11-22
Fix build to unblock yarn link in dev environment

odcheung:
- Changing to ES6 build so webpack 5 can resolve paths correctly if this package is yarn linked
- Add `process` to this build as webpack 5 cannot resolve polyfills correctly when yarn linked.
- Remove forced resolution of web3.js

## Testing
- Did a yarn link on my local environment. Errors are gone.


matt-allan (approved):
nice :+1:

PR #58, opened 2022-11-23
[SDK] Close Position, Collect Fees, Collect Rewards

atamari-orca:
## Description
This PR adds support for the following operations:
- `Whirlpool.closePosition()`
- `Position.collectRewards()`
- `Position.collectFees()`

## Reviewer TODOs:
- [x] @odcheung have some questions inline for you marked as `TODO(meep)`

## Pre Merge TODOs:
- [x] Integration tests for Close Position
- [x] If max TX size scenario for close position fails in the tests, split it into multiple TXs (maybe use the dynamic TX splitting logic @odcheung suggested)
- [x] Perform all ATA checking and creation outside of the collectFees, collectRewards functions in close position and just pass in false to the inner functions (Waiting on @odcheung to finalize the function to use)

### DIFF
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -30,6 +30,7 @@ import {
   decreaseLiquidityQuoteByLiquidityWithParams,
 } from "../quotes/public";
 import { TokenAccountInfo, TokenInfo, WhirlpoolData, WhirlpoolRewardInfo } from "../types/public";
+import { getTickArrayDataForPosition } from "../utils/builder/position-builder-util";
 import { PDAUtil, TickArrayUtil, TickUtil } from "../utils/public";
 import { getTokenMintsFromWhirlpools, resolveAtaForMints } from "../utils/whirlpool-ata-utils";
 import { Whirlpool } from "../whirlpool-client";
@@ -355,6 +356,17 @@ export class WhirlpoolImpl implements Whirlpool {
       this.ctx.provider.wallet
     );
 
+    const affiliatedMints = getTokenMintsFromWhirlpools([whirlpool]);
+    const accountExemption = await this.ctx.fetcher.getAccountRentExempt();
+    const { ataTokenAddresses, resolveAtaIxs } = await resolveAtaForMints(this.ctx, {
+      mints: affiliatedMints,
+      accountExemption,
+      receiver: destinationWallet,
+      payer: payerKey,
+    });
+
+    ataTxBuilder.addInstructions(resolveAtaIxs);
+
     const txBuilder = new TransactionBuilder(
       this.ctx.provider.connection,
       this.ctx.provider.wallet
@@ -374,21 +386,11 @@ export class WhirlpoolImpl implements Whirlpool {
       this.ctx.program.programId
     ).publicKey;
 
-    const affiliatedMints = getTokenMintsFromWhirlpools([whirlpool]);
-    const accountExemption = await this.ctx.fetcher.getAccountRentExempt();
-    const { ataTokenAddresses, resolveAtaIxs } = await resolveAtaForMints(this.ctx, {
-      mints: affiliatedMints,
-      accountExemption,
-      receiver: destinationWallet,
-      payer: payerKey,
-    });
-
-    ataTxBuilder.addInstructions(resolveAtaIxs);
-
-    const tickArrayLowerData = await this.ctx.fetcher.getTickArray(tickArrayLower, true);
-    const tickArrayUpperData = await this.ctx.fetcher.getTickArray(
-      tickArrayUpper,
-      !tickArrayUpper.equals(tickArrayLower)
+    const [tickArrayLowerData, tickArrayUpperData] = await getTickArrayDataForPosition(
+      this.ctx,
+      positionData,
+      whirlpool,
+      true
     );
 
     invariant(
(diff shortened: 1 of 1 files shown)


sdk/src/impl/position-impl.ts
```diff
@@ -183,10 +188,175 @@ export class PositionImpl implements Position {
     return txBuilder;
   }
 
+  async collectFees(
+    resolveATA: boolean = true,
+    updateFeesAndRewards: boolean = true,
+    destinationWallet?: Address,
+    positionWallet?: Address,
+    ataPayer?: Address
```
odcheung, line 196: need refresh flag (explanation below)

sdk/src/whirlpool-client.ts
```diff
@@ -335,5 +335,45 @@ export interface Position {
...
+   * @param rewardsToCollect - reward mints to collect (omitting this parameter means all rewards will be collected)
+   * @return the transaction that will collect fees from the position
+   */
+  collectRewards: (
+    resolveATA?: boolean,
+    updateFeesAndRewards?: boolean,
+    destinationWallet?: Address,
+    positionWallet?: Address,
+    ataPayer?: Address,
+    rewardsToCollect?: Address[]
```
odcheung, line 377: rewardsToCollect should come first, then updateFeeAndRewards, then resolveATA, then the other optional

sdk/src/whirlpool-client.ts
```diff
@@ -335,5 +335,45 @@ export interface Position {
...
+   * @param positionWallet - the wallet to that houses the position token. If null, the WhirlpoolContext wallet is used.
+   * @param ataPayer - wallet that will fund the creation of the new associated token accounts
+   * @return the transaction that will collect fees from the position
+   */
+  collectFees: (
+    resolveATA?: boolean,
+    updateFeesAndRewards?: boolean,
+    destinationWallet?: Address,
+    positionWallet?: Address,
+    ataPayer?: Address
```
odcheung, line 355: updateFeesAndRewards should come first as it's more relevant to the topic of collectFees. the rest are more standard

sdk/src/impl/position-impl.ts
```diff
@@ -183,10 +188,175 @@ export class PositionImpl implements Position {
...
+    ataPayer?: Address,
+    rewardsToCollect?: Address[]
+  ): Promise<TransactionBuilder> {
+    const [destinationWalletKey, positionWalletKey, ataPayerKey] = AddressUtil.toPubKeys([
+      destinationWallet ?? this.ctx.wallet.publicKey,
+      positionWallet ?? this.ctx.wallet.publicKey,
+      ataPayer ?? this.ctx.wallet.publicKey,
+    ]);
+
+    const whirlpool = await this.ctx.fetcher.getPool(this.data.whirlpool);
```
odcheung, line 265: While it's quite unlikely whirlpool reward initialize state changes that often, the user might want to always fetch the latest whirlpool stats here. We can add a refresh flag?

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -328,65 +335,117 @@ export class WhirlpoolImpl implements Whirlpool {
...
 
-    const txBuilder = new TransactionBuilder(
-      this.ctx.provider.connection,
-      this.ctx.provider.wallet
+    ataTxBuilder.addInstructions(resolveAtaIxs);
+
+    const tickArrayLowerData = await this.ctx.fetcher.getTickArray(tickArrayLower, true);
+    const tickArrayUpperData = await this.ctx.fetcher.getTickArray(
+      tickArrayUpper,
+      !tickArrayUpper.equals(tickArrayLower)
```
odcheung, line 391: `once` https://github.com/orca-so/whirlpools/pull/63 goes in, you should be able to skip all this and fetch the whirlpoolData & tick array data from the PositionImpl object

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -422,9 +519,13 @@ export class WhirlpoolImpl implements Whirlpool {
       position: positionAddress,
       positionMint: position.positionMint,
     });
+
     txBuilder.addInstruction(positionIx);
 
-    return txBuilder;
+    return {
+      ataTx: ataTxBuilder.compressIx(false).instructions.length > 0 ? ataTxBuilder : undefined,
```
odcheung, line 526: i added a txBuilder.isEmpty() method to do the same check.

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -328,65 +335,117 @@ export class WhirlpoolImpl implements Whirlpool {
     destinationWallet: PublicKey,
     positionWallet: PublicKey,
     payerKey: PublicKey
-  ): Promise<TransactionBuilder> {
+  ): Promise<{ ataTx?: TransactionBuilder; closeTx: TransactionBuilder }> {
```
odcheung, line 338: Can we return TransactionBuilder[] instead? 
I'm going to add a helper function in common-sdk to help send multiple transactions:
https://github.com/orca-so/orca/commit/642f43b22d67cec5232a3bf1e374f9246adcb80f#diff-761e6344537f10bbe5276e42c950aee2734c51cf3505d7ec6e2d23273406f5f2L235

Moving to an array also gives us the flexibility of adding transaction-packing to this method in the future (ex. make sure txn is full before moving on to the next transaction).

odcheung (commented):
looking good! got some minor comments

sdk/src/impl/whirlpool-impl.ts
odcheung: ah whoopies. What I meant was
1. For the tick-array fetching part, you can use `getTickArrayDataForPosition` which can simplify this a bit for you
2. Once you have the positionimpl object, you can call getLowerTickData / getUpperTickData.

PR #61, opened 2022-11-25
Add getWhirlpoolsWithParams method

terranmoccasin:
Adding a method to retrieve Whirlpools filtered by a set of parameters using `getProgramAccounts` in the `AccountFetcher`.

Currently filtering by only the `programId` and `configId`. In the future we can add more filter params as needed, and we can also consider adding a method to retrieve all positions for a Whirlpool.

Note that this is not added as a method on the `WhirlpoolClient`, but only as a fetcher method to retrieve `WhirlpoolData`.


sdk/src/impl/whirlpool-client-impl.ts
```diff
@@ -91,6 +91,12 @@ export class WhirlpoolClientImpl implements WhirlpoolClient {
     return whirlpools;
   }
 
+  public async getPoolsWithParams(params: GetPoolsParams, refresh = false): Promise<Whirlpool[]> {
+    const pools = await this.ctx.fetcher.listPoolsWithParams(params);
```
terranmoccasin, line 95: This method uses `getProgramAccounts`, which always returns the latest whirlpool data. With the `refresh` flag here the API feels weird. Here are two thoughts I have:

1. Add `refresh` to the fetcher and only update the cache for Whirlpools that are not in the cache - even though we have fetched the latest data. This would be used to maintain consistent SDK behavior.
2. Remove `refresh` for `getPoolsWithParams` and have the method always fetch new data

Let me know what you guys think.

sdk/src/types/public/anchor-types.ts
```diff
@@ -21,6 +22,13 @@ export enum AccountName {
   FeeTier = "FeeTier",
 }
 
+const coder = new BorshAccountsCoder(IDL);
```
terranmoccasin, line 25: Should we consider exporting this?

sdk/src/impl/whirlpool-client-impl.ts
odcheung: i'm leaning this method doesn't have a `refresh`. because `refresh = false` doesn't make much sense with the GPA call. 

This method should call this.getPools(addresses, false) though. Otherwise, it's going to make another RPC call to fetch the same data as the GPA call did.

sdk/src/whirlpool-client.ts
```diff
@@ -13,6 +13,20 @@ import {
...
+  /**
+   * Whirlpool program address.
+   */
+  programId: Address;
+  /**
+   * WhirlpoolsConfig account address.
+   */
+  configId: Address;
+};
+
```
odcheung, line 29: Can we move this to another type.ts file?

odcheung (commented):
lgtm! some small comments

terranmoccasin:
Update - after chatting with @odcheung - I've removed the `getPoolsWithParams` from the `WhirlpoolClient`. The GPA call is only in the `AccountFetcher`. If there is a need to construct `WhirlpoolImpl` objects from the GPA call we can add in the future, but currently fetching only `WhirlpoolData` is sufficient for the use-case I'm working on.

matt-allan (approved):

PR #65, opened 2022-12-04
Fix IDL import for BorshAccountsCoder

philcchen:


rawfalafel (approved):

philcchen:
for posterity, this is because the IDL generation produces both `.json` and `.ts`

at runtime the import will import from the `.json`, but at compile time typescript imports from the `.ts`

odcheung (approved):

PR #66, opened 2022-12-07
fix: getTickArrayPublicKeys edge case

yugure-orca:
Issue: https://app.asana.com/0/1202413563699470/1203471728007402/f

## fix: getTickArrayPublicKeys edge case
porting "shifted" in in_search_range of on-chain programs.
https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/state/tick.rs#L299

- getTickArrayPublicKeys: if the swap direction is B to A and tickCurrentIndex is greater than or equal to the last initializable tick in the TickArray, the TickArray adjacent to the swap direction is TickArray0
- TickArraySequence: Change the logic to determine whether TickArray0 contains tickCurrentIndex

## fix: test case & setting
- set jest timeout to 30 sec (default timeout (5 sec) is too short)
- add "integration" test to jest.config.js
- change "before" to "beforeAll"
  In jest there is no before, use beforeAll for one-time setups
    https://jestjs.io/docs/setup-teardown
- add refresh=true on getPosition
  When checking TickArrays that needs to be initialized, a "null" cache remains in the fetcher, and listing TickArrays for the position fails.

**passed all tests (integration & sdk)**

### DIFF
--- a/sdk/jest.config.js
+++ b/sdk/jest.config.js
@@ -1,7 +1,8 @@
 module.exports = {
   "roots": [
     "<rootDir>/src",
-    "<rootDir>/tests/sdk"
+    "<rootDir>/tests/sdk",
+    "<rootDir>/tests/integration"
   ],
   "testMatch": [
     "**/__tests__/**/*.+(ts|tsx|js)",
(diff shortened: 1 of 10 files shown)


odcheung (approved):
lgtm. thanks for catching this and awesome suite of tests!

PR #68, opened 2022-12-08
Support wSOL in close position and deps

atamari-orca:
- Adds wSOL support to `closePosition`, `collectRewards`, and `collectFees`
- Fixes the test utils' wSOL support
- Adds a test for `closePosition` with SOL as token A

### DIFF
--- a/sdk/tests/sdk/whirlpools/position-impl#collectRewards.test.ts
+++ b/sdk/tests/sdk/whirlpools/position-impl#collectRewards.test.ts
@@ -167,7 +167,6 @@ describe("PositionImpl#collectRewards()", () => {
 
       const r0Pubkey = await deriveATA(otherWallet.publicKey, rewards[0].rewardMint);
       const r0Account = await testCtx.whirlpoolCtx.fetcher.getTokenInfo(r0Pubkey, true);
-      console.log(r0Account?.amount);
       assert.ok(quote[0] && r0Account && r0Account.amount.eq(quote[0]));
 
       const r1Pubkey = await deriveATA(otherWallet.publicKey, rewards[1].rewardMint);
@@ -253,7 +252,6 @@ describe("PositionImpl#collectRewards()", () => {
 
       const r0Pubkey = await deriveATA(otherWallet.publicKey, rewards[0].rewardMint);
       const r0Account = await testCtx.whirlpoolCtx.fetcher.getTokenInfo(r0Pubkey, true);
-      console.log(r0Account?.amount);
       assert.ok(quote[0] && r0Account && r0Account.amount.eq(quote[0]));
 
       const r1Pubkey = await deriveATA(otherWallet.publicKey, rewards[1].rewardMint);


sdk/src/impl/position-impl.ts
```diff
@@ -246,20 +251,58 @@ export class PositionImpl implements Position {
...
+      const { ataTokenAddresses, resolveAtaIxs } = await resolveAtaForMints(this.ctx, {
+        mints: mintsToResolveAtasFor,
+        accountExemption,
+        receiver: destinationWalletKey,
+        payer: ataPayerKey,
+      });
+
+      for (const mint of Object.keys(ataTokenAddresses)) {
+        walletTokenAccountsByMint[mint] = ataTokenAddresses[mint];
+      }
```
odcheung, line 275: Use the spread operator here:
```
walletTokenAccountsByMint = {
...walletTokenAccountsByMint,
...ataTokenAddresses
}
```

sdk/src/impl/position-impl.ts
```diff
@@ -246,20 +251,58 @@ export class PositionImpl implements Position {
...
+    const solInWalletTokenAccounts = !!walletTokenAccountsByMint[NATIVE_MINT.toBase58()];
+    const mintsToResolveAtasFor = tokenAAndB.filter(
+      // 1. Filter out native mints since we don't use ATAs for them
+      // 2. Filter out mints for which we already have token accounts
+      (mint) => !TokenUtil.isNativeMint(mint) && !walletTokenAccountsByMint[mint.toBase58()]
+    );
 
-    const tokenOwnerAccountA = ataTokenAddresses[whirlpool.tokenMintA.toBase58()];
-    const tokenOwnerAccountB = ataTokenAddresses[whirlpool.tokenMintB.toBase58()];
+    if (mintsToResolveAtasFor.length > 0) {
```
odcheung, line 265: Can call resolveAta with an empty/native-mint-only dictionary to reduce complexity here. If it's empty, it'll return empty and you'll append empty onto your map later.

sdk/src/impl/position-impl.ts
```diff
@@ -271,22 +314,26 @@ export class PositionImpl implements Position {
...
     txBuilder.addInstruction(ix);
 
+    if (unwrapSolIx) {
+      txBuilder.addInstruction(unwrapSolIx);
+    }
+
     return txBuilder;
   }
 
   async collectRewards(
```
odcheung, line 333: same comments w.r.t resolveAta as `collectFees`

sdk/src/impl/position-impl.ts
```diff
@@ -315,17 +362,57 @@ export class PositionImpl implements Position {
...
-    });
+    const affiliatedMints = getTokenMintsFromWhirlpools([whirlpool]);
+    const walletTokenAccountsByMint = { ...ownerTokenAccountsRecord };
+
+    const solInRewardMints = initializedRewards.some((info) => TokenUtil.isNativeMint(info.mint));
+    const solInWalletTokenAccounts = !!walletTokenAccountsByMint[NATIVE_MINT.toBase58()];
+    const mintsToResolveAtasFor = affiliatedMints.filter(
+      // Filter out mints for which we already have token accounts
+      (mint) => !walletTokenAccountsByMint[mint.toBase58()]
+    );
```
odcheung, line 373: can we modify `getTokenMintsFromWhirlpools` to accept a flag that can return you `all-tokens`, `pool-tokens` & `reward-tokens` only?

That way we can reduce duplicated code here.

ResolveAta also automatically handles mints that already has an ATA, so you don't have to perform the check here.

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -522,10 +550,20 @@ export class WhirlpoolImpl implements Whirlpool {
...
     txBuilder.addInstruction(positionIx);
 
+    if (createAndInitWsolAccountIx) {
+      // Append the wSOL token account creation IXs to the end of the first TX
+      tokenAccountsTxBuilder.addInstruction(createAndInitWsolAccountIx);
+    }
+
+    if (unwrapWsolIx) {
+      // Append the wSOL unwrap IX (i.e. wSOL token account close IX) to the end of the second TX
+      txBuilder.addInstruction(unwrapWsolIx);
```
odcheung, line 560: can we append the wSOL wrapping ix (create & cleanup) to this txBuilder?
We seem to be adding a lot more paths to keep track of the wSOL ixs.

sdk/src/impl/position-impl.ts
```diff
@@ -224,7 +229,7 @@ export class PositionImpl implements Position {
 
   async collectFees(
     updateFeesAndRewards: boolean = true,
-    resolveATA: boolean = true,
```
odcheung, line 227: i think we should pass in resolveATA:boolean instead.
When you are composing this fn, you are resolving ATA outside and you can use false here. The method will still derive the aTA but will not resolve it.

sdk/tests/sdk/whirlpools/whirlpool-impl.test.ts
```diff
@@ -495,4 +495,217 @@ describe("whirlpool-impl", () => {
...
+     * 1. withdraw value from decrease_liq (decrease_quote, though not always accurate)
+     * 2. accrrued fees from trade (fee_quote)
+     * 3. Position PDA account rent return (balance from position address account)
+     * 4. wSOL rent-exemption close (getAccountExemption)
+     * 5. Position token account rent return (getAccountExemption)
+     *
+     * Other costs from payer, but not received by other wallet
+     * 1. close_position tx cost
+     * 2. ATA account initialization
+     */
```
odcheung, line 693: @atamari-orca - you were close!
The values missing were the rent collected from the closing of the Position PDA & the Position token ATA on the user's wallet.

sdk/tests/sdk/whirlpools/whirlpool-impl.test.ts
```diff
@@ -495,4 +495,217 @@ describe("whirlpool-impl", () => {
...
+
+    await ataTx?.buildAndExecute();
+    await closeTx.addSigner(otherWallet).buildAndExecute();
+
+    const otherWalletBalanceAfter = await ctx.connection.getBalance(otherWallet.publicKey);
+
+    const minAccountExempt = await ctx.fetcher.getAccountRentExempt();
+    const solReceived = otherWalletBalanceAfter - otherWalletBalanceBefore;
+
+    // TODO: Why is the reward quote only correct when data is collected after the tx is ran?
```
odcheung, line 675: I'm still investigating this one though. 

CollectRewardsQuote should return the correct value using the data prior to the collection of rewards, not after.

sdk/src/impl/position-impl.ts
```diff
@@ -315,17 +362,57 @@ export class PositionImpl implements Position {
...
+    const tokenOwnerAccountA = walletTokenAccountsByMint[whirlpool.tokenMintA.toBase58()];
+    invariant(
+      !!tokenOwnerAccountA,
+      `No owner token account provided for wallet ${destinationWalletKey.toBase58()} for token A ${whirlpool.tokenMintA.toBase58()}`
+    );
+    const tokenOwnerAccountB = walletTokenAccountsByMint[whirlpool.tokenMintB.toBase58()];
+    invariant(
+      !!tokenOwnerAccountB,
+      `No owner token account provided for wallet ${destinationWalletKey.toBase58()} for token B ${whirlpool.tokenMintB.toBase58()}`
+    );
```
odcheung, line 414: collectrewards needs to verify reward mint token accounts, not the pool token accounts.

odcheung:
Pushed out a PR for the comments here - https://github.com/orca-so/whirlpools/pull/69

sdk/src/utils/spl-token-utils.ts
```diff
@@ -19,3 +28,61 @@ export function getAssociatedTokenAddressSync(
 
   return address;
 }
+
+// TODO: Move this to common-sdk and use that in this SDK later
+export function wrapSOL(
```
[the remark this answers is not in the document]
odcheung: yeah i think moving to sync-native will simplify this flow. Let's do it in a dedicated effort though because if we do that we need to verify we are not breaking all the other SDK methods.

odcheung:
also good work id-ing all the other bugs!

sdk/tests/sdk/whirlpools/whirlpool-impl#closePosition.test.ts
```diff
@@ -0,0 +1,552 @@
...
+      whirlpool: poolData,
+      position: positionData,
+      tickLower: tickLowerData,
+      tickUpper: tickUpperData,
+    });
+
+    assert.ok(quote.feeOwedA.gtn(0) || quote.feeOwedB.gtn(0));
+  }
+
+  async function accrueRewards(fixture: WhirlpoolTestFixture) {
```
odcheung, line 131: The quotes doesn't actually accrue rewards. The rewards are accrued when the swap occurs (in accrue fees)

sdk/tests/sdk/whirlpools/whirlpool-impl#closePosition.test.ts
```diff
@@ -0,0 +1,552 @@
...
+            vaultAmount: new u64(vaultStartBalance),
+          },
+          {
+            emissionsPerSecondX64: MathUtil.toX64(new Decimal(10)),
+            vaultAmount: new u64(vaultStartBalance),
+          },
+        ],
+      });
+
+      await accrueRewards(fixture);
```
odcheung, line 273: this is true for other "rewards only" tests - no rewards is accrued because a swap must happen for the rewards to accumulate.

To test rewards only, call accrue_fees, then collect_fees, then perform your test on testClosePosition

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -497,16 +502,6 @@ export class WhirlpoolImpl implements Whirlpool {
...
-    } else if (shouldCollectFees || shouldCollectRewards) {
-      // We need to manually udpate the fees/rewards since there is no liquidity IX to do so
-      txBuilder.addInstruction(
-        updateFeesAndRewardsIx(this.ctx.program, {
-          whirlpool: positionData.whirlpool,
-          position: positionAddress,
-          tickArrayLower,
-          tickArrayUpper,
-        })
-      );
```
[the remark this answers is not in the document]
odcheung: Nvm. If decrease_l isn't needed then liquidity is 0, then update_fee isn't needed.

odcheung (approved):
Approve after we remove the console.logs. Can address the tests later.

PR #71, opened 2022-12-20
Update typedocs & env setup for 0.7.2

odcheung:
- Define typedoc setting in typedoc.json
- yarn add typedoc to dev-dependency
- Update category documentation for the newer classes
- Regenerate typedocs

## Note
- typedoc 0.23's new theme is not visually appealing compared to the current. Sticking with typedoc 0.22 for the time being.


yugure-orca (approved):
Thanks for updating docs!
I agree to stick with typedoc 0.22.

PR #72, opened 2022-12-23
Fix understated collectRewardsQuote and unit-tests for new collect functions

odcheung:
- Do not append updateFee ix if liquidity is zero in collectFee & collectReward util fn
- Fix collectRewardQuote use an estimated rewardGlobal accrued till the current timestamp
- Prior to this patch, collectRewardQuote is understating amount collected from call
- Fix unit-tests to strictly verify amount collected from collectFee, collectReward and closePosition

## Testing
- Verified using refactored / new unit-tests that compares the collectRewardsQuote value to the actual collected amount . The quote is generated using the same timestamp as the one executed on-chain in the test case to ensure the clock values are the same.
- Verified that Position.collectReward & collectFees will not fail on-chain if position liquidity is zero & updateFeesAndRewards is enabled
- Ran all the other tests, all pass.

### DIFF
--- a/sdk/src/impl/position-impl.ts
+++ b/sdk/src/impl/position-impl.ts
@@ -299,7 +299,7 @@ export class PositionImpl implements Position {
 
     const positionTokenAccount = await deriveATA(positionWalletKey, this.data.positionMint);
 
-    if (updateFeesAndRewards) {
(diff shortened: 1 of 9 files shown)


yugure-orca (approved):
LGTM!
Thanks for reverting "beforeAll" to "before" 🙏

PR #73, opened 2023-01-04
Collect protocol fees

rawfalafel:


sdk/src/impl/whirlpool-client-impl.ts
```diff
@@ -149,8 +152,8 @@ export class WhirlpoolClientImpl implements WhirlpoolClient {
               pos.whirlpool,
               this.ctx.program.programId
             ).publicKey;
-            tickArrayAddresses.add(lowerTickArrayPda);
-            tickArrayAddresses.add(upperTickArrayPda);
+            tickArrayAddresses.add(lowerTickArrayPda.toBase58());
```
rawfalafel, line 155: this used to insert duplicate addresses, since two `PublicKey` objects with the same addresses were considered to be different

sdk/src/utils/whirlpool-ata-utils.ts
```diff
@@ -49,14 +62,15 @@ export function getTokenMintsFromWhirlpools(
               hasNativeMint = true;
             }
             if (PoolUtil.isRewardInitialized(reward)) {
-              accu.add(reward.mint);
+              accu.add(reward.mint.toBase58());
```
rawfalafel, line 65: i saw that we were doing the same thing here, so i fixed it

sdk/src/instructions/composites/collect-protocol-fees.ts
```diff
@@ -0,0 +1,80 @@
...
+        tokenVaultB: pool.tokenVaultB,
+        tokenOwnerAccountA: ataTokenAddresses[pool.tokenMintA.toBase58()],
+        tokenOwnerAccountB: ataTokenAddresses[pool.tokenMintB.toBase58()],
+        collectProtocolFeesAuthority: poolConfig.collectProtocolFeesAuthority,
+      })
+    );
+  }
+
+  txBuilder.addInstructions(instructions);
+  const txSize = await txBuilder.txnSize({ latestBlockhash });
```
rawfalafel, line 74: this function doesn't try to build multiple transactions to fit an arbitrary amount of instructions. i wasn't really sure how to dedupe this transaction packing logic, so i left it like this to get something working quickly. feel free to suggest changes or make them yourself!

odcheung (approved):
the new util fn lgtm. I think the omission of txn packing is fine for protocol fees since it's not a widely used method. We can get back to it when we find the need for it in a future release.

With this change and the new request on exposing internal ix helper fns, we should expand the concept of sync/async helper methods like we do for the quoting stack. 
- async methods to help the user fetch everything and construct the txn (WhirlpoolClient)
- sync methods for users who only want the core business logic & txn construction.

WhirlpoolClient will only target users who want a plug and play experience, so should only get access to the most commonly used methods (whirlpool fetch, position management, swap etc)

I'm going to merge this one in and will move this method away from the WhirlpoolClient and into the new layer before publishing.

PR #74, opened 2023-01-04
expose a lot of internal functionality

mschneider:
It's a bit difficult to build a router based on orca whirpool, because one needs access to a bunch of it's internal methods. This PR makes a bunch of internal modules public and adds a new utility method to create whirpool swap ix without any convenince /magic that could cause a network request.

### DIFF
--- a/sdk/src/impl/index.ts
+++ b/sdk/src/impl/index.ts
@@ -1,0 +1,4 @@
+export * from "./position-impl";
+export * from "./util";
+export * from "./whirlpool-client-impl";
+export * from "./whirlpool-impl";
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -9,7 +9,7 @@ import {
 } from "@orca-so/common-sdk";
 import { Address, BN, translateAddress } from "@project-serum/anchor";
 import { NATIVE_MINT } from "@solana/spl-token";
-import { Keypair, PublicKey } from "@solana/web3.js";
+import { Keypair, PublicKey, TransactionInstruction } from "@solana/web3.js";
 import invariant from "tiny-invariant";
 import { WhirlpoolContext } from "../context";
 import {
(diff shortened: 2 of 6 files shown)


odcheung:
hey max! thank you for submitting this. Following up on the request in this PR, I'm working on exposing the internal util methods to construct instructions. 

Can you also expand on the use-cases for exposing the internal swap-quote classes?
You can get the sync-version of the quote with `swapQuoteWithParams`. That calls `simulateSwap` directly.

mschneider:
Oh i think that was just me trying to get quick access to some internal
methods while prototyping. I don't think I need it anymore, but when
building trading bots it's really helpful to be able to call every method
directly.

FYI: the project that made me write the above changes is open source:
https://github.com/blockworks-foundation/mango-router

Did you see my most recent commit? The typechecker complained, that one of
the methods in the SDK wasn't actually async although declared as such 🤨

odcheung:
Gotcha. I am following up to the requests in this PR in two separate PRs:
[1. Add WhirlpoolIxBuilders namespace and expose internal swap method](https://github.com/orca-so/whirlpools/pull/76)
[2. Remove async from TickUtil.getTickArrayPDAs](https://github.com/orca-so/whirlpools/pull/77)


Post 0.7.3, you will be able to:
1. You can now access the swap tx construction with
```
const txBuilder =  WhirlpoolIxBuilders.swap(this.ctx, {
      input: quote,
      whirlpool: this.data,
      poolAddress: this.address,
      wallet: sourceWalletKey,
      resolveAssociatedTokenAccounts: false
    });
```

Note that we'll still derive the ATA for you. It is the same as mango's getAssociatedTokenAddress util method. 

2. Once you have the txBuilder, you can call this to get the instruction
```
txBuilder.compressIx(false).instructions[0]
```
Not the most straightforward syntax... we'll make an update to `TransactionBuilder` soon to have a sync method to make this easier.

3. We're not going to expose the internal swap quote methods as it seems like the current API is sufficient for the use-case. We want to have a consistent quote public API and also the flexibility to swap out the quote implementation down the road without too much disruption to SDK users. But please do let us know if there's an edge use-case where this is necessary! happy to talk about how to proceed further


I looked over the mango router and i think these changes these should be sufficient. If not, let me know otherwise!

PR #77, opened 2023-01-10
Remove async from TickUtil.getTickArrayPDAs

odcheung:
TickUtil.getTickArrayPDAs to be a sync method and should not be async

This is a follow up to max's comment here - https://github.com/orca-so/whirlpools/pull/74


yugure-orca (approved):
LGTM!
note: this change seems a kind of "breaking change".

odcheung:
> LGTM! note: this change seems a kind of "breaking change".

I think it won't be a breaking change as long as the return value is the same. Normally async methods return Promise<>. But in this case, we are always returning PDA[]. 

I made a similar change [here](https://github.com/orca-so/whirlpools/commit/2a077a7c8ff250999d3bf59d33ecc192486933d0).

PR #79, opened 2023-01-11
Catch tick boundary exception in priceToTickIndex

rawfalafel:
- When converting prick to tick index, ensure that the price is within bounds to prevent exception from being thrown
- Remove yarn-error.log


sdk/src/utils/public/price-math.ts
```diff
@@ -107,9 +107,18 @@ export class PriceMath {
...
+
+    if (sqrtPriceX64.lt(new BN(MIN_SQRT_PRICE))) {
+      console.warn("Provided price is too low, returning min tick index.");
+      return MIN_TICK_INDEX;
+    }
+
+    return PriceMath.sqrtPriceX64ToTickIndex(sqrtPriceX64);
   }
 
   public static priceToInitializableTickIndex(
```
rawfalafel, line 124: if the caller specified a price that was too low or too high, `sqrtPriceX64ToTickIndex` would throw an exception. if the price is too low or too high, return the min or max valid tick index

rawfalafel:
fixes: https://app.asana.com/0/search?any_tags.ids=1203705058463326&child=1203651365525314

odcheung:
I feel that these util methods should fail loudly if it cannot handle the input parameters. Otherwise, it'll be difficult to notice things going wrong. We should also avoid changing the expectation for all of our SDK consumers since this is a very commonly used method on our live external API.

Reading through the bug report, I believe the UI should be the one clarifying it's bounds and responsible for catching/handling out of bounds prices into it's defined UX behaviour (auto change to min/max price).

rawfalafel:
good point. i think what's dangerous is that it's not obvious that `priceToTickIndex` can even throw an exception, since the exception gets thrown by an inner function, `sqrtPriceX64ToTickIndex`. this is what led to the entire page crashing, which we definitely want to avoid

i see two solutions:
 - add an annotation to every function that calls `sqrtPriceX64ToTickIndex`. however, this is easy to miss if we add a new function that calls `sqrtPriceX64ToTickIndex` in the future
 - have `sqrtPriceX64ToTickIndex` potentially return an error type, so that it's obvious that it can return an invalid value. this requires a larger change
 
 thoughts?

rawfalafel:
closing this b/c we went down a different route for the fix

PR #81, opened 2023-01-23
Make collect all completely parallel

yugure-orca:
All TransactionBuilders generated by collectAllForPositionAddressesTxns can now be executed in parallel. Since ATAs are created with CreateIdempotent when necessary, the first TransactionBuilder no longer needs to be executed first.

Added a test case that calls collectAllForPositionAddressesTxns.

``anchor test`` passed.

Issue: https://app.asana.com/0/1202413563699470/1203609084790769/f

### DIFF
--- a/sdk/tests/sdk/whirlpools/whirlpool-impl#collectFeesAndRewardsForPositions.test.ts
+++ b/sdk/tests/sdk/whirlpools/whirlpool-impl#collectFeesAndRewardsForPositions.test.ts
@@ -320,6 +320,7 @@ describe("WhirlpoolImpl#collectFeesAndRewardsForPositions()", () => {
         position: positionData,
         tickLower: tickLowerData,
         tickUpper: tickUpperData,
+        timeStampInSeconds: poolData!.rewardLastUpdatedTimestamp,
       });
 
       assert.ok(feeQuote.feeOwedA.gt(ZERO));
@@ -370,6 +371,7 @@ describe("WhirlpoolImpl#collectFeesAndRewardsForPositions()", () => {
         position: positionData,
         tickLower: tickLowerData,
         tickUpper: tickUpperData,
+        timeStampInSeconds: poolData!.rewardLastUpdatedTimestamp,
       });
 
       assert.ok(feeQuote.feeOwedA.eq(ZERO));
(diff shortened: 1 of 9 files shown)


sdk/tests/sdk/whirlpools/whirlpool-impl#collectFeesAndRewardsForPositions.test.ts
```diff
@@ -0,0 +1,402 @@
...
+      const tickUpperData = position.getLowerTickData();
+
+      const feeQuote = collectFeesQuote({
+        whirlpool: poolData!,
+        position: positionData,
+        tickLower: tickLowerData,
+        tickUpper: tickUpperData,
+      });
+
+      const rewardQuote = collectRewardsQuote({
```
odcheung, line 368: pass in `timeStampInSeconds: pool.getData().rewardLastUpdatedTimestamp` here.
CollectRewardQuote needs the latest timestamp on-chain to derive the correct derived reward values on-chain. (should be fine since you are not checking exact numbers, but still good to have it just in case)

rawfalafel:
is this blocked on something @yugure-orca ?

sdk/tests/sdk/whirlpools/whirlpool-impl#collectFeesAndRewardsForPositions.test.ts
yugure-orca: makes a lot of sense!
I have added timeStampInSeconds param. 🙂

yugure-orca:
@rawfalafel 
sorry, there is no block factor.
I have committed the updates and should be able to merge them soon.

sdk/tests/sdk/whirlpools/whirlpool-impl#collectFeesAndRewardsForPositions.test.ts
yugure-orca: @atamari-orca 
Thanks for your review!

nope.
Since rewardLastUpdatedTimestamp is not updated until a transaction occurs, specifying rewardLastUpdatedTimestamp would underestimate the value in a pool where rewards are being paid out but no swap has occurred at all.

In testing, it makes sense to check the exact value based on rewards paid out up to a certain point in time.
On the other hand, a script examining pending reward should also take into account the passage of time up to now.
Since omitting this argument uses the current time, I believe it is correct to omit it and use the default value.

PR #90, opened 2023-03-14
Add PoolGraph API and AdjacencyGraph implementation

odcheung:
- Add PoolGraph API to allow trade route path finding with max hop of 2
- Add tests to verify behavior
- Update docs for PoolGraph API
- bump common-sdk to v0.1.12

Adjacency Graph Implementation Author - @philcchen


sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    if (poolGraph[mintB] === undefined) {
+      poolGraph[mintB] = new Set();
+    }
+
+    poolGraph[mintA].add({ address: addr, otherToken: mintB });
+    poolGraph[mintB].add({ address: addr, otherToken: mintA });
+    return poolGraph;
+  }, {});
```
yugure-orca, line 100: set cannot deduplicate for objects.
```suggestion
  const processed = new Set<string>();
  const poolGraphSet = pools.reduce((poolGraph: Record<string, PoolGraphEdge[]>, pool) => {
    const { address, tokenMintA, tokenMintB } = pool;
    const [addr, mintA, mintB] = AddressUtil.toPubKeys([address, tokenMintA, tokenMintB]).map(
      (pk) => pk.toBase58()
    );

    if (processed.has(addr)) {
      return poolGraph;
    }

    if (poolGraph[mintA] === undefined) {
      poolGraph[mintA] = [];
    }

    if (poolGraph[mintB] === undefined) {
      poolGraph[mintB] = [];
    }

    poolGraph[mintA].push({ address: addr, otherToken: mintB });
    poolGraph[mintB].push({ address: addr, otherToken: mintA });
    processed.add(addr);
    return poolGraph;
  }, {});
```

yugure-orca (commented):
When developers start using getRoute and getAllRoutes, I feel like I will have to repeatedly explain the direction of the route and mint ordering.

IMO, It seems counter-intuitive that a route that is intended to be A-X-B may return B-X-A. 🤔

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const walkEntries = Object.entries(walkMap).map(([routeId, walks]) => {
+      const [startMint, endMint] = routeId.split("-");
+      const paths = walks.map<Route>((walk) => {
+        return {
+          startMint,
+          endMint,
+          edges: walk,
+        };
+      });
```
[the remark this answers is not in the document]
odcheung: good points. directionality can make things simpler. 
The edges one is currently whirlpool addrs but i think we can take it a step further to provide trade directions

sdk/src/utils/public/pool-graph.ts
```diff
@@ -0,0 +1,154 @@
...
+   * Returns a route id for a swap between source & destination mint for the Orca UI.
+   * The route id is a string of the two mints in alphabetical order, separated by a dash.
+   *
+   * @param sourceMint - The token the swap is trading from.
+   * @param destinationMint - The token the swap is trading for.
+   * @returns A string representing the routeId between the two provided tokens.
+   */
+  static getRouteId(tokenA: Address, tokenB: Address): string {
+    const mints = [AddressUtil.toString(tokenA), AddressUtil.toString(tokenB)];
+    const sortedMints = mints.sort();
```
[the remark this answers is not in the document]
odcheung: let me rethink about the route-id a bit together with the directionality conversation. But originally, routeId is just to maintain the same key between two tokens within the adjacency pool graph structure. The correct ordering does not matter as long as the same ordering is maintained. Doing a Pubkey comparison is slower than string.

sdk/src/utils/graphs/adjacency-pool-graph.ts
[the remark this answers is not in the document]
odcheung: will address this with a comment below w.r.t route Id changes

sdk/src/utils/public/pool-graph.ts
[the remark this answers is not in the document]
odcheung: will address this with a comment below w.r.t route Id changes

odcheung:
@yugure-orca Good catch on the confusing nature of the previous usage of route id.

I've made the following changes:

**External API changes*
1. Renamed routeId to searchRouteId. searchRouteId is a simple concat between `tokenFrom-tokenTo`.
2. RouteMap has now been converted to RouteSearchEntires. entires are of the form of `[searchRouteId, [routes]`. If searchRouteId returns no route, it'll return []. The number of entires will equal the number of search input.
3. Fixed directionality. A search on [A, B] will return routes that goes from [A -> ? -> B], vice versa. However, users would still have to figure out the correct `aToB` when they use these pools.
4. Renamed `edges` to `hops` in `Route`. Also wrapped `address` in a `Hop` type to make it easier if we want to add additional info like `aToB` down the road.

**Internal usage of RouteId**
- Internally, an internalRouteId is used by the AdjacencyPoolGraph to store saved walks.
- Fixed a bug on the directionality. Previously there was no connection between routeId ordering and the stored route direction. Now it always stores in the direction of the internal route id ordering.
- Optimized a little bit - Given search token pairs [tokenA, tokenB] & [tokenB, tokenA], it will only walk once using the internal route id ordering (ex. A-B)

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+  readonly graph: {
+    [k: string]: PoolGraphEdge[];
+  };
+
+  constructor(pools: PoolTokenPair[]) {
+    this.graph = buildPoolGraph(pools);
+  }
+
+  getRoute(startMint: Address, endMint: Address, options?: RouteFindOptions): Route[] {
+    const [startMintKey, endMintKey] = [
```
philcchen, line 24: nit: this method can just be `getAllRoutes([[startMint, endMint]], options)`

sdk/src/utils/public/pool-graph.ts
```diff
@@ -0,0 +1,163 @@
...
+  poolAddress: Address;
+}
+
+/**
+ * Options for finding a route between two tokens
+ *
+ * @category PoolGraph
+ * @param intermediateTokens - A list of tokens that can be used as intermediate hops
+ */
+export type RouteFindOptions = {
```
philcchen, line 56: why `search` vs `find`?

sdk/src/utils/public/pool-graph.ts
```diff
@@ -0,0 +1,163 @@
...
+    return new AdjacencyPoolGraph(poolTokenPairs);
+  }
+}
+
+/**
+ * A utility class for working with pool graphs
+ * @category PoolGraph
+ */
+export class PoolGraphUtils {
+  static readonly ROUTE_ID_DELIMINTER = "-";
```
philcchen, line 136: nit: `DELIMITER`

sdk/src/utils/public/pool-graph.ts
```diff
@@ -0,0 +1,163 @@
...
+  static buildPoolGraph(poolTokenPairs: PoolTokenPair[]): PoolGraph {
+    return new AdjacencyPoolGraph(poolTokenPairs);
+  }
+}
+
+/**
+ * A utility class for working with pool graphs
+ * @category PoolGraph
+ */
+export class PoolGraphUtils {
```
philcchen, line 135: nit: Generally prefer separate files for separate classes

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+      AddressUtil.toString(endMint),
+    ];
+
+    const walkMap = findWalks(
+      [[startMintKey, endMintKey]],
+      this.graph,
+      options?.intermediateTokens.map((token) => AddressUtil.toString(token))
+    );
+
+    const internalRouteId = getInternalRouteId(startMintKey, endMintKey);
```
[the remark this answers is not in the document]
odcheung: i think the complexity isn't that great here but the space savings can be quite large (basically doubling the size). I'd prefer if we make this trade off knowing that our whirlpool size will only grow

sdk/src/utils/graphs/adjacency-pool-graph.ts
odcheung: yikes. good call.

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+type PoolGraphEdge = {
+  address: string;
+  otherToken: string;
+};
+
+// A record of route-id (tokenA-tokenB) to a list of edges
+type PoolWalks = Record<string, string[][]>;
+
+function buildPoolGraph(pools: PoolTokenPair[]) {
```
yugure-orca, line 103: nit: how about clarifying the return type (AdjacencyPoolGraphMap)

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+  PoolTokenPair,
+  Route,
+  RouteFindOptions,
+  RouteSearchEntires
+} from "../public/pool-graph";
+
+export class AdjacencyPoolGraph implements PoolGraph {
+  readonly graph: {
+    [k: string]: PoolGraphEdge[];
+  };
```
yugure-orca, line 17: nit: can we replate the type definition with AdjacencyPoolGraphMap ?

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    if (!existingAddressesForB.includes(addr)) {
+      poolGraph[mintB].push({ address: addr, otherToken: mintA });
+    }
+
+    return poolGraph;
+  }, {});
+
+  return Object.fromEntries(
+    Object.entries(poolGraphSet).map(([mint, otherMints]) => [mint, Array.from(otherMints)])
```
yugure-orca, line 132: nit: I think otherMints is now an array.

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+  poolGraph: AdjacencyPoolGraphMap,
+  intermediateTokens?: string[]
+) {
+  const walks: PoolWalks = {};
+
+  tokenPairs.forEach(([tokenMintFrom, tokenMintTo]) => {
+    let routes = [];
+
+    const internalRouteId = getInternalRouteId(tokenMintFrom, tokenMintTo)
+    const poolsForTokenFrom = poolGraph[tokenMintFrom] || [];
```
yugure-orca, line 149: If we adjust the from/to from deconstructRouteId at this point, there would be no need to reverse at the end.

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  return Object.fromEntries(
+    Object.entries(poolGraphSet).map(([mint, otherMints]) => [mint, Array.from(otherMints)])
+  );
+}
+
+// This is currently hardcoded to find walks of max length 2, generalizing to longer walks
+// may mean that a adjacency matrix might have better performance
+function findWalks(
+  tokenPairs: (readonly [string, string])[],
```
yugure-orca, line 139: I noticed that there are edge case inputs.
e.g.: [SOL, SOL], [USDC, USDC]

We didn't think of it because we are thinking of it for trading purposes, but it is likely to be the first input for users who use it to try arbitrage.
It might be better to handle this independently from the current logic.🤔

- single hop: no routes
- two hops: same pair with different fee tiers
  ok: SOL -- SOL/USDC(64) --> USDC -- SOL/USDC(8) --> SOL
  ok: SOL -- SOL/USDC(8) --> USDC --> SOL/USDC(64) --> SOL
  ng: SOL -- SOL/USDC8(64) --> USDC -- SOL/USDC(64) --> SOL

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+
+
+
+function getHopsFromRoute(internalRouteId: string, searchRouteId: string, route: string[]): Hop[] {
+  const [intStartA] = PoolGraphUtils.deconstructRouteId(internalRouteId);
+  const [searchStartA] = PoolGraphUtils.deconstructRouteId(searchRouteId);
+  const shouldReverseRoute = searchStartA !== intStartA;
+  const finalRoutes = shouldReverseRoute ? route.reverse() : route;
```
yugure-orca, line 87: is ``route.slice().reverse()``  safer ?

In my understanding, .reverse() is in-place method, so I am concerned walkMap[internalRouteId] is also reversed.
I am concerned about whether a problem will arise if the reversal order is made first and the nonreversal order comes later.
edge case input: [[A, B], [B, A]] and [[B, A], [A, B]]

yugure-orca (commented):
Thanks for the update!
I think it is a good idea to make it intuitive for SDK users.👍
I also agree with keeping the results in a specific direction internally to save half the memory.

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+    this.graph = buildPoolGraph(pools);
+  }
+
+  getRoute(startMint: Address, endMint: Address, options?: RouteFindOptions): Route[] {
+    const [startMintKey, endMintKey] = [
+      AddressUtil.toString(startMint),
+      AddressUtil.toString(endMint),
+    ];
+
+    const walkMap = findWalks(
```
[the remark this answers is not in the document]
odcheung: good call. added a cache

sdk/src/utils/graphs/adjacency-pool-graph.ts
odcheung: ah yes. given that we don't use it, i would say let's throw an error and mark unsupported. Devs who has a need for this usecase can contribute themselves since this is an open-source project.

sdk/src/utils/graphs/adjacency-pool-graph.ts
odcheung: good catch!

sdk/src/utils/graphs/adjacency-pool-graph.ts
odcheung: to clarify, i ended up not throwing an error and these routes will just return [].

sdk/src/utils/graphs/adjacency-pool-graph.ts
odcheung: nice!

odcheung:
@philcchen @yugure-orca - thank you for the reviews! I've pushed out another set of changes containing your suggestions.

sdk/src/utils/graphs/adjacency-pool-graph.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    if (poolGraph[mintA] === undefined) {
+      poolGraph[mintA] = [];
+    }
+
+    if (poolGraph[mintB] === undefined) {
+      poolGraph[mintB] = [];
+    }
+
+    const existingAddressesForA = poolGraph[mintA].map((p) => p.address);
```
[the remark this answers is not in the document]
odcheung: thanks! I ended up adding a Record<mint, Set<string>> to keep track of what had been added for a particular mint.

odcheung:
> @odcheung (cc: @philcchen ) Regarding the cache, if intermediate tokens are changed, the cache may need to be cleared. However, routes can be cached externally, and I believe UI also caches them with useQuery. I think the cache should not be internalized as it would increase memory usage, and unless the route is computed over 2 steps, there seems to be little benefit from caching. 🤔

Thinking about this more - We do need a cache somewhere and I do buy the argument that this should fall into the responsibility of the UI / sdk-user who may have a more opinionated way of implementing the cache. Therefore, reducing complexity for this base level class makes sense. I've updated the patch to reflect this

yugure-orca (approved):
Thanks for the update!
LGTM! 🚀

PR #92, opened 2023-03-16
fix: test cases (tokenAIsNative related)

yugure-orca:
- tokenAIsNative was extended to reuseTokenA, but there was a bug that was fixed.
- Change the generation logic so that the mint specified in reuseTokenA is always tokenA (first mint). 
  Removed cause of accidental SOL-SPL pool test failures.

### DIFF
--- a/sdk/tests/utils/test-builders.ts
+++ b/sdk/tests/utils/test-builders.ts
@@ -59,13 +59,12 @@ export const createInOrderMints = async (context: WhirlpoolContext, reuseTokenA?
   const tokenXMintPubKey = reuseTokenA ?? (await createMint(provider));
 
   // ensure reuseTokenA is the first mint if reuseTokenA is provided
-  while (true) {
+  let ordered;
+  do {
     const tokenYMintPubKey = await createMint(provider);
-    const ordered = PoolUtil.orderMints(tokenXMintPubKey, tokenYMintPubKey).map(AddressUtil.toPubKey);
-    if (!reuseTokenA || ordered[0].equals(reuseTokenA)) {
-      return ordered;
-    }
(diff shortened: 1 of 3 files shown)


sdk/tests/utils/test-builders.ts
```diff
@@ -56,9 +56,16 @@ export const generateDefaultConfigParams = (
 
 export const createInOrderMints = async (context: WhirlpoolContext, reuseTokenA?: PublicKey) => {
   const provider = context.provider;
-  const tokenXMintPubKey = reuseTokenA || (await createMint(provider));
-  const tokenYMintPubKey = await createMint(provider);
-  return PoolUtil.orderMints(tokenXMintPubKey, tokenYMintPubKey).map(AddressUtil.toPubKey);
+  const tokenXMintPubKey = reuseTokenA ?? (await createMint(provider));
+
+  // ensure reuseTokenA is the first mint if reuseTokenA is provided
+  while (true) {
```
odcheung, line 62: this feels scary but i suppose it works XD. 

up to you since it is test code, but an alternative:
```
let ordered;
do {
    const tokenYMintPubKey = await createMint(provider);
   ordered = PoolUtil.orderMints(tokenXMintPubKey, tokenYMintPubKey).map(AddressUtil.toPubKey);
} while (!reuseTokenA || ordered[0].equals(reuseTokenA))

return ordered
```

odcheung (approved):
lgtm

sdk/tests/utils/test-builders.ts
yugure-orca: Thanks!
I've updated it to do-while style.

PR #93, opened 2023-03-17
optimize price module perf

yugure-orca:
Reflect performance-enhancing changes that were necessary to integrate into the UI.

### optimize getMostLiquidPool
Eliminate PDA calculations by first calculating the most liquid pool for each pair with quoteToken, rather than checking for its existence in the poolMap while generating PDAs.

### optimize fetchTickArraysForPools
Only fetch TickArrays in the necessary direction, since swap is only attempted in the direction of the quoteToken with the higher priority.

Although the fetching of TickArray can be done in parallel, we want to minimize the number of account fetched because it takes time to decode and parse a 10KB account.

### DIFF
--- a/sdk/src/prices/calculate-pool-prices.ts
+++ b/sdk/src/prices/calculate-pool-prices.ts
@@ -1,5 +1,5 @@
 import { AddressUtil, DecimalUtil, Percentage } from "@orca-so/common-sdk";
-import { Address, translateAddress } from "@project-serum/anchor";
+import { Address } from "@project-serum/anchor";
 import { u64 } from "@solana/spl-token";
 import { PublicKey } from "@solana/web3.js";
 import Decimal from "decimal.js";
@@ -119,10 +119,10 @@ export function calculatePricesForQuoteToken(
 
       // The quote token is the output token.
       // Therefore, if the quote token is mintB, then we are swapping from mintA to mintB.
-      const aToB = translateAddress(mintB).equals(quoteTokenMint);
+      const aToB = AddressUtil.toPubKey(mintB).equals(quoteTokenMint);
 
       const baseTokenMint = aToB ? mintA : mintB;
-      const poolCandidate = mostLiquidPools[translateAddress(baseTokenMint).toBase58()];
+      const poolCandidate = mostLiquidPools[AddressUtil.toString(baseTokenMint)];
       if (poolCandidate === undefined) {
         return [mint.toBase58(), null];
       }
--- a/sdk/src/prices/calculate-pool-prices.ts
+++ b/sdk/src/prices/calculate-pool-prices.ts
@@ -15,7 +15,6 @@ import {
 import { swapQuoteWithParams } from "../quotes/public/swap-quote";
 import { TickArray, WhirlpoolData } from "../types/public";
 import { PoolUtil, PriceMath, SwapUtils } from "../utils/public";
-import { PDAUtil } from "../utils/public/pda-utils";
 
 function checkLiquidity(
   pool: WhirlpoolData,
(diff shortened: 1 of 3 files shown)


staccDOTsol:
To optimize the `getMostLiquidPool` function, you can pre-calculate the most liquid pool for each pair with the quoteToken, so you don't have to check for its existence in the poolMap while generating PDAs. This will eliminate the need for PDA calculations and improve performance.

For `fetchTickArraysForPools`, you can optimize by only fetching TickArrays in the necessary direction, since swap is only attempted in the direction of the quoteToken with the higher priority. Although the fetching of TickArray can be done in parallel, you want to minimize the number of accounts fetched because it takes time to decode and parse a 10KB account.

Additionally, you can consider implementing caching mechanisms to reduce the number of API calls and improve performance. You can also use profiling tools to identify any bottlenecks in the code and optimize accordingly.

staccDOTsol:
To optimize the `getMostLiquidPool` function, you can first calculate the most liquid pool for each pair with `quoteToken`, rather than checking for its existence in the `poolMap` while generating PDAs. This will eliminate the need for PDA calculations and improve performance.

To optimize the `fetchTickArraysForPools` function, you can only fetch TickArrays in the necessary direction, since swap is only attempted in the direction of the `quoteToken` with the higher priority. Although the fetching of TickArray can be done in parallel, it's important to minimize the number of accounts fetched because it takes time to decode and parse a 10KB account.

sdk/src/prices/price-module.ts
```diff
@@ -280,33 +280,37 @@ export class PriceModuleUtils {
...
+    // select tick arrays based on the direction of swapQuote
+    // TickArray is a large account, which affects decoding time.
+    // Fetching can be performed in parallel, but it is preferable to fetch the minimum number of accounts necessary.
+    const tickArrayAddressSet = new Set<string>();
+    Object.entries(pools).forEach(([address, pool]) => {
+      const orderA = getQuoteTokenOrder(pool.tokenMintA);
+      const orderB = getQuoteTokenOrder(pool.tokenMintB);
+
+      if (orderA === orderB) {
+        // neigher tokenMintA nor tokenMintB is a quote token
```
odcheung, line 298: *nit - spelling `neither`

sdk/src/prices/calculate-pool-prices.ts
```diff
@@ -71,38 +70,31 @@ function checkLiquidity(
...
-    return null;
-  }
+      const baseTokenMint = pool.tokenMintA.equals(quoteTokenMint) ? mintB : mintA;
 
-  return pools.slice(1).reduce<PoolObject>((acc, { address, pool }) => {
-    if (pool.liquidity.lt(acc.pool.liquidity)) {
-      return acc;
-    }
+      const existingPool = mostLiquidPools.get(baseTokenMint);
+      if (!existingPool || pool.liquidity.gt(existingPool.pool.liquidity)) {
```
odcheung, line 92: i suppose we are not changing anything here, but is there an inherent bias towards high TS pools since on an equal price impact basis, there's technically more liquidity in each tick since each tick represents a larger bp?

odcheung: don't let this block this PR, but just want to confirm

sdk/src/prices/calculate-pool-prices.ts
```diff
@@ -127,8 +121,9 @@ export function calculatePricesForQuoteToken(
       // Therefore, if the quote token is mintB, then we are swapping from mintA to mintB.
       const aToB = translateAddress(mintB).equals(quoteTokenMint);
 
-      const poolCandidate = getMostLiquidPool(mintA, mintB, poolMap, config);
-      if (poolCandidate == null) {
+      const baseTokenMint = aToB ? mintA : mintB;
+      const poolCandidate = mostLiquidPools[translateAddress(baseTokenMint).toBase58()];
```
odcheung, line 125: can use AddressUtil.toString(baseTokenMint)

odcheung (approved):
lgtm overall! just some small comments

sdk/src/prices/calculate-pool-prices.ts
yugure-orca: When there are pools with different tickSpacing for the same token pair, I think the pools with higher fee rate are disadvantageous for the ExactOut mode estimate.
This is because more of the input to output 1000 USDC would be consumed as fees, and as a result, more input tokens are required.

Where this selection logic becomes not-optimal is when there is significant liquidity remaining where the price has moved slightly in the trade direction from the current price.
That is, a pool where liquidity is highly concentrated at a single tick and just slightly outside of that range (e.g. stable pairs like USDT/USDC).
In this case, the liquidity field of the Whirlpool account will have a smaller value, but the liquidity will be larger than in the other pools as it moves in the trade direction, and the trade results will be better.

PR #94, opened 2023-03-22
Rename PoolGraph API to use graph terminology

odcheung:
- Rename `Route` to `Path`
- Rename `Hop` to `Edge`
- Update docs
- Verified functionality with tests


philcchen (approved):
lgtm, though Path is also often used in node, but I think we should be okay?

PR #95, opened 2023-04-07
Get anchor account sizes by name instead of IDL index

terranmoccasin:
Previously whirlpool account size was hardcoded to be an IDL index with the assumption that we would not add more account types. With `PositionBundle` added, the ordering of accounts on the IDL changed.

This patch fetches account sizes by the `AccountName` instead and adds a test to verify all `AccountName` enums have an IDL account entry.

### DIFF
--- a/sdk/src/types/public/anchor-types.ts
+++ b/sdk/src/types/public/anchor-types.ts
@@ -32,10 +32,13 @@ const IDL = WhirlpoolIDL as Idl;
 export const WHIRLPOOL_CODER = new BorshAccountsCoder(IDL);
 
 /**
- * Size of the Whirlpool account in bytes.
- * @category Solana Accounts
+ * Get the size of an account owned by the Whirlpool program in bytes.
+ * @param accountName Whirlpool account name
+ * @returns Size in bytes of the account
  */
-export const WHIRLPOOL_ACCOUNT_SIZE = WHIRLPOOL_CODER.size(IDL.accounts![4]);
+export function getAccountSize(accountName: AccountName) {
+  return WHIRLPOOL_CODER.size(IDL.accounts!.find((account) => account.name === accountName)!);
+}
 
 /**
  * @category Solana Accounts
--- a/sdk/src/types/public/anchor-types.ts
+++ b/sdk/src/types/public/anchor-types.ts
@@ -37,8 +37,32 @@ export const WHIRLPOOL_CODER = new BorshAccountsCoder(IDL);
  * @returns Size in bytes of the account
  */
 export function getAccountSize(accountName: AccountName) {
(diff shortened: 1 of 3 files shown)


sdk/src/types/public/anchor-types.ts
```diff
@@ -32,10 +32,13 @@ const IDL = WhirlpoolIDL as Idl;
...
 /**
- * Size of the Whirlpool account in bytes.
- * @category Solana Accounts
+ * Get the size of an account owned by the Whirlpool program in bytes.
+ * @param accountName Whirlpool account name
+ * @returns Size in bytes of the account
  */
-export const WHIRLPOOL_ACCOUNT_SIZE = WHIRLPOOL_CODER.size(IDL.accounts![4]);
+export function getAccountSize(accountName: AccountName) {
+  return WHIRLPOOL_CODER.size(IDL.accounts!.find((account) => account.name === accountName)!);
```
yugure-orca, line 40: IDL doesn't contain the calculated account-size and the Coder calculates the total size of each field.
And PositionBundle has 64 bytes reserved space for the future use.

https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/state/position_bundle.rs#L9-L16
https://github.com/orca-so/whirlpools/blob/main/sdk/src/artifacts/whirlpool.json#L1893-L1914

So I think we need to use the following patch. (of course, this patch is not cool 😨)
```
const reserved = (accountName === AccountName.PositionBundle) ? 64 : 0;
return WHIRLPOOL_CODER.size(.....) + reserved;
```

sdk/src/types/public/anchor-types.ts
```diff
@@ -32,10 +32,13 @@ const IDL = WhirlpoolIDL as Idl;
 export const WHIRLPOOL_CODER = new BorshAccountsCoder(IDL);
 
 /**
- * Size of the Whirlpool account in bytes.
- * @category Solana Accounts
+ * Get the size of an account owned by the Whirlpool program in bytes.
+ * @param accountName Whirlpool account name
+ * @returns Size in bytes of the account
  */
-export const WHIRLPOOL_ACCOUNT_SIZE = WHIRLPOOL_CODER.size(IDL.accounts![4]);
```
yugure-orca, line 38: Since it is an externally disclosed constant, how about deprecating it for the next version ?

```
 * @deprecated Please use {@link getAccountSize} instead.
 * ...
 */
export const WHIRLPOOL_ACCOUNT_SIZE = getAccountSize(AccountName.Whirlpool);
```

yugure-orca (approved):
Thanks for the update!
oh... Whirlpool account also has reserved space 😆

now LGTM!

PR #97, opened 2023-04-17
Update SDK to use Anchor 0.27 & other settings for Smart-Router

odcheung:
## Features
- Add lookupTableFetcher into WhirlpoolContext to support lookup table fetching when building more complex transactions
- Update SDK to coral-xyz/anchor 0.27 to support VersionedTransactions
- Faster test-suite execution with faster ticks_per_slot for validator. (~2-5s per test to 5-700ms per test)
- Test-suite now runs the test-cases with metaplex out of the box

## Changes
- Update dep to `coral-xyz/anchor - 0.27.0`, `web3.js - 1.74`, `common-sdk - 0.2.0`
- Add yalc to .gitignore
- Adjust test.validator settings for metaplex cloning & faster ticks (faster tests)
- Add options to `WhirlpoolContext` to define default TransactionBuilder settings

## Test-case Changes
Updated test cases to work with upgraded web3.js & anchor
  - Conformed all test-cases to use `confirmed` commitment using TransactionBuilder default options
  - All test-cases' Provider / Connection uses a default ConfirmOption that is set to use `confirmed`
  - All test-cases uses v0 transactions to test. A few test cases that still relies on the TransactionProcessor stack continues to use legacy transactions
  - Remove usage of anchor.env as it relies on Anchor defaults to `processed` commitment. Use Anchor.local for all
  - Update error matching regex since the messages updated

## Verification
- The upgrade to web3.js caused a lot of test-failures due to the validator now caring about commitment levels
- ran all test-cases with legacy & v0 tests. Confirmed that the override options on the new TransactionBuilder & the changes to test-suite works fine under both settings.

### DIFF
--- a/sdk/src/utils/txn-utils.ts
+++ b/sdk/src/utils/txn-utils.ts
@@ -54,23 +54,20 @@ export async function checkMergedTransactionSizeIsValid(
   }
 }
 
-export function contextToBuilderOptions(
+export function contextOptionsToBuilderOptions(
   opts: WhirlpoolContextOptions
 ): TransactionBuilderOptions | undefined {
-  if (opts) {
-    return {
-      defaultBuildOption: {
-        ...defaultTransactionBuilderOptions.defaultBuildOption,
-        ...opts.userDefaultBuildOptions,
-      },
-      defaultSendOption: {
-        ...defaultTransactionBuilderOptions.defaultSendOption,
-        ...opts.userDefaultSendOptions,
-      },
-      defaultConfirmationCommitment:
-        opts.userDefaultConfirmCommitment ??
-        defaultTransactionBuilderOptions.defaultConfirmationCommitment,
-    };
-  }
-  return undefined;
+  return {
+    defaultBuildOption: {
+      ...defaultTransactionBuilderOptions.defaultBuildOption,
+      ...opts.userDefaultBuildOptions,
+    },
+    defaultSendOption: {
+      ...defaultTransactionBuilderOptions.defaultSendOption,
+      ...opts.userDefaultSendOptions,
+    },
+    defaultConfirmationCommitment:
+      opts.userDefaultConfirmCommitment ??
+      defaultTransactionBuilderOptions.defaultConfirmationCommitment,
+  };
 }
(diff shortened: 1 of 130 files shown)


sdk/src/utils/txn-utils.ts
```diff
@@ -38,13 +42,35 @@ export async function checkMergedTransactionSizeIsValid(
...
+    const size = await merged.txnSize({
+      latestBlockhash,
+    });
     return true;
   } catch (e) {
     return false;
   }
 }
+
+export function contextToBuilderOptions(
```
yugure-orca, line 57: nit(naming): context**Options**ToBuilderOptions

sdk/src/utils/txn-utils.ts
```diff
@@ -38,13 +42,35 @@ export async function checkMergedTransactionSizeIsValid(
...
+  opts: WhirlpoolContextOptions
+): TransactionBuilderOptions | undefined {
+  if (opts) {
```
yugure-orca, line 60: nit: is there some possibility that ``opts`` becomes undefined or null ?
I think its definition (``opts: WhirlpoolContextOptions``) doesn't allow it becomes to them.

yugure-orca (approved):

yugure-orca:
@odcheung 
> Faster test-suite execution with faster ticks_per_slot for validator. (~2-5s per test to 5-700ms per test)
awesome tuning! 🤯

A side effect was the problem of not allowing time for rewards to occur.
This was the cause of the cases where retries were needed.

A commit to include sleep has been added.
I also put sleep in test cases where it is not an error but it is undesirable to test in a situation where rewards do not occur.

sdk/src/utils/txn-utils.ts
odcheung: ah good catch. I was attempting to make TransactionBuilderOptions a nested partial and opts would populate it optionally, but Typescript made it quite difficult. I've reverted this to a direct conversion.

sdk/src/utils/txn-utils.ts
odcheung: done

odcheung:
> @odcheung
> 
> > Faster test-suite execution with faster ticks_per_slot for validator. (~2-5s per test to 5-700ms per test)
> 
> awesome tuning! 🤯
> 
> A side effect was the problem of not allowing time for rewards to occur. This was the cause of the cases where retries were needed.
> 
> A commit to include sleep has been added. I also put sleep in test cases where it is not an error but it is undesirable to test in a situation where rewards do not occur.

nice! That was what I was missing.

PR #99, opened 2023-04-20
fix testing two hop swaps

10xSebastian:
If two pools are initialized in order to test two hop swaps it fails upon initializing the second pool using `initTestPool` because the token account for the reused token account A has already been created for pool one.

This pull requests reuses an already existing token account thereby enabling testing two pool swaps and initializing two pools in the same test that share the same tokenA. 

Context:
We are submoduling the whirlpools repo to test an integration, reusing provided test initialisers.

### DIFF
--- a/sdk/tests/utils/token.ts
+++ b/sdk/tests/utils/token.ts
@@ -175,12 +175,25 @@ export async function createAndMintToAssociatedTokenAccount(
     return tokenAccount;
   }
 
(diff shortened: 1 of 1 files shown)


odcheung (approved):
lgtm!

MrG9090 (approved):

MrG9090 (approved):

PR #102, opened 2023-05-13
Fix deep copy crash on quote adjustment

odcheung:
Since updatedQuote is not a deep copy, assigning directly to hopQuotes can cause assign to freezed object errors


yugure-orca:
@odcheung 
Is it possible to rebase to the latest commit on main ?

yugure-orca (commented):
Thanks for the rebase.
LGTM👍

philcchen (approved):

PR #109, opened 2023-07-13
Fix WhirlpoolAccountFetcher cyclical dependency

odcheung:
- Cyclical dependency causing some export woes for `buildDefaultAccountFetcher`
- Just moving type definitions around in different files. No logic changes

## Verification
- Used a sample project to import the sdk, verified that the project builds & executes correctly using the new fetcher / buildDefaultFetcher function


yugure-orca (approved):
I have read over the reported problem once again now, and I think the problem may be in the way the dist is created.
The problem is that both network/public/fetcher/ and network/public/fetcher.js exist in the distributed package.  But network/public/fetcher.js should not exist based on current main branch.  

screenshot 2023-07-13 11 42 38

The reason of existing it is our "build" command doesn't remove dist dir first.  So old generated file will left.

So "rimraf" command have been appended to remove dist first.
(rimraf works well in both UNIX and windows)

odcheung:
> 

yeah i think it's a combination of both. I did a manual clean of dist to get rid of the duplicated fetcher issue on my local env, then ran into the cyclical dep issue as a run time error.  (index was exporting impl, but impl was referencing index)

adding rimaf is a nice addition!

PR #111, opened 2023-07-25
Add getAllWhirlpoolAccounts util method and fetcher.populatePools

odcheung:
Prequsite review orca-sdk - https://github.com/orca-so/orca-sdks/pull/55

- Add back the old `listPoolsWithParams` method that was missed in the previous refactor
- Add `populatePools` method to the fetcher to allow users to populate the cache with getAllWhirlpoolAccounts

## Verification
- Since gPA calls are not really unit-testable, verified using a test script to call the gPA call and populate the fetcher


sdk/src/network/public/fetcher/fetcher-impl.ts
```diff
@@ -151,4 +152,8 @@ export class WhirlpoolAccountFetcher implements WhirlpoolAccountFetcherInterface
   ): Promise<ReadonlyMap<string, PositionBundleData | null>> {
     return this.fetcher.getAccounts(addresses, ParsablePositionBundle, opts);
   }
+  populatePools(pools: [Address, WhirlpoolData][]): void {
+    const poolMap = new Map(pools.map(([address, pool]) => [AddressUtil.toString(address), pool]));
+    this.fetcher.populateAccounts(poolMap, ParsableWhirlpool, Date.now());
+  }
```
yugure-orca, line 158: Is there a specific reason to implement this feature for WhirlpoolData only ? 🤔

I think this feature is for advanced users who control the cache, and I don't think we need to move in the direction of adding populateX for each data type.

Can we generalize this feature like as follows ? 
```
  populateAccounts<T extends WhirlpoolSupportedTypes>(
    accounts: ReadonlyMap<string, T | null>,
    parser: ParsableEntity<T>,
    now: number
  ): void {
    this.fetcher.populateAccounts(accounts, parser, now);
  }
```

sdk/src/network/public/fetcher/fetcher-utils.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Retrieve a list of cached whirlpool addresses and accounts filtered by the given params using
+ * getProgramAccounts.
+ * @category Fetcher
+ *
+ * @param connection The connection to use to fetch accounts
+ * @param programId The Whirlpool program to search Whirlpool accounts for
+ * @param configId The {@link WhirlpoolConfig} account program address to filter by
+ * @returns tuple of whirlpool addresses and accounts
+ */
+export async function getAllWhirlpoolAccounts({
```
yugure-orca, line 17: I don't have strong opinion about it, but I think "ByConfig" could be appended to the name.
(To make sure that the name can be used with more ByX)
WDYT ?

yugure-orca (commented):
Thanks for the re-implementation of this feature!

sdk/src/network/public/fetcher/fetcher-utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  const parsedAccounts: [Address, WhirlpoolData][] = [];
+  accounts.forEach(({ pubkey, account }) => {
+    const parsedAccount = ParsableWhirlpool.parse(pubkey, account);
+    invariant(!!parsedAccount, `could not parse whirlpool: ${pubkey.toBase58()}`);
+    parsedAccounts.push([pubkey, parsedAccount]);
+  });
+
+  return new Map(parsedAccounts.map(([address, pool]) => [AddressUtil.toString(address), pool]));
```
yugure-orca, line 47: nit: does ``parsedAccounts: [string, WhirlpoolData][] = []`` simplify the code ?

sdk/src/network/public/fetcher/fetcher-types.ts
```diff
@@ -209,4 +209,12 @@ export interface WhirlpoolAccountFetcherInterface {
...
     opts?: WhirlpoolAccountFetchOptions
   ): Promise<ReadonlyMap<string, PositionBundleData | null>>;
+
+  /**
+   * Populate the fetcher's cache with the given {@link WhirlpoolsData} accounts
+   * @param pools The map of addresses to on-chain account data
+   * @param parser The {@link ParsableEntity} instance to parse the accounts
+   * @param now The current timestamp to use for the cache
+   */
+  populateCache<T extends WhirlpoolSupportedTypes>(pools: ReadonlyMap<string, T>, parser: ParsableEntity<T>, now: number): void;
```
yugure-orca, line 219: nit: ``pools`` --> `accounts`` or ``cache`` ? (comment and code)

sdk/src/network/public/fetcher/fetcher-impl.ts
```diff
@@ -151,4 +152,7 @@ export class WhirlpoolAccountFetcher implements WhirlpoolAccountFetcherInterface
   ): Promise<ReadonlyMap<string, PositionBundleData | null>> {
     return this.fetcher.getAccounts(addresses, ParsablePositionBundle, opts);
   }
+  populateCache<T extends WhirlpoolSupportedTypes>(pools: ReadonlyMap<string, T>, parser: ParsableEntity<T>, now = Date.now()): void {
```
yugure-orca, line 155: ``pools`` --> `accounts`` or ``cache`` ?

yugure-orca (approved):
Thanks for the updates!
LGTM with some nits.

Could you update docs with ``yarn run docs`` before the merge. 🙏

sdk/src/network/public/fetcher/fetcher-utils.ts
odcheung: I don't think it does. you still have to convert the Pubkey to string somewhere. 
It's either in the accounts.forEach loop or in the Map init loop.

PR #117, opened 2023-10-08
feat: Allow irrelevant tick arrays to be skipped

Arrowana:
2 sequential swap transactions are constructed at the same time by 2 different users, they are both sent. They both swap with the same a_to_b

The ascii represents the tick first, second and third tick arrays to illustrate what happens currently, ci: current index

Both users see this, and therefor construct a transaction containing those tick arrays
```
|______________ci_|______________|__________|
```

Swap are sent and processed
But the tx of the first user causes ci to move over to next tick array

```
|_______________|_ci____________|___________|
```

Second user tx will fail because `search_in_range` returns false.
I believe we can simply skip irrelevant tick arrays as they are not mutated in any way. This will resolve this case.

I might be missing some peculiar details.

Additional: If this is possible it might be interesting to explore providing one tick array before when near a "start" of a tick array, depending of a_to_b, to increase reliability further when moving near the boundaries.

Note: Raydium CLMM seems to support something similar, advancing in tick arrays until the first one is found https://github.com/raydium-io/raydium-clmm/blob/6e4639f7133a8852068d2d473c263f907b69cd4a/programs/amm/src/instructions/swap.rs#L205-L215

### DIFF
--- a/programs/whirlpool/src/state/tick.rs
+++ b/programs/whirlpool/src/state/tick.rs
@@ -157,6 +157,12 @@ impl Default for TickArray {
     }
 }
 
+pub enum NextInitTickIndex {
+    TickIndex(i32),
+    NoInitializedTickIndex,
+    NotInSearchRange,
+}
+
 impl TickArray {
     pub const LEN: usize = 8 + 36 + (Tick::LEN * TICK_ARRAY_SIZE_USIZE);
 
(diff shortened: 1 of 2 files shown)


Yihwan:
hi @Arrowana thanks for the PR! this is something we're thinking of incorporating into new programs in the future, but we're holding off on all but the most critical upgrades to the existing program to avoid unintentionally increasing the surface area for bugs/exploits.

PR #121, opened 2023-11-10
add optional position mint params

fuzzyyeti:
Add an optional parameter to specify the positionMint instead of generating it internally. This allows the ability to assign an ephemeral keypair that can be saved for later execution in the case of a multi-sig wallet.

### DIFF
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -95,7 +95,7 @@ export class WhirlpoolImpl implements Whirlpool {
     liquidityInput: IncreaseLiquidityInput,
     wallet?: Address,
     funder?: Address,
-    positionMint?: Keypair
+    positionMint?: PublicKey
   ) {
     await this.refresh();
     return this.getOpenPositionWithOptMetadataTx(
@@ -103,7 +103,9 @@ export class WhirlpoolImpl implements Whirlpool {
       tickUpper,
       liquidityInput,
       !!wallet ? AddressUtil.toPubKey(wallet) : this.ctx.wallet.publicKey,
-      !!funder ? AddressUtil.toPubKey(funder) : this.ctx.wallet.publicKey
+      !!funder ? AddressUtil.toPubKey(funder) : this.ctx.wallet.publicKey,
+      false,
+      positionMint
     );
   }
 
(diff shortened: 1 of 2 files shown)


sdk/src/impl/whirlpool-impl.ts
```diff
@@ -111,8 +112,8 @@ export class WhirlpoolImpl implements Whirlpool {
     tickUpper: number,
     liquidityInput: IncreaseLiquidityInput,
     sourceWallet?: Address,
-    positionWallet?: Address,
```
fuzzyyeti, line 114: This was an unused parameter. Currently, the parameter labeled funder in the interface was being dropped.

This will change the behavior for any client putting a non-default funder here, so we could consider putting a placeholder in this position. I don't think the current code won't behave the way the user is expecting though.

fuzzyyeti: To test this I added the SDK as a dependency to the liquidity terminal, generated a keypair to pass in when creating a position, and verified it matched the token mint on the block explorer.

yugure-orca: good catch.
I agree that unused parameters that do not match the interface definition should be removed as implementation bug.

yugure-orca (commented):
### return value of getEphemeralSigners
Could you check the return value of getEphemeralSigners method ?

As far as I read the documentation, the return value is Pubkey.
If it is a PDA, the private key should not exist theoretically.

### Checking if a method returns a different value each time
It is explained that this function is for use with createAccount, so we should be able to create an account.

However, we don't want the same value returned over and over again. Could you please ask with the Squads team to see if it changes each time or if it can be changed.

If we explain to them what we want to do, they will probably have some good advice.

sdk/src/whirlpool-client.ts
```diff
@@ -1,6 +1,6 @@
 import { Address } from "@coral-xyz/anchor";
 import { Percentage, TransactionBuilder } from "@orca-so/common-sdk";
-import { PublicKey } from "@solana/web3.js";
+import {Keypair, PublicKey } from "@solana/web3.js";
```
yugure-orca, line 3: nit: ``Keypair`` is no longer used.

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -276,14 +281,15 @@ export class WhirlpoolImpl implements Whirlpool {
       `upper tick ${tickUpper} is not an initializable tick for tick-spacing ${whirlpool.tickSpacing}`
     );
 
-    const positionMintKeypair = Keypair.generate();
+    const positionMintKeypair =  Keypair.generate();
```
yugure-orca, line 284: super nit: This change is not needed (" " added)

yugure-orca (approved):
LGTM with nit.

yugure-orca:
Note: The ephemeral signer is the same each time you ask for it 👉until it is used.👈

PR #122, opened 2023-11-13
bump version and update docs

fuzzyyeti:
I ran "yarn run docs" and committed the regenerated docs, then bumped from 0.11.6 to 0.11.7.


yugure-orca:
replaced by: https://github.com/orca-so/whirlpools/pull/123

PR #125, opened 2023-12-17
Update sdk README.md to use coral/anchor instead of serum/anchor

PrasoonPratham:
Update dependency in readme for SDK.


yugure-orca (approved):
Thanks! 🙏

PR #128, opened 2024-01-03
Reduced close position transaction builder to a single transaction

wjthieme:
Right now the closePosition transaction builder potentially creates two transactions if ATAs need to be created. Tx one contains the CreateATA instructions and tx two contains the instructions related to closing the position.

This PR changes this logic and puts all instructions for closePosition in a single transaction (like we were already doing for collectFees and collectRewards).

@yugure-orca opted for not changing the instruction to CreateATAIdempotent because the logic is fetching the accounts either way. If we want that improvement there is a bigger change needed in `common-sdk`.

### DIFF
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -584,15 +584,24 @@ export class WhirlpoolImpl implements Whirlpool {
 
     txBuilder.addInstruction(positionIx);
 
-    const txBuilders: TransactionBuilder[] = [];
-
-    if (!tokenAccountsTxBuilder.isEmpty()) {
-      txBuilders.push(tokenAccountsTxBuilder);
+    if (tokenAccountsTxBuilder.isEmpty()) {
+      return [txBuilder]
     }
 
-    txBuilders.push(txBuilder);
+    // This handles an edge case where the instructions are too
+    // large to fit in a single transaction and we need to split the
+    // instructions into two transactions.
+    if (
+      !tokenAccountsTxBuilder.isEmpty() &&
+      affiliatedMints.hasNativeMint &&
+      resolveAtaIxs.length >= 2 &&
+      this.ctx.opts.accountResolverOptions?.createWrappedSolAccountMethod != "ata"
+    ) {
+      return [tokenAccountsTxBuilder, txBuilder]
+    }
 
-    return txBuilders;
+    tokenAccountsTxBuilder.addInstruction(txBuilder.compressIx(false));
+    return [tokenAccountsTxBuilder]
   }
 
   private async refresh() {
--- a/sdk/src/impl/whirlpool-impl.ts
+++ b/sdk/src/impl/whirlpool-impl.ts
@@ -1,6 +1,7 @@
 import { Address, BN, translateAddress } from "@coral-xyz/anchor";
 import {
   AddressUtil,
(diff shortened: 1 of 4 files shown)


yugure-orca (commented):
Thanks for the improvement.

I found the following description at the initial implementation PR.

https://github.com/orca-so/whirlpools/pull/58
>  If max TX size scenario for close position fails in the tests, split it into multiple TXs (maybe use the dynamic TX splitting logic odcheung suggested)

Could you confirm (test) max TX size scenario ?
If the case is included in test cases, you can check by ``anchor test``.

wjthieme:
I got it to work all in one transaction! There was one more thing I needed to add to get it to fit in one transaction. Before this change the transaction was using an ephemeral wsol token account created with the `keypair` method, this needs three instructions (create, initialize and close to unwrap at the end). I've added a new method called `ata` which can do the same thing in two instructions (https://github.com/orca-so/orca-sdks/pull/66) (createATA and close to unwrap at the end) provided you don't need to wrap any SOL tokens (which with closePosition we don't).

There is one edge case where the `closePosition` will fail and I've just covered that by throwing an error (since it is very unlikely):
* All 5 ATAs need to be created
* One of the ATAs is wSOL
* wSOL creation method is not `ata`

wjthieme:
https://github.com/orca-so/orca-sdks/pull/66 is not deployed yet so that needs to happen first

wjthieme:
@yugure-orca added the new approach with combining the txBuilders if possible which most of the cases it is. Only in the following cases it is not possible and we still need two transactions. If all these conditions are met:
* Method is not “ata”
* Contains wSOL
* Need to create 2 or more ATAs

sdk/src/impl/whirlpool-impl.ts
```diff
@@ -584,15 +584,24 @@ export class WhirlpoolImpl implements Whirlpool {
...
-    txBuilders.push(txBuilder);
+    // This handles an edge case where the instructions are too
+    // large to fit in a single transaction and we need to split the
+    // instructions into two transactions.
+    if (
+      !tokenAccountsTxBuilder.isEmpty() &&
+      affiliatedMints.hasNativeMint &&
+      resolveAtaIxs.length >= 2 &&
+      this.ctx.opts.accountResolverOptions?.createWrappedSolAccountMethod != "ata"
+    ) {
```
yugure-orca, line 599: Is there reason to avoid using ``checkMergedTransactionSizeIsValid`` to check if 2 transaction builder can be merged within tx size limit ?

https://github.com/orca-so/whirlpools/blob/main/sdk/src/instructions/composites/collect-all-txn.ts#L174-L183

To reduce additional fetch of blockhash, we can use this dummy because it is used as "padding data".
https://github.com/orca-so/orca-sdks/blob/main/packages/common-sdk/src/web3/transactions/constants.ts#L10

yugure-orca (approved):
Thanks for the update!

nit: I think it is one idea to write the edge case in specific terms.
On the other hand would explicitly using checkMergedTransactionSizeIsValid make clear the intent to merge TransactionBuilders into one ?

In any case, agreed with Merge. 👍

sdk/src/impl/whirlpool-impl.ts
wjthieme: Nice. Yep measurement blockhash is what I needed since I was hesitant of using because it is async

PR #133, opened 2024-02-12
Fix PositionUtil.getPositionStatus bug

yugure-orca:
Ticket: https://app.asana.com/0/1202413563699470/1206467231374946/f

In https://github.com/orca-so/whirlpools/pull/126, PositionUtil.getPositionStatus have been updated to return PositionStatus.BelowRange when ``tickCurrentIndex == tickLowerIndex`` is true.

But when ``toSqrtPrice(tickLowerIndex) < sqrtPrice < toSqrtPrice(tickLowerIndex + 1)`` is true, position status should be InRange.

In this PR, the condition have been reverted.
To avoid zero-div in increaseLiquidityQuoteByInputToken calc, new strict method ``PositionUtil.getStrictPositionStatus`` have been added.  "Strict" means that the use of sqrtPrice and the clarification of the behavior when sqrtPrice is on the boundary.

### DIFF
--- a/sdk/tests/sdk/whirlpools/quote/decrease-liquidity-quote.test.ts
+++ b/sdk/tests/sdk/whirlpools/quote/decrease-liquidity-quote.test.ts
@@ -1,0 +1,61 @@
+import *  as assert from "assert";
+import { PriceMath, decreaseLiquidityQuoteByLiquidityWithParams } from "../../../../src";
+import { BN } from "bn.js";
+import { PublicKey } from "@solana/web3.js";
+import { Percentage } from "@orca-so/common-sdk";
+
+describe("edge cases", () => {
+  const tokenMintA = new PublicKey("orcaEKTdK7LKz57vaAYr9QeNsVEPfiu6QeMU1kektZE");
+  const tokenMintB = new PublicKey("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v");
+
+  it("sqrtPrice on lower bound", async () => {
+    const quote = decreaseLiquidityQuoteByLiquidityWithParams({
+      liquidity: new BN(100000),
+      sqrtPrice: PriceMath.tickIndexToSqrtPriceX64(0),
+      tickLowerIndex: 0,
+      tickUpperIndex: 64,
+      tickCurrentIndex: 0,
+      slippageTolerance: Percentage.fromFraction(0, 100),
+    });
+
+    assert.ok(quote.tokenEstA.gtn(0));
+    assert.ok(quote.tokenEstB.isZero());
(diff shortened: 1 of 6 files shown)


sdk/tests/sdk/whirlpools/quote/decrease-liquidity-quote.test.ts
```diff
@@ -0,0 +1,61 @@
...
+      tickUpperIndex: 64,
+      tickCurrentIndex: 64,
+      slippageTolerance: Percentage.fromFraction(0, 100),
+    });
+
+    assert.ok(quote.tokenEstA.isZero());
+    assert.ok(quote.tokenEstB.gtn(0));
+    assert.ok(quote.tokenMinA.isZero());
+    assert.ok(quote.tokenMinB.gtn(0));
+  });
```
wjthieme, line 59: Nice!

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      sqrtPrice: PriceMath.tickIndexToSqrtPriceX64(0),
+      tokenMintA,
+      tokenMintB,
+      tickLowerIndex: 0,
+      tickUpperIndex: 64,
+      tickCurrentIndex: 0,
+      slippageTolerance: Percentage.fromFraction(0, 100),
+    });
+
+    assert.ok(quote.liquidityAmount.isZero());
```
wjthieme, line 24: For my understanding: basically this means you want to deposit x about of tokenB even though with these prices it should be a single-sided token A deposit. Meaning you can try to add as much of tokenB as you want but since it is single-sided token A deposit you'll always get a zero-liquidity quote.

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      sqrtPrice: PriceMath.tickIndexToSqrtPriceX64(1).subn(1),
+      tokenMintA,
+      tokenMintB,
+      tickLowerIndex: 0,
+      tickUpperIndex: 64,
+      tickCurrentIndex: 0,
+      slippageTolerance: Percentage.fromFraction(0, 100),
+    });
+
+    assert.ok(quote.liquidityAmount.gtn(0));
```
wjthieme, line 88: Guess this is the edge case that the old method failed on. This should be a non-zero-liquidity quote but because the other method was using tickIndexes for current price it would get a sqrtDiff of 0. Here you'd get a non-zero sqrtDiff which means we can get a valid quote

wjthieme (approved):
🚢

PR #135, opened 2024-02-16
Price-based slippage calculation for increase liquidity

odcheung:
- Deprecated the old `increaseLiquidityQuoteByInputToken` quote function & added a new `increaseLiquidityQuoteByInputTokenUsingPriceSlippage`
- Adding `increaseLiquidityQuoteByLiquidity` quote function to generate a quote by a set liquidity value
- This new quote functions will determine `tokenMax` based on price movement, rather than the old method of setting the slippage threshold on the token estimation amounts.


sdk/src/quotes/public/increase-liquidity-quote.ts
```diff
@@ -59,7 +61,7 @@ export type IncreaseLiquidityQuote = IncreaseLiquidityInput & { tokenEstA: BN; t
  * @param whirlpool - A Whirlpool helper class to help interact with the Whirlpool account.
  * @returns An IncreaseLiquidityInput object detailing the required token amounts & liquidity values to use when calling increase-liquidity-ix.
  */
-export function increaseLiquidityQuoteByInputToken(
+export function increaseLiquidityQuoteByInputToken_PriceSlippage(
```
yugure-orca, line 64: I think SDK uses lower camel case for names.
Is there strong reason to use underscore here ?
One thing came to mind is `UsingPriceSlippage`.

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote.test.ts
```diff
@@ -1,10 +1,226 @@
...
+        tickCurrentIndex: 0,
+        inputTokenAmount: new BN(100000),
+        isTokenA,
+        slippage
+      }));
+    });
+
+    function testVariation(params: { pTickLowerIndex: number, pTickUpperIndex: number, tickCurrentIndex: number, inputTokenAmount: BN, isTokenA: boolean, slippage: Percentage }) {
+      const { pTickLowerIndex, pTickUpperIndex, tickCurrentIndex, inputTokenAmount, isTokenA } = params;
+      const slippage = Percentage.fromFraction(1, 100);
```
yugure-orca, line 155: Does this override `slippage` param, and always 1% slippage is used ?

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote.test.ts
```diff
@@ -1,10 +1,226 @@
...
+      const quote = increaseLiquidityQuoteByInputTokenWithParams_PriceSlippage({
+        inputTokenAmount,
+        inputTokenMint,
+        sqrtPrice,
+        tokenMintA,
+        tokenMintB,
+        tickLowerIndex: pTickLowerIndex,
+        tickUpperIndex: pTickUpperIndex,
+        tickCurrentIndex,
+        slippageTolerance: Percentage.fromFraction(1, 100),
```
yugure-orca, line 170: `slippage` ?

sdk/src/quotes/public/increase-liquidity-quote.ts
```diff
@@ -109,32 +143,41 @@ export function increaseLiquidityQuoteByInputTokenWithParams(
...
     case PositionStatus.AboveRange:
-      return quotePositionAboveRange(param);
+      return getQuotePositionAboveRange(param);
     default:
       throw new Error(`type ${positionStatus} is an unknown PositionStatus`);
   }
 }
 
-/*** Private ***/
+function adjustPriceForSlippage(
```
yugure-orca, line 156: hmm, since sqrtPrice is a square root, I would think that when this value is multiplied by 4, the actual price would move by a factor of 2. 
How about applying slippage to price instead of sqrtPrice or multiplied by the square of slippage ? 🤔

sdk/src/quotes/public/increase-liquidity-quote.ts
```diff
@@ -101,6 +104,37 @@ export function increaseLiquidityQuoteByInputTokenWithParams(
...
+    sqrtPrice: upperAdjustedSqrtPrice,
+    tickCurrentIndex: upperAdjustedTickCurrentIndex,
+  });
+
+  return {
+    tokenMaxA: BN.max(BN.max(lowerQuote.tokenEstA, upperQuote.tokenEstA), quote.tokenEstA),
+    tokenMaxB: BN.max(BN.max(lowerQuote.tokenEstB, upperQuote.tokenEstB), quote.tokenEstB),
+    tokenEstA: quote.tokenEstA,
+    tokenEstB: quote.tokenEstB,
+    liquidityAmount: quote.liquidityAmount,
```
yugure-orca, line 133: in trade context, "slippage" is used to fail transaction if the amount is beyond the threshold.
And it will work to protect users from sandwitch atack.

I think in the following case tokenEstA will be larger than required and deposit will succeed even if the price move to lower than slippage applied price.

- position range [92, 94] USDC/SOL
- current price: 93
- lowerAdjustedPrice: 93 x 99 = 92.07
- upperAdjustedPrice: 93 x 101 = 93.93
- input token: USDC
- input amount: 100 USDC

lowerQuote.liquidity will be calculated to deposit 100 USDC between [92, 92.07]. It is tight and results higher liquidity.
Higher liquidity will require more SOL distributed to 92.07 ~ 94.
The SOL amount will be larger than required SOL amount at 92.07 with quote.liquidity (smaller than lowerQuote.liquidity)

Since the Whirlpool contract requires a fixed liquidity input, I think we need to have maxA and maxB as the minimum amount of tokens that can achieve "liquidity" even if price moved between [current price - slippage, current price + slippage], WDYT?

yugure-orca (changes requested):
Added some comments 🙏

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote.test.ts
odcheung: removed!

sdk/src/quotes/public/increase-liquidity-quote.ts
odcheung: The new function `increaseLiquidityQuoteByInputTokenUsingPriceSlippage`'s only job now is to infer liquidity from the input token amount and then call `increaseLiquidityQuoteByLiquidityWithParams`.

`increaseLiquidityQuoteByLiquidityWithParams` handles the slippage calculation as well. 

Each new quote function now has it's own set of test suite to verify tokenEst & tokenMax values.

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote-by-liq.test.ts
```diff
@@ -0,0 +1,275 @@
...
+  const { liquidity, sqrtPrice, currentTickIndex, lowerTickIndex, upperTickIndex } = params;
+
+  const upperSqrtPrice = PriceMath.tickIndexToSqrtPriceX64(upperTickIndex);
+
+  if (currentTickIndex >= upperTickIndex) {
+    return ZERO;
+  }
+
+  if (currentTickIndex < lowerTickIndex) {
+    return getTokenAFromLiquidity(liquidity, sqrtPrice, upperSqrtPrice, true);
```
yugure-orca, line 248: ``(liquidity, lowerSqrtPrice, upperSqrtPrice)`` ?

sdk/tests/sdk/whirlpools/quote/increase-liquidity-quote-by-liq.test.ts
```diff
@@ -0,0 +1,275 @@
...
+  getTokenAFromLiquidity,
+  getTokenBFromLiquidity,
+} from "../../../../src/utils/position-util";
+
+const variations = [
+  [Percentage.fromFraction(0, 100), new BN(17733543)] as const,
+  [Percentage.fromFraction(1, 1000), new BN(17733543)] as const,
+  [Percentage.fromFraction(1, 100), new BN(17733543)] as const,
+];
+
```
yugure-orca, line 18: Generally same review comments for increase-liquidity-quote-by-input-token.test.ts 🙏

yugure-orca (commented):
Added some comments 🙏

odcheung:
@yugure-orca - Apologies. I've added the following comment to clarify what each test-case means:

```
// [---P---] = P is the current price & [] is the slippage boundary
// |-------| = Position Boundary
```
I've also cleaned up and modified the test cases to use relative ticks to the current tick rather than absolute ticks. Additional test cases with current tick variations have also been added. These will address your comments w.r.t test comments above. Let me know if I'm missing ssomething.

0% slippage test also has been removed as it doesn't apply to these test cases that tests position & slippage range overlaps.

yugure-orca (approved):
Thanks for updating.
Added some comments but I believe these are not critical comments.

PR #142, opened 2024-03-26
use idempotent on whirlpool.swap (swapAsync)

yugure-orca:
Asana: https://app.asana.com/0/1202413563699470/1206926743112719/f

Resolve double ATA creation issue.
It happens because some users will build and send multiple transactions simultaneously.

## Test
- ``anchor test`` passed
- added new 1 test case to ensure Idempotent is used


wjthieme (approved):
🚢

PR #144, opened 2024-04-18
Add security txt file to the whirlpool program

wjthieme:
See: https://github.com/neodyme-labs/solana-security-txt

This will show up in explorers like solscan.


programs/whirlpool/Cargo.toml
```diff
@@ -23,6 +23,7 @@ thiserror = "1.0"
 uint = {version = "0.9.1", default-features = false}
 borsh = "0.9.1"
 mpl-token-metadata = { version = "1.7", features = ["no-entrypoint"] }
+solana-security-txt = { version = "^1.1.1" }
```
yugure-orca, line 26: how about use `version = "=1.1.1"`.
I trust Neodyme, but I think we may be a little cautious about running macros in lib.rs.
WDYT ?

yugure-orca (commented):
Thanks for the improvement.
Almost LGTM.
Added 1 small question about `Cargo.toml`.

# About the merge timing
I am a little concerned that this will affect the Verifiable Build verification (although I am not sure if this is done regularly).

I think the contract will be upgraded after the following PR is merged, so 👉I think it is better to merge at the same time as the PR#134 merge.👈 Conflicts obviously do not exist.
https://github.com/orca-so/whirlpools/pull/134

programs/whirlpool/Cargo.toml
wjthieme: Yep think pinning is a good idea!

yugure-orca (approved):
Thanks for the update.

will merge with TE 👍

yugure-orca:
https://github.com/orca-so/whirlpools/pull/134 have been merged, time to merge this PR 🙂

PR #148, opened 2024-05-25
add ConfidentialTransferFeeConfig extension support

yugure-orca:
When both `TransferFeeConfig` and `ConfidentialTransferMint` are initialized, `ConfidentialTransferFeeConfig` must be initialized.

https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/mod.rs#L1292

So we need to support `ConfidentialTransferFeeConfig` extension.

Whirlpool uses only non-confidential transfers, so this extension is not used in Whirlpool's operation, but must be allowed to be set to mint.

### DIFF
--- a/sdk/src/utils/public/pool-utils.ts
+++ b/sdk/src/utils/public/pool-utils.ts
@@ -204,13 +204,16 @@ export class PoolUtil {
       return false;
     }
 
-    if (mintWithTokenProgram.freezeAuthority !== null) {
-      return false;
-    }
-
     const tokenBadgePda = PDAUtil.getTokenBadge(ctx.program.programId, whirlpoolsConfig, tokenMintKey);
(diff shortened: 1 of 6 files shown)


sdk/src/utils/public/pool-utils.ts
```diff
@@ -204,14 +204,17 @@ export class PoolUtil {
...
-      return false;
-    }
-
     const tokenBadgePda = PDAUtil.getTokenBadge(ctx.program.programId, whirlpoolsConfig, tokenMintKey);
     const tokenBadge = await ctx.fetcher.getTokenBadge(tokenBadgePda.publicKey);
     const isTokenBadgeInitialized = tokenBadge !== null;
 
+    if (mintWithTokenProgram.freezeAuthority !== null && !isTokenBadgeInitialized) {
+      return false;
+    }
```
yugure-orca, line 213: synced to the contract

philcchen (approved):

PR #149, opened 2024-05-28
Add price-slippage variant to the decreaseLiquidityQuote helper function

odcheung:
- Add a price-slippage variant to `decreaseLiquidityQuoteByLiquidity`
- un-deprecate the token-slippage variant for `increaseLiquidityQuoteByLiquidity` since it's actually a valid calculation option

### DIFF
--- a/sdk/src/quotes/public/decrease-liquidity-quote.ts
+++ b/sdk/src/quotes/public/decrease-liquidity-quote.ts
@@ -3,15 +3,15 @@ import { Percentage, ZERO } from "@orca-so/common-sdk";
 import invariant from "tiny-invariant";
 import { DecreaseLiquidityInput } from "../../instructions";
 import {
+  PositionStatus,
+  PositionUtil,
   adjustForSlippage,
   getTokenAFromLiquidity,
   getTokenBFromLiquidity,
-  PositionStatus,
-  PositionUtil,
 } from "../../utils/position-util";
 import { PriceMath, TickUtil } from "../../utils/public";
-import { Position, Whirlpool } from "../../whirlpool-client";
 import { TokenExtensionContextForPool, TokenExtensionUtil } from "../../utils/public/token-extension-util";
+import { Position, Whirlpool } from "../../whirlpool-client";
 
 /**
  * @category Quotes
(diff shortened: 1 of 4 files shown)


yugure-orca:
looking🫡

sdk/src/quotes/public/decrease-liquidity-quote.ts
```diff
@@ -91,115 +91,126 @@ export function decreaseLiquidityQuoteByLiquidity(
...
+  const tokenMinBExcluded = TokenExtensionUtil.calculateTransferFeeExcludedAmount(tokenMinB, tokenExtensionCtx.tokenMintWithProgramB, tokenExtensionCtx.currentEpoch);
+  const tokenEstBExcluded = TokenExtensionUtil.calculateTransferFeeExcludedAmount(tokenEstB, tokenExtensionCtx.tokenMintWithProgramB, tokenExtensionCtx.currentEpoch);
 
   return {
-    liquidityAmount: liquidity,
-    tokenMinA: tokenMinAExcluded.amount,
-    tokenMinB: ZERO,
-    tokenEstA: tokenEstAExcluded.amount,
-    tokenEstB: ZERO,
+    tokenMinA,
```
yugure-orca, line 113: hmm, if TransferFee is enabled, `tokenMin` and `tokenEst` will be `original amount - transfer fee`.
So original code sounds good to me.
WDYT ?

sdk/src/quotes/public/decrease-liquidity-quote.ts
```diff
@@ -91,115 +91,126 @@ export function decreaseLiquidityQuoteByLiquidity(
...
   const tokenMinBExcluded = TokenExtensionUtil.calculateTransferFeeExcludedAmount(tokenMinB, tokenExtensionCtx.tokenMintWithProgramB, tokenExtensionCtx.currentEpoch);
   const tokenEstBExcluded = TokenExtensionUtil.calculateTransferFeeExcludedAmount(tokenEstB, tokenExtensionCtx.tokenMintWithProgramB, tokenExtensionCtx.currentEpoch);
 
   return {
-    liquidityAmount: liquidity,
-    tokenMinA: tokenMinAExcluded.amount,
-    tokenMinB: tokenMinBExcluded.amount,
-    tokenEstA: tokenEstAExcluded.amount,
-    tokenEstB: tokenEstBExcluded.amount,
+    tokenMinA,
```
yugure-orca, line 179: (same to the above)
I think we need to use `xExcluded.amount` here.

sdk/src/quotes/public/decrease-liquidity-quote.ts
```diff
@@ -91,115 +91,126 @@ export function decreaseLiquidityQuoteByLiquidity(
...
 }
 
-function quotePositionAboveRange(param: DecreaseLiquidityQuoteParam): DecreaseLiquidityQuote {
-  const { tickLowerIndex, tickUpperIndex, liquidity, slippageTolerance: slippageTolerance, tokenExtensionCtx } = param;
+function getTokenEstimatesFromLiquidity(params: DecreaseLiquidityQuoteParam) {
+  const { liquidity, tickLowerIndex, tickUpperIndex, sqrtPrice } = params;
 
-  const sqrtPriceLowerX64 = PriceMath.tickIndexToSqrtPriceX64(tickLowerIndex);
-  const sqrtPriceUpperX64 = PriceMath.tickIndexToSqrtPriceX64(tickUpperIndex);
+  if (liquidity.eq(ZERO)) {
```
yugure-orca, line 196: hmm, `decreaseLiquidityQuoteByLiquidityWithParamsUsingPriceSlippage` has early return path for zero liquidity.
On the other hand, original `decreaseLiquidityQuoteByLiquidityWithParams` doesn't.
So how about adding early return logic for it.
(or simply return zero here 🤔)

yugure-orca (commented):
added some comments related TransferFee excluded amount and zero handling. 🙏

yugure-orca (approved):
Thanks for the update!
LGTM 🚀

PR #156, opened 2024-06-21
update anchor and solana

DemidovVladimir:


Target911:
Hope i get right!

DemidovVladimir:
I am not sure if the update PR I created is the appropriate way but, anyways. I still can not build it locally, but our company need it as we rely on it and making multiple CPI requests against this program. 
Maybe anyone can guide me or collaborate with me in order to upgrade this program to use the latest versions of anchor and   solana, as it is more secure and they made a lot of changes that makes programs more efficient.
Here are some issues I faced during this upgrade:

1.  There are multiple big accounts that does not fit stack anymore, example (Whirlpool), but aligning fields, making it zero_copy and use it as AccountLoader helps. However there are other accounts  that needs to be fixed.
2. Without idl-build = ["anchor-lang/idl-build", "anchor-spl/idl-build"] and overflow-checks = true I can not build anchor I guess from v 0.28, I added it as well.
3. Solana 1.18.17 uses some solana deps which are v2, need to forcefully downgrade at list this one: https://solana.stackexchange.com/questions/14973/error-after-updating-solana-program-to-1-18-17-and-anchor-to-0-30-1
4. Remaining accounts now have lifetimes and this is a problem, because Account::try_from(account_info) does not live long enough: https://solana.stackexchange.com/questions/13310/how-to-use-lifetime-on-context-correctly
5. Use spl-transfer-hook-interface v0.6.3
6. cargo update -p solana-zk-token-sdk@2.0.1 --precise 1.18.17 (https://github.com/coral-xyz/anchor/issues/3044)
Some other fixes can be found in the PR.
I am not sure how necessary they are so I am asking you guys to help me here please.

wjthieme (commented):
Thanks for getting started with this. Updating to 0.30 is definitely on our roadmap. I'm afraid with the current state of the PR already existing whirlpool accounts can no longer be parsed.

I made some minor tweaks to the versions and config but it will probably take a little while before we can fully upgrade to 0.30.

In the meantime if you are just trying to do CPIs (depending on what your use-case is) you might be able to use our cpi crate: https://github.com/orca-so/whirlpool-cpi

wjthieme:
Closing this since it is superseded by https://github.com/orca-so/whirlpools/pull/174

PR #161, opened 2024-07-08
Disallow non-full-range positions on pools with tick spacing >= 2^15

wjthieme:
* Added check to disallow non-full-range positions on pools with tick spacing >= 2^15
* Added tests for `open_position`, `open_position_with_metadata` and `open_bundled_position`

### DIFF
--- a/programs/whirlpool/src/math/tick_math.rs
+++ b/programs/whirlpool/src/math/tick_math.rs
@@ -10,6 +10,8 @@ const BIT_PRECISION: u32 = 14;
 const LOG_B_P_ERR_MARGIN_LOWER_X64: i128 = 184467440737095516i128; // 0.01
 const LOG_B_P_ERR_MARGIN_UPPER_X64: i128 = 15793534762490258745i128; // 2^-precision / log_2_b + 0.01
 
+pub const INFINITY_POOL_SPACING_FLAG: u16 = 32768; // 2^15
+
 /// Derive the sqrt-price from a tick index. The precision of this method is only guarranted
 /// if tick is within the bounds of {max, min} tick-index.
 ///
--- a/programs/whirlpool/src/math/tick_math.rs
+++ b/programs/whirlpool/src/math/tick_math.rs
@@ -10,7 +10,7 @@ const BIT_PRECISION: u32 = 14;
 const LOG_B_P_ERR_MARGIN_LOWER_X64: i128 = 184467440737095516i128; // 0.01
 const LOG_B_P_ERR_MARGIN_UPPER_X64: i128 = 15793534762490258745i128; // 2^-precision / log_2_b + 0.01
 
-pub const INFINITY_POOL_SPACING_FLAG: u16 = 32768; // 2^15
+pub const FULL_RANGE_ONLY_TICK_SPACING_THRESHOLD: u16 = 32768; // 2^15
 
 /// Derive the sqrt-price from a tick index. The precision of this method is only guarranted
 /// if tick is within the bounds of {max, min} tick-index.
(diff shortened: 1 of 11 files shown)


sdk/tests/integration/open_bundled_position.test.ts
```diff
@@ -603,4 +633,20 @@ describe("open_bundled_position", () => {
...
+    const positionBundleInfo = await initializePositionBundle(ctx, ctx.wallet.publicKey);
+    const bundleIndex = 0;
+    await assert.rejects(
+      openBundledPosition(
+        ctx,
+        whirlpoolPda.publicKey,
+        positionBundleInfo.positionBundleMintKeypair.publicKey,
+        bundleIndex,
+        tickLowerIndex,
+        tickUpperIndex
```
yugure-orca, line 646: Is the reason for the error is that the tick is not initializable (not multiple of tickSpacing) ?

programs/whirlpool/src/state/position.rs
```diff
@@ -69,6 +69,16 @@ impl Position {
...
 
+        // On tick spacing >= 2^15, should only be able to open full range positions (infinity pool)
+        if whirlpool.tick_spacing >= 2u16.pow(15) {
+            // Use gt or lt here instead of neq since indexes still need to be on tick spacing
+            if tick_lower_index < MIN_TICK_INDEX
+                || tick_upper_index > MAX_TICK_INDEX
+            {
+                return Err(ErrorCode::InvalidTickIndex.into());
+            }
+        }
```
yugure-orca, line 80: How about changing `2u16.pow(15)`  to constant (pre calculated).

I guess this condition will not reject `[-32768, +32768]` and `[-32768, 0]`, etc.
I think we need to calculate full range lower index & upper index based on tick spacing.

Adding an error code indicating "FullRange only" would be better for the developer to know the cause.

yugure-orca (changes requested):

sdk/tests/integration/open_bundled_position.test.ts
wjthieme: Good point. I think increasing the `tickUpperIndex` to 2^15 should fix this? This would also work in other cases since 2^15 is a multiple of the other tick spacings as well

programs/whirlpool/src/state/position.rs
wjthieme: Ahh yes. Changed this because I didn't want to have to calculate the exact full range indexes to do an eq check (initially thought this would get an identical result but is not the case)

sdk/tests/integration/open_bundled_position.test.ts
yugure-orca: yep, I think 32768 works.

yugure-orca (commented):
Thanks for the update!
Added some comments related "infinity" wording.

Could you add the following on SDK ?
(naming is just idea)

- PoolUtil.isFullRangeForced(tickSpacing): boolean;
- const FULL_RANGE_FORCED_TICK_SPACING_THRESHOLD = 32768; (src/types/public/constants.ts)
- other helpful function to implement UI 🙂

yugure-orca:
@philcchen 
mainly for visibility 🙂

programs/whirlpool/src/math/tick_math.rs
```diff
@@ -10,6 +10,8 @@ const BIT_PRECISION: u32 = 14;
 const LOG_B_P_ERR_MARGIN_LOWER_X64: i128 = 184467440737095516i128; // 0.01
 const LOG_B_P_ERR_MARGIN_UPPER_X64: i128 = 15793534762490258745i128; // 2^-precision / log_2_b + 0.01
 
+pub const INFINITY_POOL_SPACING_FLAG: u16 = 32768; // 2^15
```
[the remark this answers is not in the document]
wjthieme: Yep agree with the naming. Something descriptive is better than the meta name

yugure-orca (approved):
Thanks for the update!
LGTM with 1 last nit.

PR #162, opened 2024-07-09
SparseSwap (SDK)

yugure-orca:
This PR introduces the benefits provided by SparseSwap into SDK.

## We no longer need to stop quote right before uninitialized tickarray
The swapQuoteX function gets the estimate without any concern for the existence of uninitialized TickArrays.
Whirlpool.swap and the swapAsync behind it no longer pre-verify whether a TickArray is initialized or not.

To enjoy the benefit of not having to worry about uninitialized TickArrays by default, the SDK interpolate uninitialized TickArrays with zeroed data in swapQuoteWithParams function.
As a result, functions that use swapQuoteWithParams, such as sqapQuoteByInputToken, also benefit.

- `swapQuoteByInputToken` --> `swapQuoteWithParams`
- `swapQuoteByOutputToken` --> `swapQuoteWithParams`
- `swapQuoteWithParams` (direct use)

## User can insert fallback tickarray if they want
On the other hand, the behavior of supplemental tick arrays and preliminary additions to previous tick arrays are disabled by default. (`UseFallbackTickArray.Never`)
This is to prevent breaking changes on the function interface.

If users want to add a fallback, they must explicitly use `UseFallbackTickArray.Always` or `UseFallbackTickArray.Situational`.

```
export async function swapQuoteByInputToken(
  whirlpool: Whirlpool,
  inputTokenMint: Address,
  tokenAmount: BN,
  slippageTolerance: Percentage,
  programId: Address,
  fetcher: WhirlpoolAccountFetcherInterface,
  opts?: WhirlpoolAccountFetchOptions,
  useFallbackTickArray: UseFallbackTickArray = UseFallbackTickArray.Never, 👈
): Promise<SwapQuote> {
```

### UseFallbackTickArray.Situational
Add a TickArray in the opposite direction as a fallback only if the price is at the end 1/4 of the current TickArray.

```
b to a
[  fallback  ][//p///      ta0   ][      ta1   ][      ta2   ]  ---> fallback will be added
[  fallback  ][//////  p   ta0   ][      ta1   ][      ta2   ]  ---> fallback will NOT be added

a to b
[   ta2      ][   ta1      ][   ta0             //p///][  fallback  ]  ---> fallback will be added
[   ta2      ][   ta1      ][   ta0        p    //////][  fallback  ]  ---> fallback will NOT be added
```

1/4 is fixed value. Devs who want to tune the behavior shoud use `swapQuoteWithparams` and `SwapUtils.getFallbackTickArrayPublicKey` directly.

### How to add fallback tick array
If only one or two tickarrays are used in a trade (ta1=ta2), add it as ta2. (it is safe way because ta2 is dup of ta1)
This way is prefer because V1 users can also benefit.

On the other hand, if three TickArrays are used, it is added to the estimate as `supplementalTickArrays` and used only in the V2 instruction.

`Whirlpool.swap` uses V2 if the pool uses TokenExtension or if `supplementalTickArrays` is not empty.

### DIFF
--- a/sdk/src/utils/public/swap-utils.ts
+++ b/sdk/src/utils/public/swap-utils.ts
@@ -2,14 +2,13 @@ import { Address } from "@coral-xyz/anchor";
 import { AddressUtil, Percentage, U64_MAX, ZERO } from "@orca-so/common-sdk";
 import { PublicKey } from "@solana/web3.js";
 import BN from "bn.js";
-import { WhirlpoolContext } from "../..";
+import { TickUtil, WhirlpoolContext } from "../..";
 import {
   WhirlpoolAccountFetchOptions,
   WhirlpoolAccountFetcherInterface,
 } from "../../network/public/fetcher";
 import {
   MAX_SQRT_PRICE,
-  MAX_SWAP_TICK_ARRAYS,
   MIN_SQRT_PRICE,
   SwapInput,
   SwapParams,
@@ -20,8 +19,8 @@ import { Whirlpool } from "../../whirlpool-client";
 import { adjustForSlippage } from "../math/token-math";
 import { PDAUtil } from "./pda-utils";
 import { PoolUtil } from "./pool-utils";
-import { TickUtil } from "./tick-utils";
 import { SwapDirection, TokenType } from "./types";
+import { TickArrayAddress, buildZeroedTickArray, getTickArrayPublicKeysWithStartTickIndex } from "../swap-utils";
 
 /**
  * A request to fetch the tick-arrays that a swap may traverse across.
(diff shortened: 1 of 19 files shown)


sdk/tests/sdk/whirlpools/swap/swap-array.test.ts
```diff
@@ -170,7 +302,7 @@ describe("swap arrays test", () => {
       adjustForSlippage(quote.estimatedAmountOut, slippageTolerance, false).toString()
     );
     assert.equal(quote.estimatedAmountIn.toString(), tradeAmount);
-    assert.doesNotThrow(async () => await (await whirlpool.swap(quote)).buildAndExecute());
+    await assert.doesNotReject(async () => await (await whirlpool.swap(quote)).buildAndExecute());
```
yugure-orca, line 305: for async fn, we should use `doesNotReject`.

sdk/src/utils/public/swap-utils.ts
```diff
@@ -179,20 +198,39 @@ export class SwapUtils {
...
   }
 
+  /**
+   * Given a set of tickArrays, interpolate the tickArrays with zeroed tick data if they are not initialized.
+   * 
+   * @param whirlpoolAddress - PublicKey of the whirlpool to swap on.
+   * @param tickArrays - Fetched tickArrays to interpolate.
+   * @returns An array of TickArray objects with zeroed tick data if they are not initialized.
+   */
+  public static interporateUninitializedTickArrays(
```
philcchen, line 223: nit: interpolate

philcchen (approved):

sdk/src/utils/public/swap-utils.ts
yugure-orca: oh...🤦‍
Good catch! 🙏

PR #164, opened 2024-07-10
Create monorepo using nx build system

wjthieme:
Hi @yugure-orca! Thinking about v1 of the ts sdk and figured it would be a good idea to set up the repo as a monorepo. This allows us to more easily have the v0.x.x version of the ts sdk together with a new version (which will use new web3.js). This is required because we will still need the v0 sdk because the integration test rely on them. I’ve set it up in the following way and made a couple dev-ex improvements as well:

* Set up the repo using NX (similar to turbo but imo better) so that local deps can be automatically resolved and built.
* Set up several packages already such as legacy (sdk), scripts, programs, etc.
* Added gh actions so that tests can be automatically run on PRs (integration tests are disabled on gh actions because they take 2h+ to run).
* Added `CODEOWNERS` file so that reviews will be automatically requested (and required) when changes are made to certain files.
* Added a couple of github community files such as issue/pr templates
* There was a typedocs for the v0 ts sdk but don’t think it was actually published anywhere. We can add it back if needed.
* Updated yarn to the latest version
* Added CodeQL code scanning and dependabot
* Added lint staged pre-commit hook (but not yet enabled which I will do in a separate PR afterwards)

Let me know what you think!


wjthieme:
Was expecting tests to 'only' take like 30mins but 2h+ is a little bit long. I am considering disabling the integration tests in gh actions until we switch to something like bankrun

wjthieme:
Since this PR is already quite large I will:
* Remove the lint-staged changes here and do that in a separate PR
* Make sure sdk/legacy package and tests stay pretty much identical (apart from the rename/move). Ill just skip 'legacy' integration test in gh actions but we can always still run them manually (like we are doing now).

yugure-orca:
@wjthieme 
hmm, which project issue is this change for ?

It would be nice to see this after the Sparse Swap code audit and merge is done.

I would like to double-check any changes to the Rust code, even if they are mechanical.
So I think it would be good to have a PR for cargo fmt only, for ease of review. 🤔

wjthieme:
> hmm, which project issue is this change for ?

TS SDK v2

> It would be nice to see this after the Sparse Swap code audit and merge is done.

Yeah was not planning on blocking SparseSwap in any way with this. Just thinking about how we can set up the repo in a more dev-friendly way and allow us to keep the old ts sdk together with development of the new ts sdk.

> I would like to double-check any changes to the Rust code, even if they are mechanical.
So I think it would be good to have a PR for cargo fmt only, for ease of review. 🤔

Yeah I'm removing all formatting changes from this PR

wjthieme:
Integration test result:
```
$yarn test legacy/sdk --output-style stream
2122 passing (36m)
```

yugure-orca:
looking 🫡

yugure-orca (commented):
> There was a typedocs for the v0 ts sdk but don’t think it was actually published anywhere. We can add it back if needed.

It is publicly published as GitHub Pages: https://orca-so.github.io/whirlpools/
We need to keep it (of course structure change is acceptable)
link: https://github.com/orca-so/whirlpools?tab=readme-ov-file#generate-typedoc

README.md
```diff
@@ -1,95 +1,104 @@
...
 
-## Installation
+* Install NodeJS and gcc-12 using `brew install node gcc@12`.
+* Add gcc-12 headers to your cpath using `export CPATH="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"`.
+* Install Rust lang using `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`.
+* Install the Solana CLI using `curl -sSfL https://release.solana.com/v1.17.22/install | sh`.
+* Add the Solana CLI to your path using `export PATH="/Users/$(whoami)/.local/share/solana/install/active_release/bin:$PATH"`.
+* Install Anchor version manager using `cargo install --git https://github.com/coral-xyz/anchor avm --locked --force`.
+* Install the latest Anchor version using `avm install 0.29.0 && avm use 0.29.0`.
+* Clone this repository using `git clone https://github.com/jewl-app/.github`.
```
yugure-orca, line 54: Is jewl-app an example used to describe this?

README.md
```diff
@@ -1,95 +1,104 @@
...
 
-## Run Typescript tests via local validator
+This repository is split up into sevaral parts. The following is a (non-exhaustive) list of the components and their purpose.
 
-In the whirlpools/sdk folder, run:
+* **`/programs/*`** - Rust programs that are deployed on Solana.
+* **`/typescript/*`** (WIP) - Typescript SDKs for interacting with the programs.
+* **`/rust/*`** (WIP) - Rust SDKs for interacting with the programs.
+* **`/docs/*`** (WIP) - Documentation for the programs and SDKs.
+* **`/legacy/*`** - Legacy Typescript SDKs and integration tests.
```
yugure-orca, line 77: I believe that `/typescript`, `/rust` and `/legacy`  are all SDK and only `/program` is contract.
So WDYT the following:

- /docs
- /program
- /sdk
    - /legacy
    - /typescript
    - /rust

yugure-orca (commented):
@wjthieme 
Thanks for the hard work.

Could you clearly described what is tested after the change of structure on PR. 🙏
I would like to confirm the following are tested.

- cargo test on program
- anchor test (all existing test cases should be covered)
- yarn run build on legacy sdk and its contents is same to the latest package.

wjthieme:
> It is publicly published as GitHub Pages: https://orca-so.github.io/whirlpools/
We need to keep it (of course structure change is acceptable)
link: https://github.com/orca-so/whirlpools?tab=readme-ov-file#generate-typedoc

Yep will make sure to get this back in.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - type: textarea
+  id: expected
+  attributes:
+    label: Expected behavior
+    description: A clear and concise description of what was supposed to happen.
+    placeholder: The following things were supposed to happen.
+  validations:
+    required: true
+  - type: input
+  id: browser
```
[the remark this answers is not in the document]
wjthieme: This was copy pasted from another repo I had lying around. Will remove

.gitignore
```diff
@@ -10,3 +10,13 @@ sdk/node_modules
 .vscode/
 **/.yalc/
 **/yalc.lock
+dist
+artifacts
```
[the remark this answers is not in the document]
wjthieme: `artifacts` are now included in the build step of the legacy sdk. This way it will always have the latest version and we don't have to manually copy them over. Because of them being automatically copied over during build we don't have to include in git anymore

README.md
wjthieme: Think this is similar to how anza does their structure. I am not the biggest fan of the extra nesting since `legacy`, `typescript` and `rust` each contain multiple package. If you feel strongly about this one I can change

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,20 @@
+name: Dependabot
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+
+  automerge:
```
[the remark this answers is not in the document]
yugure-orca: Thanks for the explanation. 🙏
it makes sense!

README.md
yugure-orca: I think `/sdk` is immediately recognisable as an SDK to devs who come looking for an SDK.
The location of the SDK is clear if they come via the Developer Portal, and the README.md is also useful. However, it would be nice if the structure itself is descriptive, as there are many cases where people do not read the documentation (including me).😂

If this change would significantly affect subsequent PRs, I agree to merge this PR as-is for the time being, as it can be changed after all the current PRs have been merged.

wjthieme: We can also go for 'legacy-sdk', 'rust-sdk' and 'ts-sdk' so we don't have the extra nesting. What do you think?

.gitignore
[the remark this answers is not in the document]
yugure-orca: it makes sense.
Is there an edge-case that 1) clone repo 2) edit contract 3) anchor test  (no artifact?) .
If so how about adding step to instllation process to build SDK package at least once.

README.md
wjthieme: But that might also be a little weird cause in #172 it would then add 'sdk/rust/sdk' and 'sdk/rust/client' or 'rust-sdk/sdk' etc.

.gitignore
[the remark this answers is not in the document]
wjthieme: If you run 'yarn test' instead of 'anchor test' it resolves and builds all dependencies correctly

README.md
yugure-orca: @wjthieme 
Yep, I agree that `x-sdk` naming is nice alternative.
Let's go with it 🚀

yugure-orca (approved):
Thanks for the update.
LGTM. 
I left some additional comments, but those are not large.

.gitignore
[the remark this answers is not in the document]
yugure-orca: > yarn test
noted! 🫡

PR #165, opened 2024-07-17
Add full-range-only tick spacing to `ORCA_SUPPORTED_TICK_SPACINGS`

wjthieme:


yugure-orca (approved):
LGTM.
nit: bump version on package.json ...?

Could you merge the PR "after" fee tier initialization. 🙏
(I believe UI doesn't use this constant and has own one)

PR #172, opened 2024-08-05
Generate rust and ts clients using kinobi

wjthieme:
Kinobi requires `address` field in Anchor idl which Anchor only added in v0.30. This causes the build to fail. I'm updating to anchor v0.30 in #174.

This PR requires account `DISCRIMINATORS` to be exposed by Kinobi. Opened a PR to expose them: https://github.com/kinobi-so/kinobi/pull/163


yugure-orca (approved):
I think the project is in very early stage and code looks good.
In the following PR, is it possible to distinguish between tool generated code and not-generated code ? (in PR comment etc)

Based on our conversation in slack, I believe that new sdk package will not depend on large packages. 🙂

wjthieme:
> is it possible to distinguish between tool generated code and not-generated code

Tool-generated code is gitignored. The build commands create all the tool-generated code on demand. Any code you see in the PRs is written code.

wjthieme:
> Based on our conversation in slack, I believe that new sdk package will not depend on large packages.

Yep in ts-sdk/client package.json you can see all the required deps (new web3.js and no umi or anchor-js). Same goes for cargo.toml file in rust-sdk/client you can see that anchor is optional

ts-sdk/client/src/gpa/tokenBadge.ts
```diff
@@ -0,0 +1,41 @@
...
+      offset: 40n, bytes: getBase58Decoder().decode(getAddressEncoder().encode(address)), encoding: "base58"
+    }
+  } as TokenBadgeFilter;
+}
+
+export async function fetchAllTokenBadgeWithFilter(rpc: Rpc<GetProgramAccountsApi>, ...filters: TokenBadgeFilter[]): Promise<Account<TokenBadge>[]> {
+  const discriminator = getBase58Decoder().decode(TOKEN_BADGE_DISCRIMINATOR);
+  const discriminatorFilter: GetProgramAccountsMemcmpFilter = {
+    memcmp: {
+      offset: 62n,
```
yugure-orca, line 30: hmm, 0n ?

wjthieme: Good catch! Let me see if I can adjust the tests to that it also catches this filter

wjthieme:
Test failures are unrelated so will continue with merging

PR #197, opened 2024-08-12
Add auto-generated docs to the monorepo

wjthieme:
Goal is to deploy them to github pages using GH actions. They are all built on-demand. The actual docs part is not filled out but most important is that we have parity with the current docs (legacy-sdk) which it does

Screenshot 2024-08-15 at 5 30 15 PM

Screenshot 2024-08-15 at 5 40 34 PM


yugure-orca (commented):
maybe noob question.
After the merge of this PR, can we see the typedoc immediately on github wiki ?

Because this PR removes almost all file in docs, I wonder if the current TypeDoc will be removed.
current TypeDoc: https://orca-so.github.io/whirlpools/

wjthieme:
> After the merge of this PR, can we see the typedoc immediately on github wiki ?

I'll manually deploy to wiki so we can see what it looks like. Theoretically it should auto-update on push with github actions but haven't tested that yet.

After this PR is merged I'll disable the github pages since that will then no longer work.

wjthieme:
Looks like github wiki is pretty limited in what it can do with linking pages together. Let me investigate if there is a better alternative

wjthieme:
Used Docusaurus to create a place for docs and guides. This we will then still deploy to GH pages. Might be nice to add an `orca.so` subdomain to it as well at some point

yugure-orca (approved):
LGTM🚀

PR #198, opened 2024-08-12
Create a helper function for creating splash pools

wjthieme:
Add a helper function for creating splash pools. Some people have complained that it is not clear how to create a splash pool from the UI. This PR is a short term solution for this. In v2 sdk we should think of a better way to set this up and make the devex better

### DIFF
--- a/legacy-sdk/whirlpool/src/whirlpool-client.ts
+++ b/legacy-sdk/whirlpool/src/whirlpool-client.ts
@@ -97,6 +97,24 @@ export interface WhirlpoolClient {
    * @param whirlpoolConfig the address of the whirlpool config
    * @param tokenMintA the address of the token A
    * @param tokenMintB the address of the token B
+   * @param initialTick the initial tick that the pool is set to (derived from initial price)
+   * @param funder the account to debit SOL from to fund the creation of the account(s)
+   * @return `poolKey`: The public key of the newly created whirlpool account. `tx`: The transaction containing instructions for the on-chain operations.
+   * @throws error when the tokens are not in the canonical byte-based ordering. To resolve this, invert the token order and the initialTick (see `TickUtil.invertTick()`, `PriceMath.invertSqrtPriceX64()`, or `PriceMath.invertPrice()`).
+   */
+    createSplashPool: (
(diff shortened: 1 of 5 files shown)


legacy-sdk/whirlpool/src/whirlpool-client.ts
```diff
@@ -92,6 +92,24 @@ export interface WhirlpoolClient {
...
+   * Create a Whirlpool account for a group of token A, token B and tick spacing
+   * @param whirlpoolConfig the address of the whirlpool config
+   * @param tokenMintA the address of the token A
+   * @param tokenMintB the address of the token B
+   * @param initialTick the initial tick that the pool is set to (derived from initial price)
+   * @param funder the account to debit SOL from to fund the creation of the account(s)
+   * @return `poolKey`: The public key of the newly created whirlpool account. `tx`: The transaction containing instructions for the on-chain operations.
+   * @throws error when the tokens are not in the canonical byte-based ordering. To resolve this, invert the token order and the initialTick (see `TickUtil.invertTick()`, `PriceMath.invertSqrtPriceX64()`, or `PriceMath.invertPrice()`).
+   */
+    createSplashPool: (
```
odcheung, line 105: to match CPMM, we need to hide the "tick" & "tick-array" concepts from the dev.

- this method sets the initial price to some default (ex $1)
- creates both tick-arrays in a splash-pool in the same call (createPool only creates 1 of them)

wjthieme: Got it. I assume it would be better to still be able to set initial price though

odcheung: can we have it as an optional?

odcheung (approved):
lgtm! I think we should make iniPrice as an optional parameter to mimic the full CPMM experience. approved otherwise

legacy-sdk/whirlpool/src/whirlpool-client.ts
wjthieme: 1 A/B as the default?

PR #202, opened 2024-08-14
bump legacy-sdk version to 0.13.4

yugure-orca:
bump & release legacy-sdk version to make SparseSwap feature public.


wjthieme:
Updated in #198 already. I'll deploying new version

PR #208, opened 2024-08-15
Update anchor to 0.30 and solana-program to 1.18

wjthieme:
Just keeping this here in case we might need it in the future. Will leave it as a draft since we'll probably skip this version and go straight to 0.31.

* Anchor changed something which makes the address constraint only work with either a constant or a property of a different account (but not with other expressions). I’ve changed them to use plain constraint instead.
* Had to Box a couple more accounts to fix Access violation in stack frame. We are getting close to what we can do with Box'ing accounts 😢
* Since we use the rent sysvar only for cpi and the account data from sysvar rent is not used in our program, I changed the type to UncheckedAccount which is only takes 8 stack space as opposed to Sysvar taking 32.
Anchor builds idl in snake_case. Need to convert it back to camelCase to keep sdk running with minimal changes.
Integration test results:

```
$anchor test
  2122 passing (38m)
  4 pending
```


MrG9090 (approved):

MrG9090 (approved):

baiwfg2:
@wjthieme  May I know why this is closed ? Any plan to upgrade to anchor 0.31.1 ?

PR #210, opened 2024-08-15
chore: update RouterUtils.html

eltociear:
existance -> existence


wjthieme:
Thanks for the spelling fix. This is a generated file. Would you be able to fix it in the source (./legacy-sdk/whirlpool/src/router/public/router-utils.ts)?

wjthieme:
Thanks. Closing this now

PR #244, opened 2024-08-19
Update swap math for splash pool edge-case

yugure-orca:
## Issue
https://linear.app/orca-so/issue/INF-94/[amm]-handle-u64-overflow-on-quoteexecution

Since SplashPool has only 2 initialized ticks at both ends, the possibility of exceeding u64 when calculating "delta amount" is higher than concentrated pools.

## Important points for updating logic
This problem occurs with ExactIn. The reason is that in ExactOut, "fixed delta" never exceeds the amount of tokens present in the pool and is clearly within the u64 range.

On the other hand, for ExactIn, "fixed delta" may exceed u64 because it calculates the amount of tokens needed to move the price to the end.

However, the primary purpose of initial calculation of "fixed delta" is to determine whether or not the iteration is "max swap" or not. So the info that “the amount of tokens required exceeds the u64 range” is sufficient to determine that the iteration is NOT "max swap".

### introduce `try_` functions
Add `try_get_amount_delta_a` and `try_get_amount_delta_b` functions to wrap and output errors exceeding u64 to avoid copying logic.

`get_amount_delta_a` and `get_amount_delta_b` work exactly as before by unwrapping the wrapped error.

```
[before]

get_amount_fixed_delta (& other caller) --> get_amount_delta_a
get_amount_fixed_delta (& other caller) --> get_amount_delta_b

[after]

get_amount_fixed_delta (& other caller) --> get_amount_delta_a --> try_get_amount_delta_a
get_amount_fixed_delta (& other caller) --> get_amount_delta_b --> try_get_amount_delta_b

try_get_amount_fixed_delta --> try_get_amount_delta_a
try_get_amount_fixed_delta --> try_get_amount_delta_b
```

### do not cover MultiplicationOverflow on `try_get_amount_delta_a`
`try_get_amount_delta_a` calculates `numerator = liquidity * sqrt_price_diff * 2^64` with U256.

However, it is possible that an overflow could occur because the maximum value of each variable is as follows:

- liquidity: ~ 2^110
- sqrt_price_diff: ~ 2^96

I did not change the conventional behavior since it involves a significant change to modify U256 handling.

The good news is that this overflow will never occur in SplashPool because SplashPool liquidity is constrained around 2^64.

The case where liquidity is 2^110 is a case where a large number of tokens are deposited in a very small range ([tick, tick+1] at the both end).
Even the liquidity at the center of USDC/USDT(1) is about 2^53.

### do not cover overflow error on `get_amount_unfixed_delta`
As mentioned above, in ExactIn, `get_amount_fixed_delta` may exceed u64 range, but `get_amount_unfixed_delta` does not.
On the other hand, in ExactOut, `get_amount_fixed_delta` does not exceed u64 range, but `get_amount_unfixed_delta` may exceed u64 range. 

In this case, however, the trade is logically impossible, meaning that an input of more than u64 tokens is required to obtain the specified amount. Therefore, there is no need to remedy this error.

### SDK side update
Updated SDK code to synchronize with contract logic updates.
In the process of verification, I fixed some issues that were originally present on the SDK side.
The fixed parts were commented on GitHub.

### generate_swap_testcase_json.ts
This script was used to generate `swap_test_cases.json` to test extreme inputs.
I updaated this script to generate `swap_test_cases_splash_pool.json` too.

Also, I found some bugs on the script and fixed them. (`Note(yugure)` is mark of fix)

Original script (contributor only): https://github.com/orca-so/whirlpool/blob/Swap-Integ-Test-Script/scripts/genSwapTestCases.ts
How to make these test cases (contributor only): https://www.notion.so/orcaso/292293f9b01a4b4091afc9133f07712e

### swap_test_cases.json
Some of the cases that were previously errors are no longer errors.
Conversely, there is no test cases which can no longer be processed that could be processed in the past.

Even in cases that can be processed, small inputs near extreme prices are often collected as fees without moving the price. However, this case is not executed due to unfavorable rates at the quote stage. Even if it is executed, setting other_amount_threshold appropriately will prevent a zero output.


legacy-sdk/whirlpool/src/utils/math/bit-math.ts
```diff
@@ -64,7 +66,7 @@ export class BitMath {
       );
     }
     const result = MathUtil.fromX64_BN(p);
-    const shouldRound = roundUp && result.and(U64_MAX).gt(ZERO);
+    const shouldRound = roundUp && p.and(U64_MAX).gt(ZERO);
```
yugure-orca, line 69: fix existing bug (result is already shifted, so result.and(U64_MAX) is almost always greater than zero.
(math in the contract-side is safe 🙂)

legacy-sdk/whirlpool/src/quotes/swap/swap-manager.ts
```diff
@@ -64,6 +65,19 @@ export function computeSwap(
...
+        SwapErrorCode.AmountRemainingOverflow,
+      );  
+    }
+    if (amountCalculated.gt(U64_MAX)) {
+      throw new WhirlpoolsError(
+        "Amount calculated is greater than U64_MAX.",
+        SwapErrorCode.AmountCalcOverflow,
+      );
+    }
+
```
yugure-orca, line 80: This part is corresponding to `checked_sub` and `checked_add` on the contract.

programs/whirlpool/src/math/token_math.rs
```diff
@@ -84,11 +133,42 @@ pub fn get_amount_delta_b(
...
-    checked_mul_shift_right_round_up_if(liquidity, price_upper - price_lower, round_up)
+pub fn try_get_amount_delta_b(
+    sqrt_price_0: u128,
+    sqrt_price_1: u128,
+    liquidity: u128,
+    round_up: bool,
+) -> Result<AmountDeltaU64, ErrorCode> {
+    let (sqrt_price_lower, sqrt_price_upper) = increasing_price_order(sqrt_price_0, sqrt_price_1);
+
+    // customized checked_mul_shift_right_round_up_if
```
philcchen, line 151: sorry if I'm missing something, why can't this function call into `checked_mul_shift_right_round_up_if` rather than duplicating the logic here?

I think it should be possible to just match on the error codes

programs/whirlpool/src/math/swap_math.rs
```diff
@@ -63,15 +65,17 @@ pub fn compute_swap(
     )?;
 
     // If the swap is not at the max, we need to readjust the amount of the fixed token we are using
-    if !is_max_swap {
-        amount_fixed_delta = get_amount_fixed_delta(
+    let amount_fixed_delta = if !is_max_swap || initial_amount_fixed_delta.exceeds_max() {
```
philcchen, line 68: nit: can we add a comment to explain that we are capping the amount delta here?

philcchen (approved):

programs/whirlpool/src/math/token_math.rs
yugure-orca: I changed this from the original draft using `match` because I felt that it may be an anti-pattern for the caller to know about all errors generated by callee functions.

If the error code is changed or the same error code is returned in multiple places in the function, it is difficult for the caller to notice and to identify the difference.

programs/whirlpool/src/tests/swap_integration_tests.rs
```diff
@@ -78,9 +81,31 @@ const CATCHABLE_ERRORS: [(&str, ErrorCode); 8] = [
...
+/// A total of 2880 tests on these variables:
+/// 1. FeeRate ([MAX_FEE, MAX_PROTOCOL_FEE], [65535, 600], [700, 300], [0, 0])
+/// 2. CurrentTickPosition (-443500, -223027, 0, 223027, 443500)
+/// 3. Liquidity (0, 2^32, 2^64)
+/// 4. TickSpacing (32768+1, 32768+64, 32768+128)
+/// 5. TradeAmount (0, 10^9, 10^12, U64::max)
+/// 6. Trade Direction (a->b, b->a)
+/// 7. TradeAmountToken (amountIsInput, amountIsOutput)
+fn run_splash_pool_swap_integration_tests() {
+    run_swap_integration_tests(
```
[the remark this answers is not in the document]
yugure-orca: yep, `read_to_string` is now integrated into `run_swap_integration_tests` function.

programs/whirlpool/src/math/swap_math.rs
yugure-orca: yep.
I added detailed comment on the head of the function, and added comment here to clarify that the result (`amount_fixed_delta`) will be in the u64 range.

legacy-sdk/whirlpool/src/errors/errors.ts
```diff
@@ -20,6 +20,8 @@ export enum SwapErrorCode {
   TickArrayCrossingAboveMax = `TickArrayCrossingAboveMax`,
   TickArrayIndexNotInitialized = `TickArrayIndexNotInitialized`,
   TickArraySequenceInvalid = `TickArraySequenceInvalid`,
+  AmountRemainingOverflow = `AmountRemainingOverflow`,
+  AmountCalcOverflow = `AmountCalcOverflow`,
```
odcheung, line 24: can we add the failure expectations to the `swap` method comments in `swap-ix.ts` & `ix.ts`

odcheung (commented):
finished reviewing the rust code. going through the TS then the generation script

programs/whirlpool/src/math/swap_math.rs
```diff
@@ -40,7 +50,7 @@ pub fn compute_swap(
         .try_into()?;
     }
 
-    let next_sqrt_price = if amount_calc >= amount_fixed_delta {
+    let next_sqrt_price = if initial_amount_fixed_delta.lte(amount_calc) {
```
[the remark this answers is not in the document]
odcheung: ah nevermind. just by reading the code, this seems like the equal case switched conditions.

on a deeper dive - amount_calc === initial_amount_fixed_delta is required to set `next_sqrt_price` to `sqrt_price_target`, so `is_max_swap` is set. So the equal case must === sqrt_price_target.

legacy-sdk/whirlpool/src/utils/math/bit-math.ts
odcheung: ah good catch

legacy-sdk/whirlpool/tests/integration/multi-ix/splash_pool.test.ts
```diff
@@ -54,9 +61,343 @@ describe("splash pool tests", () => {
...
+      //
+
+      ////////////////////////////////////////////////////////////////////////////////
+      // ExactIn
+      ////////////////////////////////////////////////////////////////////////////////
+
+      // ExactIn, BtoA, min to ...
+      {
+        figure: "(toB) |-----mS----l**********T*********|********************u-----x-----| (toA)",
+        poolTickSpacing: 32768 + 128,
```
odcheung, line 100: why +128?

odcheung (approved):
looked over the following:
-  the splash pool test cases. looks legit w/ 1 clarification!
- typescript sdk changes
- gen test case script (good catches!)

legacy-sdk/whirlpool/tests/integration/multi-ix/splash_pool.test.ts
yugure-orca: 32768 is `2^15`.
It can be used as minimum tickspacing for SplashPool, but I used `32896 (32768 + 128)` as tickspacing for testing because it is used in production (mainnet) and we will not use 32768.

The following "rule" is convenient.

1% --> 128 (concentrated) & 32768 + 128 (splash)
n% --> x (concentrated) & 32768 + x (splash)

legacy-sdk/whirlpool/src/errors/errors.ts
yugure-orca: good point!
added👍️ (these errors and recently added errors)

PR #250, opened 2024-08-21
Ready to Rust 1.18 (use repr(C, packed))

yugure-orca:
## Change from repr(Rust, packed) to repr(C, packed)
Now `repr(packed)` (=`repr(Rust, packed)`) is used for `Tick` and `TickArray`.

In Rust 1.80, the order of fields will be adjusted and it occurs data corruption.
To protect struct from field order adjustment, we need to use `repr(C, packed)` explicitly.

## Add test cases
Added unit tests to detect unintended changes to the data layout for all state.

internal: https://linear.app/orca-so/issue/INF-93/[amm]-use-reprcpacked-on-tickarray-and-tick

## Test
1.78.0-x86_64-apple-darwin: passed (cargo test) note: without `C`, it worked
1.80.1-x86_64-apple-darwin: passed (cargo test) note: without `C`, it failed


wjthieme:
Can you also update the github actions to use rustc 1.80?

yugure-orca:
> Can you also update the github actions to use rustc 1.80?

yep, the latest checks have been done with Rust 1.80. 🙂

wjthieme (approved):
🚢

programs/whirlpool/src/state/tick.rs
```diff
@@ -752,3 +752,86 @@ mod array_update_tests {
...
+        tick_data[offset..offset + 16].copy_from_slice(&tick_fee_growth_outside_b.to_le_bytes());
+        offset += 16;
+        for i in 0..NUM_REWARDS {
+            tick_data[offset..offset + 16]
+                .copy_from_slice(&tick_reward_growths_outside[i].to_le_bytes());
+            offset += 16;
+        }
+
+        // manually build the expected TickArray data layout
+        // note: no discriminator
```
philcchen, line 796: any reason to not test with the discriminator/with `TickArray::try_deserialize` similar to the other tests?

philcchen (approved):

programs/whirlpool/src/state/tick.rs
yugure-orca: Good point!
Yep, there is a clear reason.

TickArray account doesn't implement `try_deserialize` because it is implemented by `AccountDeserialize` trait.
`AccountDeserialize` trait will be added by using `account` attribute macro (e.g. Whirlpool account).
https://docs.rs/anchor-lang/latest/anchor_lang/attr.account.html

But TickArray account uses `zero_copy` and it doesn't add `AccountDeserialize` trait.
So we need to test data layout in AccountLoader way. (bytemuck).
https://docs.rs/anchor-lang/latest/anchor_lang/attr.zero_copy.html

`bytemuck::from_bytes` (in `load()`) will be applied to [8..], so to test the validity to data layout mapping, first 8 bytes is not needed.
https://docs.rs/anchor-lang/latest/src/anchor_lang/accounts/account_loader.rs.html#165
```
        Ok(Ref::map(data, |data| {
            bytemuck::from_bytes(&data[8..mem::size_of::<T>() + 8])
        }))
```

philcchen: Ah, sorry if I'm misunderstanding, but I think `[account(zero_copy)]` will still generate `AccountDeserialize` (but not `AccountSerialize`, `AnchorSerialize` or `AnchorDeserialize`)

The `AccountDeserialize` it generates does the above + the discriminator checking ?
https://github.com/coral-xyz/anchor/blob/v0.29.0/lang/attribute/account/src/lib.rs#L180

Totally okay to keep as is though, since it's eventually how we load the TickArrays anyways in `peek_tick_array` https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/util/sparse_swap.rs#L290

yugure-orca: oh sorry, I overlooked that it still implement AnchorDeserialize 🤯
Yeah, I agree that we should check bytemuck instead of borsh. 👍

PR #270, opened 2024-08-26
Dependabot group program dependency updates

wjthieme:
This should create a single PR for dependencies of the program. This will still create separate dependencies for security-updates. If a crate is marked as vulnerable and an update is available this will still be in a separate PR.


yugure-orca (approved):
Thanks!🙏

PR #277, opened 2024-08-28
Reject partial fill in ExactOut if sqrt_price_limit = 0

yugure-orca:
## The risk of partial fill in ExactOut
Trades may complete with partial fill when sqrt_price_limit is reached in both ExactIn and ExactOut.

In ExactIn, the amount of output is verified by other_amount_threshold, so even if a partial fill occurs, the transaction will fail unless the trader obtains the desired output for the input.

In ExactOut, on the other hand, other_amount_threshold defines an upper limit on the amount of input, so that a transaction will not fail when a part of input amount is used. However, this may result in an unfavorable rate if the specified amount of output is not obtained due to partial fill.

For traders who wish to explicitly reject the occurrence of Partial fill in ExactOut, the program will accept an input of sqrt_price_limit = 0. If sqrt_price_limit = 0, then if a Partial fill occurs in ExactOut, the transaction will be rejected.

Note that sqrt_price_limit = 0 does not overlap with the previously accepted value, so a breaking change will not occur.

## TwoHopSwap (v1) partial fill rejection
TwoHopSwap receives sqrt_price_limit_one and sqrt_price_limit_two, which may result in partial fill.

Among these cases, the following cases will result in an excess or shortage of intermediate tokens.
- Partial fill of the second swap in ExactIn 
- Partial fill of the first swap in ExactOut

In both cases, the pool is protected even in the case of over/shortage, since the transfer between the pool and the owner is always used. However, unintended intermediate tokens may be used or left over.
To avoid this, if the output and input amounts of intermediate tokens do not match, the transaction will fail.

V2 already performs this verification for vault to vault transfers.

### DIFF
--- a/legacy-sdk/whirlpool/tests/integration/two_hop_swap.test.ts
+++ b/legacy-sdk/whirlpool/tests/integration/two_hop_swap.test.ts
@@ -1037,6 +1037,7 @@ describe("two-hop swap", () => {
           ctx,
           WhirlpoolIx.twoHopSwapIx(ctx.program, {
             ...twoHopQuote,
+            sqrtPriceLimitOne: MIN_SQRT_PRICE_BN, // Partial fill is allowed
             sqrtPriceLimitTwo: new BN(0), // Partial fill is NOT allowed
             ...getParamsFromPools([pools[0], pools[1]], tokenAccounts),
             tokenAuthority: ctx.wallet.publicKey,
@@ -1049,6 +1050,7 @@ describe("two-hop swap", () => {
         ctx,
         WhirlpoolIx.twoHopSwapIx(ctx.program, {
           ...twoHopQuote,
+          sqrtPriceLimitOne: MIN_SQRT_PRICE_BN, // Partial fill is allowed
           sqrtPriceLimitTwo: MIN_SQRT_PRICE_BN, // Partial fill is allowed
           ...getParamsFromPools([pools[0], pools[1]], tokenAccounts),
           tokenAuthority: ctx.wallet.publicKey,
@@ -1130,6 +1132,7 @@ describe("two-hop swap", () => {
           WhirlpoolIx.twoHopSwapIx(ctx.program, {
             ...twoHopQuote,
             sqrtPriceLimitOne: new BN(0), // Partial fill is NOT allowed
+            sqrtPriceLimitTwo: MIN_SQRT_PRICE_BN, // Partial fill is allowed
             ...getParamsFromPools([pools[0], pools[1]], tokenAccounts),
             tokenAuthority: ctx.wallet.publicKey,
           }),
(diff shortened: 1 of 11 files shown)


legacy-sdk/whirlpool/tests/integration/v2/collect_fees_v2.test.ts
yugure-orca, line 1: These changes are fix for revert miss from Anchor 0.30 PR. (not related to this PR change, but to pass all test🙂)

programs/whirlpool/src/instructions/two_hop_swap.rs
```diff
@@ -202,6 +202,21 @@ pub fn handler(
         (swap_calc_one, swap_calc_two)
     };
 
+    // All output token should be consumed by the second swap
+    let swap_calc_one_output = if a_to_b_one {
```
philcchen, line 206: This is _technically_ a breaking change, though it is still a reasonable guardrail

legacy-sdk/whirlpool/tests/integration/two_hop_swap.test.ts
```diff
@@ -829,6 +827,359 @@ describe("two-hop swap", () => {
     );
   });
 
+  describe("partial fill", () => {
```
philcchen, line 830: I think it's still possible to partial fill on two hop swaps in the following cases:

ExactOut, where sqrt_price_limit_two != 0, partial filled on second leg
ExactIn.  , where sqrt_price_limit_one != 0, partial filled on first leg

As long as the other amount thresholds values are sufficiently low/high

Should we add additional test cases here? (success + failure depending on the other_amount_threshold?)

programs/whirlpool/src/instructions/two_hop_swap.rs
yugure-orca: Yep, strictly speaking, it is breaking.
Good news (?) is almost all use of two_hop_swap is from Orca site. (Because its execution number is quite lower than swap).
And Orca backend doesn't use Partial fill now.

philcchen (approved):

legacy-sdk/whirlpool/tests/integration/two_hop_swap.test.ts
yugure-orca: Good point!
Added 2 test cases for two_hop_swap v1 & v2.

legacy-sdk/whirlpool/tests/integration/v2/two_hop_swap_v2.test.ts
```diff
@@ -2270,6 +2275,378 @@ describe("two_hop_swap_v2", () => {
...
+      );
+    
+      const twoHopQuote = twoHopSwapQuoteFromSwapQuotes(quoteFirst, quoteSecond);
+  
+      await assert.rejects(
+        toTx(
+          ctx,
+          WhirlpoolIx.twoHopSwapV2Ix(ctx.program, {
+            ...twoHopQuote,
+            sqrtPriceLimitTwo: new BN(0), // Partial fill is NOT allowed
```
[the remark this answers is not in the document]
yugure-orca: It makes sense!
Added `sqrtPriceLimitOne` for this and other test cases.
Now both `sqrtPriceLimintOne` and `sqrtPriceLimitTwo` are clearly described.

philcchen (approved):

wjthieme (approved):
🚀

PR #282, opened 2024-08-30
Add github actions publish script for npm and crates

wjthieme:
This would only publish `client` packages on tag. The other packages are still WIP and will not be published yet

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -43,7 +43,7 @@ jobs:
       uses: ./.github/actions/anchor
       with:
         run: |
-          sed -i 's/version = v[0-9]+.[0-9]+.[0-9]+/version = $(basename ${{ github.ref }})/' rust-sdk/${{ matrix.package }}/Cargo.toml
+          sed -i 's/^version = ".*"$/version = "$(basename ${{ github.ref }})"/' rust-sdk/${{ matrix.package }}/Cargo.toml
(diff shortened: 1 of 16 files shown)


odcheung (approved):

.github/workflows/publish.yml
```diff
@@ -43,23 +43,30 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
     - name: Checkout Repository
       uses: actions/checkout@v4
       with:
         fetch-depth: 0
-    - name: Deploy cargo
-      uses: ./.github/actions/anchor
+    - name: Verifiable build
+      run: anchor build --verifiable
```
yugure-orca, line 61: Is it needed to use `--verifiable`?
In my understanding, Anchor verifiable build is not updated and we are using new one from Ellipsis-Labs.
https://github.com/Ellipsis-Labs/solana-verifiable-build
(Of course I believe that tech behind them is same (use some prebuild docker env to build), but I think the image is different)

If the purpose of this build is getting IDL, IDL is not affected from build environment.

yugure-orca (approved):
LGTM with small q. (not critical)

.github/workflows/publish.yml
wjthieme: This step doesn't do anything that we need to rely on. All it does now is add the .so binary and idl to the github release

PR #291, opened 2024-09-03
Fix TransferHook additional accounts derive logic

yugure-orca:
https://linear.app/orca-so/issue/INF-92/[amm]-immunefi-34511-transferhook-parameter-bug

In `transfer_from_vault_to_owner_v2`:
```
        spl_transfer_hook_interface::onchain::add_extra_accounts_for_execute_cpi(
            &mut instruction,
            &mut account_infos,
            &hook_program_id,
            token_owner_account.to_account_info(),
            token_mint.to_account_info(),
            token_vault.to_account_info(),
            whirlpool.to_account_info(),
```
should be
```
        spl_transfer_hook_interface::onchain::add_extra_accounts_for_execute_cpi(
            &mut instruction,
            &mut account_infos,
            &hook_program_id,
            token_vault.to_account_info(),
            token_mint.to_account_info(),
            token_owner_account.to_account_info(),
            whirlpool.to_account_info(),
```
because `token_vault` is SOURCE and `token_owner_account` is DESTINATION.

This logic is used to derive the additional accounts required for TransferHook, which is an error when using a PDA that varies depending on source and destination.

Since it is also validated by transfer_checked in Token-2022, the TransferHook program will not execute using the incorrectly derived PDA, but the transaction will fail.

## Test update
Improved the transfer_hook_counter program used for testing to detect this problem.
Add a new PDA that seeds source, mint, destination, and owner addresses in this order. This allows us to detect problems in integration tests.

https://github.com/yugure-orca/transfer-hook-counter/blob/main/programs/transfer-hook-counter/src/lib.rs#L46-L57


programs/whirlpool/src/util/v2/token.rs
```diff
@@ -154,10 +159,11 @@ pub fn transfer_from_vault_to_owner_v2<'info>(
...
             &mut account_infos,
             &hook_program_id,
-            token_owner_account.to_account_info(),
-            token_mint.to_account_info(),
-            token_vault.to_account_info(),
-            whirlpool.to_account_info(),
+            // vault to owner
+            token_vault.to_account_info(), // from (vault account)
+            token_mint.to_account_info(), // mint
+            token_owner_account.to_account_info(), // to (owner account)
```
yugure-orca, line 165: This part is the issue.

philcchen (approved):

wjthieme (approved):
🚢

PR #294, opened 2024-09-04
InterestBearing support

yugure-orca:
https://linear.app/orca-so/issue/INF-149/[amm]-interestbearing-support

## Add InterestBearing extension as supported extensions
- The program handles u64 amount (raw amount) only. (not handle UI Amount)
- TokenBadge is not required.

## Current state: ~~Pending~~ Go
~~The handling of InterestBearing is being discussed so as not to create confusion throughout the Solana ecosystem.
Supporting it should wait for now.~~

There is no problem to enable it on the contract side, so we'll merge the PR.
We need to tackling to the difficulty in handling InterestBearing on the UI side. 😂


wjthieme (approved):
🚢

philcchen (approved):

Theghost256 (approved):

PR #429, opened 2024-10-29
Include low-level sdk in rustdoc/typedoc

wjthieme:
Before the Rust/TS API reference page of the docs only included the high-level SDK. This PR changes that to also include the low level SDKs. @calintje let me know what you think!

**Before**
Screenshot 2024-10-29 at 5 14 21 PM
Screenshot 2024-10-29 at 5 14 34 PM

**After**
Screenshot 2024-10-29 at 5 12 36 PM
Screenshot 2024-10-29 at 5 12 16 PM


calintje (commented):
Nice! I think it’s helpful to have Typedocs/Rustdocs available for all released packages.

The core documentation already looks good, but it would be even better to align its style with the high-level SDK for consistency. I can create a ticket to update the inline documentation to support this.

For the autogenerated code, Codama exports a lot of boilerplate and internal-use type definitions and functions, which might not be very helpful and could create confusion for developers. It may be worth exploring if Codama can be configured to reduce this and provide more developer-focused output.

Additionally, would it be possible to update the navbar? Grouping "Rust SDK Reference" and "TS SDK Reference" under a dropdown titled "Whirlpools SDK Reference" would align it with the sidebar terminology and improve navigation.
Screenshot 2024-10-30 005907

wjthieme:
> The core documentation already looks good, but it would be even better to align its style with the high-level SDK for consistency. I can create a ticket to update the inline documentation to support this.

Guess you are referring to ts-core? This one is difficult because it is built from the rust code (the rust comments get copied over). Either we would need some way to transform this (which I don't know if exists) or have double documentation 

> For the autogenerated code, Codama exports a lot of boilerplate and internal-use type definitions and functions, which might not be very helpful and could create confusion for developers

Since this is low-level I wouldn't worry too much about this

> Additionally, would it be possible to update the navbar? Grouping "Rust SDK Reference" and "TS SDK Reference" under a dropdown titled "Whirlpools SDK Reference" would align it with the sidebar terminology and improve navigation.

Yep feel free to create a ticket/pr for this!

calintje:
> Guess you are referring to ts-core? This one is difficult because it is built from the rust code (the rust comments get copied over). Either we would need some way to transform this (which I don't know if exists) or have double documentation

Yes, I do. And that makes sense. The only thing we could add then are example usages, like so:
**Before**
Screenshot 2024-10-30 162510
**After**
Screenshot 2024-10-30 163408

Although I don't think we should prioritize that now, unless devs will start requesting this.

> Since this is low-level I wouldn't worry too much about this
Cool.

> Yep feel free to create a ticket/pr for this!
Will do!

odcheung (approved):
code lgtm but i'll defer to the sign off of the dev doc structure to @calintje

calintje (approved):
Very nice addition!

PR #435, opened 2024-10-30
Add network selector

calintje:
### Title
Add network selector

### Description
Add a network selector function to abstract away the notion of WhirlpoolsConfig accounts. This function allows users to select a network (SolanaMainnet, SolanaDevnet, EclipseMainnet, EclipseTestnet) and automatically sets the corresponding WhirlpoolsConfig address.

- Added `setNetwork` function to set the WhirlpoolsConfig address based on the selected network.
- Included TypeDoc comments for the new function.
- Added tests to verify the functionality of the `setNetwork` function.


ts-sdk/whirlpool/tests/config.test.ts
```diff
@@ -43,6 +45,18 @@ describe("Configuration", () => {
     );
   });
 
+  it("Should be able to set the network", async () => {
```
wjthieme, line 48: Nit: should be able to set whirlpools config based on network

You're not actually setting the network

ts-sdk/whirlpool/src/config.ts
```diff
@@ -49,6 +58,18 @@ export async function setWhirlpoolsConfig(
     );
 }
 
+/**
+ * Sets the WhirlpoolsConfig address based on the selected network.
+ *
+ * @param {keyof typeof WHIRLPOOLS_CONFIG_ADDRESSES} network - The network to set. Valid options are: SolanaMainnet, SolanaDevnet, EclipseMainnet, EclipseTestnet.
+ * @throws {Error} If the network is invalid.
+ */
+export async function setNetwork(
```
wjthieme, line 67: Nit: this is not actually setting the network so not sure if the naming is fully appropriate. You could overload the already existing 'setWhirlpoolsConfig' function.

The solana cli for example takes a rpcUrlOrCanonialName parameter when setting the rpc url

wjthieme (commented):
Awesome! Couple nits on the naming but a nice addition

ts-sdk/whirlpool/src/config.ts
calintje: You are right. I went for this name, because I wanted to approach it from the perspective of the developer with the goal to abstract away the notion of WhirlpoolsConfig account.

How about `configureNetwork`, `setNetworkConfig`, or `useNetwork`?

If you prefer the overload option, I would just have to explain it well in the Docs.

wjthieme: > How about configureNetwork, setNetworkConfig, or useNetwork?

Think all of these still have the same problem. This function is not doing anything with the network. It is setting the whirlpoolConfigAddress based on the network. I guess maybe something like `setWhirlpoolsConfigByNetwork` could work but then we might as well overload it. Another option could be `configure` or `configureSdk` but not sure if that is descriptive enough either

ts-sdk/whirlpool/src/config.ts
```diff
@@ -10,18 +10,18 @@ export const DEFAULT_ADDRESS = address("11111111111111111111111111111111");
 /**
  * The WhirlpoolsConfig addresses for various networks.
  */
-export const WHIRLPOOLS_CONFIG_ADDRESSES = {
-  SolanaMainnet: address("2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ"),
-  SolanaDevnet: address("FcrweFY1G9HJAHG5inkGB6pKg1HZ6x9UC2WioAfWrGkR"),
-  EclipseMainnet: address("FVG4oDbGv16hqTUbovjyGmtYikn6UBEnazz6RVDMEFwv"),
-  EclipseTestnet: address("FPydDjRdZu9sT7HVd6ANhfjh85KLq21Pefr5YWWMRPFp"),
+export const DEFAULT_WHIRLPOOLS_CONFIG_ADDRESSES = {
+  solanaMainnet: address("2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ"),
```
wjthieme, line 14: should we just call it `solana` and `eclipse` without mainnet?

ts-sdk/whirlpool/src/config.ts
```diff
@@ -45,31 +45,26 @@ export let WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Address =
...
+ * @param {Address | keyof typeof NETWORK_ADDRESSES} config - A WhirlpoolsConfig address or a network name.
  * @returns {Promise<void>} - Resolves when the addresses have been updated.
  */
 export async function setWhirlpoolsConfig(
-  whirlpoolsConfigAddress: Address,
+  config: Address | keyof typeof DEFAULT_WHIRLPOOLS_CONFIG_ADDRESSES,
 ): Promise<void> {
+  let whirlpoolsConfigAddress: Address;
+  if (isAddress(config)) {
+    whirlpoolsConfigAddress = config;
```
wjthieme, line 56: Think you can just set `WHIRLPOOLS_CONFIG_ADDRESS` directly?

wjthieme (approved):
Two small nits but LGTM 🚢

ts-sdk/whirlpool/src/config.ts
calintje: Personally, I prefer with mainnet, because it's more descriptive? You have the final say.

PR #436, opened 2024-10-30
Add Whirlpools SDK dropdown to navbar

calintje:
### Title
Add Whirlpools SDK dropdown to navbar

### Description
- Added a "Whirlpools SDK Reference" dropdown to the navbar.
- The dropdown includes links to "TS SDK Reference" and "Rust SDK Reference".
- This change groups the SDK references under a single menu item for better organization and easier navigation.
Screenshot 2024-10-30 164545

### DIFF
--- a/docs/whirlpool/docusaurus.config.js
+++ b/docs/whirlpool/docusaurus.config.js
@@ -49,16 +49,20 @@ export default {
       items: [
         { to: "/", label: "Docs", position: "left" },
         {
-          href: "/orca_whirlpools_docs/",
-          label: "Rust SDK Reference",
+          label: "Whirlpools SDK Reference",
           position: "left",
-          target: "_blank",
-        },
-        {
-          href: "/ts/",
-          label: "TS SDK Reference",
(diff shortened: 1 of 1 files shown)


wjthieme:
Should we also add Legacy SDK in the dropdown?

calintje:
> Should we also add Legacy SDK in the dropdown?

If you agree with the sidebar organization of PR concerning migration of developer protal docs, I would also keep the distinction in the navbar between 'Whirlpools SDK' (new SDK) and 'Legacy SDK'
Screenshot 2024-10-30 183507

wjthieme:
> If you agree with the sidebar organization of PR concerning migration of developer protal docs, I would also keep the distinction in the navbar between 'Whirlpools SDK' (new SDK) and 'Legacy SDK'

Can we hide Legacy SDK reference away in the sidebar somewhere in that case? I don't think it needs to be that prominent on the nav bar (cause we want users to use the new sdk and not the old one)

wjthieme (approved):

PR #448, opened 2024-11-01
Improve the ts-sdk testing setup

wjthieme:
* Add tests for fetching positions
* Use actual instructions for initializing pools/positions/tokens/etc. instead of trying to write the data ourselves
* Bankrun does not support gpa yet so some functions cannot be tested through bankrun


ts-sdk/whirlpool/tests/e2e.test.ts
```diff
@@ -7,87 +7,37 @@ import {
...
-        isNative: null,
-        delegatedAmount: 0,
-        closeAuthority: null,
-      }),
-      TOKEN_PROGRAM_ADDRESS,
-    );
+    const mint1 = await setupMint({ decimals: 9 });
+    const mint2 = await setupMint({ decimals: 6 });
+    [mintA, mintB] =
+      Buffer.from(mint1) < Buffer.from(mint2) ? [mint1, mint2] : [mint2, mint1];
```
yugure-orca, line 38: nit: orderMints

ts-sdk/whirlpool/tests/pool.test.ts
```diff
@@ -1,5 +1,89 @@
...
+  let mintB: Address;
+  let defaultPool: Address;
+  let concentratedPool: Address;
+  let splashPool: Address;
+
+  beforeAll(async () => {
+    const mint1 = await setupMint();
+    const mint2 = await setupMint();
+    [mintA, mintB] =
+      Buffer.from(mint1) < Buffer.from(mint2) ? [mint1, mint2] : [mint2, mint1];
```
yugure-orca, line 29: nit: orderMints

ts-sdk/whirlpool/tests/position.test.ts
```diff
@@ -1,5 +1,49 @@
...
+  let mintA: Address;
+  let mintB: Address;
+  let pool: Address;
+  let splashPool: Address;
+
+  beforeAll(async () => {
+    const mint1 = await setupMint();
+    const mint2 = await setupMint();
+    [mintA, mintB] =
+      Buffer.from(mint1) < Buffer.from(mint2) ? [mint1, mint2] : [mint2, mint1];
```
yugure-orca, line 25: nit: orderMints

ts-sdk/whirlpool/tests/token.test.ts
```diff
@@ -207,32 +146,23 @@ describe("Token Account Creation", () => {
     setSolWrappingStrategy("none");
 
     const result = await prepareTokenAccountsInstructions(rpc, signer, {
-      [TOKEN_MINT_1]: 100n,
-      [TOKEN_MINT_2]: 100n,
+      [mintA]: 100n,
+      [mintB]: 100n,
       [NATIVE_MINT]: 100n,
```
yugure-orca, line 151: Just to be sure, in this case (strategy is none), the balance of ATA for native mint should be zero even if 100n is specified.

wjthieme: Yep. This is how it is currently explained in the docs:
```
 * - **None**:
 *   Uses or creates the ATA without performing any SOL wrapping or unwrapping.
 ```

It might make sense to throw if the balance is insufficient (also for the non-SOL mints). Basically this can be used to say: "I need x amount in this token account" and is currently only doing something for SOL (if wrapping strategy is not none). What do you think?

yugure-orca: Thanks for the explanation!
It makes sense🙂 (memo: the instruction requires X SOL, but my ATA already has >  X SOL, so no need to wrap)

yugure-orca (approved):
🚀

PR #457, opened 2024-11-04
Fix a bug ordering tokens

wjthieme:
Issue: ts-code was comparing `utf8(address_in_base58)` while it should have been comparing `address_in_bytes`. This is because `Buffer.from(Address)` does not return the address bytes but instead takes the string (which Address basically is) and convert that to utf8 bytes.

### DIFF
--- a/ts-sdk/whirlpool/src/token.ts
+++ b/ts-sdk/whirlpool/src/token.ts
@@ -288,5 +288,9 @@ export function getCurrentTransferFee(
  */
 export function orderMints(mint1: Address, mint2: Address): [Address, Address] {
   const encoder = getAddressEncoder();
-  return encoder.encode(mint1) < encoder.encode(mint2) ? [mint1, mint2] : [mint2, mint1];
+  const mint1Bytes = new Uint8Array(encoder.encode(mint1));
+  const mint2Bytes = new Uint8Array(encoder.encode(mint2));
+  return Buffer.compare(mint1Bytes, mint2Bytes) < 0
+    ? [mint1, mint2]
+    : [mint2, mint1];
 }
--- a/ts-sdk/whirlpool/src/token.ts
+++ b/ts-sdk/whirlpool/src/token.ts
@@ -279,3 +279,14 @@ export function getCurrentTransferFee(
     maxFee: transferFee.maximumFee,
   };
 }
+
+/**
+ * Orders two mints by canonical byte order.
+ * @param {Address} mint1
+ * @param {Address} mint2
+ * @returns {[Address, Address]} [mint1, mint2] if mint1 should come first, [mint2, mint1] otherwise
+ */
+export function orderMints(mint1: Address, mint2: Address): [Address, Address] {
+  const encoder = getAddressEncoder();
+  return encoder.encode(mint1) < encoder.encode(mint2) ? [mint1, mint2] : [mint2, mint1];
+}
--- a/ts-sdk/whirlpool/tests/token.test.ts
+++ b/ts-sdk/whirlpool/tests/token.test.ts
@@ -602,8 +602,8 @@ describe("Token Account Creation", () => {
   });
 
   it("Should order mints by canonical byte order", () => {
-    const mint1 = address("BRjpCHtyQLNCo8gqRUr8jtdAj5AjPYQaoqbvcZiHok1k");
-    const mint2 = address("Jd4M8bfJG3sAkd82RsGWyEXoaBXQP7njFzBwEaCTuDa");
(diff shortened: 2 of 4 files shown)


ts-sdk/whirlpool/tests/token.test.ts
```diff
@@ -600,4 +600,16 @@ describe("Token Account Creation", () => {
...
+  it("Should order mints by canonical byte order", () => {
+    const mint1 = address("BRjpCHtyQLNCo8gqRUr8jtdAj5AjPYQaoqbvcZiHok1k");
+    const mint2 = address("Jd4M8bfJG3sAkd82RsGWyEXoaBXQP7njFzBwEaCTuDa");
+    const [mintA, mintB] = orderMints(mint1, mint2);
+    assert.strictEqual(mintA, mint1);
+    assert.strictEqual(mintB, mint2);
+
+    const [mintC, mintD] = orderMints(mint2, mint1);
+    assert.strictEqual(mintC, mint1);
+    assert.strictEqual(mintD, mint2);
```
yugure-orca, line 613: hmm, is this test data correct ?

```
import { PoolUtil } from "@orca-so/whirlpools-sdk";
import { PublicKey } from "@solana/web3.js";

async function main() {
    const mint1 = new PublicKey("BRjpCHtyQLNCo8gqRUr8jtdAj5AjPYQaoqbvcZiHok1k");
    const mint2 = new PublicKey("Jd4M8bfJG3sAkd82RsGWyEXoaBXQP7njFzBwEaCTuDa");
    const [mintA, mintB] = PoolUtil.orderMints(mint1, mint2);
    console.log("mintA == mint1", mintA.toString() === mint1.toString());
    console.log("mintB == mint2", mintB.toString() === mint2.toString());
    console.log("mintA", mintA.toString());
    console.log("mintB", mintB.toString());
}

main();
```
shows
```
mintA == mint1 false
mintB == mint2 false
mintA Jd4M8bfJG3sAkd82RsGWyEXoaBXQP7njFzBwEaCTuDa
mintB BRjpCHtyQLNCo8gqRUr8jtdAj5AjPYQaoqbvcZiHok1k
```

[Converter](https://appdevtools.com/base58-encoder-decoder) shows

- `BRjpCHtyQLNCo8gqRUr8jtdAj5AjPYQaoqbvcZiHok1k`: `9aeb37...`
- `Jd4M8bfJG3sAkd82RsGWyEXoaBXQP7njFzBwEaCTuDa`: `0483a4...`

ts-sdk/whirlpool/src/token.ts
```diff
@@ -279,3 +279,14 @@ export function getCurrentTransferFee(
...
+
+/**
+ * Orders two mints by canonical byte order.
+ * @param {Address} mint1
+ * @param {Address} mint2
+ * @returns {[Address, Address]} [mint1, mint2] if mint1 should come first, [mint2, mint1] otherwise
+ */
+export function orderMints(mint1: Address, mint2: Address): [Address, Address] {
+  const encoder = getAddressEncoder();
+  return encoder.encode(mint1) < encoder.encode(mint2) ? [mint1, mint2] : [mint2, mint1];
```
yugure-orca, line 291: `<` will try to convert operands to string ("[20, 3, 4]", "[1, 2, 3]").

```
   const array1 = new Uint8Array([1, 2, 3]);
   const array2 = new Uint8Array([2, 3, 4]);
   console.log("array1 < array2", array1 < array2);

   const array3 = new Uint8Array([100, 2, 3]);
   const array4 = new Uint8Array([2, 3, 4]);
   console.log("array3 < array4", array3 < array4);

   const array5 = new Uint8Array([3, 4, 5]);
   const array6 = new Uint8Array([20, 3, 4]);
   console.log("array5 < array6", array5 < array6);
```

yugure-orca (changes requested):
Comparisons by < do not appear to be working as expected.

ts-sdk/whirlpool/src/token.ts
wjthieme: Interesting. So buffer.compare works differently than '<'. I'll copy the legacy code cause we know that works

ts-sdk/whirlpool/tests/token.test.ts
wjthieme: It should be flipped. I misread whoever reported the issue

ts-sdk/whirlpool/src/token.ts
yugure-orca: memo: `Buffer.compare` can be used for both `Buffer` and `Uint8Array`
https://nodejs.org/api/buffer.html#static-method-buffercomparebuf1-buf2

yugure-orca (approved):
🚀

PR #458, opened 2024-11-04
Update and fix documentation

calintje:
**Title**
Update docs

**Details**
- Update wallet management, preventing users from not storing their keypairs and losing their assets.
- Fix imports and return types of code examples
- Fix typedocs


yugure-orca (approved):
Thank you!

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -60,17 +60,18 @@ export type CreatePoolInstructions = {
  *
  * @example
  * import { createSplashPoolInstructions } from '@orca-so/whirlpools';
- * import { generateKeyPairSigner, createSolanaRpc, devnet } from '@solana/web3.js';
+ * import { createKeyPairSignerFromBytes, createSolanaRpc, devnet, lamports } from '@solana/web3.js';
  *
  * const devnetRpc = createSolanaRpc(devnet('https://api.devnet.solana.com'));
- * const wallet = await generateKeyPairSigner();
+ * const keyPairBytes = new Uint8Array(JSON.parse(fs.readFileSync('path/to/solana-keypair.json', 'utf8')));
```
wjthieme, line 66: On the one hand I like this because it is better practice than generatekeypair. On the other hand now each example requires a setup step. 

I'm in favor of leaving this as generatekeypair but wouldn't be opposed to adding a comment: "see wallat management on how to load a wallet" or something

wjthieme: On the other hand don't really care as much as long as it is clear to the user

wjthieme (approved):

ts-sdk/whirlpool/src/createPool.ts
calintje: Links to documention inside an example code-block does not look nice.
image

I will change to:
```tsx
const wallet = await generateKeyPairSigner(); // CAUTION: this wallet is no persistent.
```
Since the functions only generate instructions, and there's no mentioning of "SOL airdrop on Devnet", this should be fine.

PR #464, opened 2024-11-05
Fixes for ts-high level sdk

wjthieme:
### Fix creation of pools that require vault accounts with extensions

It calculates space for base token accounts only at the moment. When you try to create a pool (and thus vaults) with a token that has an extension that requires data on the account, that is currently not taken into account for create pool.

### Improve ata creation function

* Check if required balances are met for non-native mint
* Actually simulate transactions using bankrun in tests
* Only wrap the required amount with `ata` strategy if there is already a balance in the ata
* Fix a few minor bugs that became apparent when actually sending the TXs in the tests

### Rename SOL_WRAPPING_STRATEGY to NATIVE_MINT_WRAPPING_STRATEGY

Since whirlpools is multichain `SOL_...` is no longer correct. Native mint on Eclipse is ETH.

### Add pool price to fetch pool functions

Currently only returns the sqrtPrice and tickIndex. A different package (@orca-so/whirlpools-core) is needed to turn that into a normal price. For convenience I added it


ts-sdk/whirlpool/src/token.ts
```diff
@@ -280,8 +311,46 @@ export function getCurrentTransferFee(
...
 
+/**
+ * Builds the required account extensions for a given mint.
+ *
+ * https://github.com/solana-labs/solana-program-library/blob/3844bfac50990c1aa4dfb30f244f8c13178fc3fa/token/program-2022/src/extension/mod.rs#L1276
+ *
+ * @param {Mint} mint - The mint account to build extensions for.
+ * @returns {ExtensionArgs[]} An array of extension arguments.
+ */
+export function getAccountExtensions(mint: Mint): ExtensionArgs[] {
```
yugure-orca, line 322: nit: I think this function works well for token vault (non ATA). so it is not an issue for now.
If this is used for ATA, `ImmutableOwner` will use a bit space ?

yugure-orca (approved):
🚀 (with nit comment)

ts-sdk/whirlpool/src/token.ts
wjthieme: Yep but since ATAs are initialized through the ATA program we don't have to worry about reserving space and calculating rent. I'll add a comment to this function though

PR #491, opened 2024-11-11
Small improvement to fetchPositionsForOwner

wjthieme:
Internally we used the property `owner` to store the `tokenProgram` in a `Token` object. This might create confusion since `Token` already has `owner` which is the owner of the token account. 

Externally this doesn't change anything since the `Token` object is not returned by the function.


yugure-orca (approved):
LGTM

PR #492, opened 2024-11-11
Add whirlpools cli package

wjthieme:
Copy paste from our internal repo with some minor changes:
* Use `tsx` instead of `ts-node` (which allows top-level await)
* Added index file that just prompts for which command to run
* Use `prompts` instead of `prompt`


yugure-orca (approved):
Thank you! 🚀

PR #501, opened 2024-11-13
Add unit tests for createPool.ts

calintje:
**Title**
Add unit tests for createPool.ts

**Details**
- Fixed the calculation of `estInitializationCosts `in the `createConcentratedLiquidityPool `function.
- Added tests to verify the correct calculation of estInitialization costs.
- Updated setupWhirlpool to support the V2 instruction of our program.


ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -204,9 +205,17 @@ export async function createConcentratedLiquidityPoolInstructions(
     }),
   );
 
-  stateSpace += getTokenSize(getAccountExtensions(mintA.data));
-  stateSpace += getTokenSize(getAccountExtensions(mintB.data));
-  stateSpace += getWhirlpoolSize();
+  stateSpaces.push(
+    tokenProgramA === TOKEN_2022_PROGRAM_ADDRESS
+      ? getTokenSize(getAccountExtensions(mintA.data))
```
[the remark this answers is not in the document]
wjthieme: Suspect we will need the same thing in increase_liquidity too. We can put it in token.ts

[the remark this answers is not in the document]
wjthieme: And then we can write a test for it 😄

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,143 @@
...
+      createConcentratedLiquidityPoolInstructions(rpc, mintA, mintB, 64, 1),
+      {
+        name: 'AssertionError',
+        message: 'Either supply a funder or set the default funder'
+      }
+    );
+    setDefaultFunder(signer);
+  });
+
+  it("Should throw an error if token mints are not ordered", async () => {
```
wjthieme, line 47: nit: they are ordered, just not ordered correctly

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,143 @@
...
+    await assert.rejects(
+      createConcentratedLiquidityPoolInstructions(rpc, mintB, mintA, 64, 1),
+      {
+        name: 'AssertionError',
```
wjthieme, line 51: Not sure if it exists but is there a new AssertionError("Token order needs to be flipped to match the canonical ordering (i.e. sorted on the byte repr. of the mint pubkeys)")

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,143 @@
...
+        message: 'Token order needs to be flipped to match the canonical ordering (i.e. sorted on the byte repr. of the mint pubkeys)'
```
wjthieme, line 52: nit: we use double quotes for strings in ts (but you can run yarn format and that should format your files correctly

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,143 @@
...
+
+  it("Should create valid splash pool instructions", async () => {
+    const price = 10;
+    const sqrtPrice = priceToSqrtPrice(price, 6, 6)
+
+    const { instructions, poolAddress } =
+      await createSplashPoolInstructions(rpc, mintA, mintB, price);
+    await sendTransaction(instructions);
+
+    const pool = await fetchWhirlpool(rpc, poolAddress);
```
wjthieme, line 65: I would structure the test like this:
* create splash pool
* fetch maybe pool before
* fetch signer balance before
* assert pool does not exist
* send tx
* fetch maybe pool
* fetch signer blance after
* assert pool exists
* assert sqrt, mintA, mintB, and tick spacing
* assert that estInitialization cost equals difference in signer balance

That way you don't need the `Should estimate initialization costs correctly` test case entirely

ts-sdk/whirlpool/tests/utils/program.ts
```diff
@@ -3,16 +3,20 @@ import {
...
 import { generateKeyPairSigner } from "@solana/web3.js";
-import { sendTransaction, signer } from "./mockRpc";
+import { rpc, sendTransaction, signer } from "./mockRpc";
 import {
   SPLASH_POOL_TICK_SPACING,
   WHIRLPOOLS_CONFIG_ADDRESS,
 } from "../../src/config";
 import { tickIndexToSqrtPrice } from "@orca-so/whirlpools-core";
+import { fetchMint } from "@solana-program/token";
+import { a } from "vitest/dist/chunks/suite.B2jumIFP.js";
```
wjthieme, line 19: ?

ts-sdk/whirlpool/tests/createPool.test.ts
wjthieme: And then create more tests with similar structure for:
* Splash pool with TE token
* CL pool
* CL pool with TE token
* etc.

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -242,12 +243,19 @@ export async function createConcentratedLiquidityPoolInstructions(
...
-    stateSpace += getTickArraySize();
+    stateSpaces.push(getTickArraySize());
   }
 
-  const nonRefundableRent = await rpc
-    .getMinimumBalanceForRentExemption(BigInt(stateSpace))
-    .send();
+  const nonRefundableRents: Lamports[] = await Promise.all(
+    stateSpaces.map(async (space) => {
+      const rentExemption = await calculateMinimumBalance(rpc, space);
```
wjthieme, line 251: This is still doing the same thing where it fetches the rent 5 (or 6 times).

The way I would do it. Change the function to:

```
export function calculateMinimumBalance(
  rent: Rent,
  dataSize: number,
): Lamports {
```

Fetcht he rent sysvar at the beginning of this function and then instead of keeping track of `stateSpace: Array<bigint>` you can just straight up keep track of `estInitializationCost: Lamports`.

And then instead of doing `stateSpace.push(...)` you can just do estInitializationCost += calculateMinimumBalance(rent, ...)

ts-sdk/whirlpool/src/sysvar.ts
```diff
@@ -0,0 +1,28 @@
...
+const ACCOUNT_STORAGE_OVERHEAD = 128;
+
+/**
+ * Calculates the minimum balance required for rent exemption for a given account size.
+ *
+ * @param {Rpc} rpc - The Solana RPC client to fetch sysvar rent data.
+ * @param {number} dataSize - The size of the account data in bytes.
+ * @returns {Promise<BigInt>} The minimum balance required for rent exemption in lamports.
+ */
+export async function calculateMinimumBalance(
```
wjthieme, line 17: Can you write a test for this?

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,263 @@
...
+    await assert.rejects(
+      createConcentratedLiquidityPoolInstructions(rpc, mintB, mintA, 64, 1),
+    );
+  });
+
+  it("Should create splash pool", async () => {
+    const price = 10;
+    const sqrtPrice = priceToSqrtPrice(price, 6, 6);
+
+    let signerAccount = await rpc.getAccountInfo(signer.address).send();
```
wjthieme, line 49: Think nicer if we implement getBalance on the mockRpc so we don't have to duplicate this every time.

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,263 @@
...
+    const maybePool = await fetchMaybeWhirlpool(rpc, poolAddress);
+    assert.strictEqual(maybePool.exists, false);
+
+    await sendTransaction(instructions);
+
+    const pool = await fetchMaybeWhirlpool(rpc, poolAddress);
+    assertAccountExists(pool);
+
+    signerAccount = await rpc.getAccountInfo(signer.address).send();
+    const balanceAfter = signerAccount.value?.lamports ?? lamports(0n);
```
wjthieme, line 64: nit: I would group all the fetch together before starting on any assertions. poolBefore together with balanceBefore, poolAfter together with balanceAfter.

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,263 @@
...
+    const balanceChange = balanceBefore - balanceAfter;
+    const txFee = 15000n; // 3 signing accounts * 5000 lamports
+    const minRentExempt = balanceChange - txFee;
+    assert.strictEqual(estInitializationCost, minRentExempt);
```
wjthieme, line 68: Let me see if I can make the tx fee go to 0 in bankrun so that this assertion can just become:
```
assert.strictEqual(estInitializationCost, balanceBefore - balanceAfter);
```

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,263 @@
...
+
+    assert.strictEqual(sqrtPrice, pool.data.sqrtPrice);
+    assert.strictEqual(mintA, pool.data.tokenMintA);
+    assert.strictEqual(mintB, pool.data.tokenMintB);
+    assert.strictEqual(SPLASH_POOL_TICK_SPACING, pool.data.tickSpacing);
+  });
+
+  it("Should create splash pool with 1 TE token", async () => {
+    const mint1 = await setupMint();
```
wjthieme, line 77: I would set up these mints in the beforeAll as well since that will get faster execution (init the TE mint once opposed to doing it for each TE test)

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,263 @@
...
+    assert.strictEqual(estInitializationCost, minRentExempt);
+
+    assert.strictEqual(sqrtPrice, pool.data.sqrtPrice);
+    assert.strictEqual(mintC, pool.data.tokenMintA);
+    assert.strictEqual(mintD, pool.data.tokenMintB);
+    assert.strictEqual(SPLASH_POOL_TICK_SPACING, pool.data.tickSpacing);
+  });
+
+  it("Should create splash pool with 2 TE tokens", async () => {
+    const mint1 = await setupMintTEFee();
```
wjthieme, line 111: For the tests I would do (instead of trying to combine Transfer fee and 1/2 TE token tests):
* 1 TE token (without extensions)
* 2 TE tokens (without extensions)
* 1 TE token (with transfer fee)

ts-sdk/whirlpool/src/sysvar.ts
```diff
@@ -0,0 +1,28 @@
+import { fetchSysvarRent } from "@solana/sysvars";
```
wjthieme, line 1: nit: This should be re-exported by @solana/web3.js so you can use that as import

ts-sdk/whirlpool/src/sysvar.ts
```diff
@@ -12,17 +10,15 @@ const ACCOUNT_STORAGE_OVERHEAD = 128;
  *
  * @param {Rpc} rpc - The Solana RPC client to fetch sysvar rent data.
  * @param {number} dataSize - The size of the account data in bytes.
- * @returns {Promise<BigInt>} The minimum balance required for rent exemption in lamports.
+ * @returns {bigint} The minimum balance required for rent exemption in lamports.
  */
-export async function calculateMinimumBalance(
-  rpc: Rpc<GetAccountInfoApi>,
+export function calculateMinimumBalance(
```
wjthieme, line 15: nit: maybe we can be a bit more descriptive with this naming. `getMinimumBalanceForRentExemption` is what the rpc method that does the same is called

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -9,26 +9,34 @@ import {
...
   beforeAll(async () => {
-    const mint1 = await setupMint();
-    const mint2 = await setupMint();
-    [mintA, mintB] = orderMints(mint1, mint2);
+    mint1 = await setupMint();
+    mint2 = await setupMint();
+    mintTE1 = await setupMintTE();
+    mintTE2 = await setupMintTE();
+    mintTEFee1 = await setupMintTEFee();
+    mintTEFee2 = await setupMintTEFee();
```
wjthieme, line 35: Do you really need to have this second TEFee mint?

ts-sdk/whirlpool/tests/sysvar.test.ts
```diff
@@ -0,0 +1,41 @@
...
+import { setupAta, setupMint } from "./utils/token";
+import { fetchMint } from "@solana-program/token-2022";
+import { rpc } from "./utils/mockRpc";
+import { getTokenSizeForMint } from "../src/token";
+
+describe("Sysvar", () => {
+  let rent: SysvarRent;
+
+  beforeAll(async () => {
+    rent = {
```
wjthieme, line 15: This doesn't necessarily have to be in the beforeAll (since it is not asynchronous). You can just init the rent directly as a const

ts-sdk/whirlpool/tests/sysvar.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      lamportsPerByteYear: lamports(3480n),
+      exemptionThreshold: 2.0,
+      burnPercent: 0,
+    };
+  });
+
+  it("Should calculate the correct minimum balance for a token account", async () => {
+    const mint = await setupMint();
```
wjthieme, line 23: You can make this test synchronous since you don't need to store/fetch anything in bankrun

ts-sdk/whirlpool/tests/utils/mockRpc.ts
```diff
@@ -103,7 +103,19 @@ const decoders: Record<string, VariableSizeDecoder<string>> = {
   base64: getBase64Decoder(),
 };
 
-async function getAccountData<T>(address: unknown, opts: unknown): Promise<T> {
+interface AccountData {
```
wjthieme, line 106: Not 100% sure if the AccountData is always exactly in this format. IIRC data depends on the encoding specified and is not always [string, string]. lamports/rentEpoch might be numbers I think (hence why I put generic T).

Instead of adding this interface you could also just do an assertion in `getBalance` to check if `lamports` exists in the object and then return that:

```
    const accountDataForBalance = await getAccountData(
        addressForBalance,
        config.payload.params[1],
      );
      assert("lamports" in accountDataForBalance);
      return getResponseWithContext<T>(accountDataForBalance.lamports);
 ```

**Review by wjthieme — APPROVED** [review]

File: ts-sdk/whirlpool/tests/createPool.test.ts [hunk 20]
```diff
@@ -1,5 +1,301 @@
...
+
+  it("Should create splash pool", async () => {
+    const price = 10;
+    const sqrtPrice = priceToSqrtPrice(price, 6, 6);
+
+    const { instructions, poolAddress, estInitializationCost } =
+      await createSplashPoolInstructions(rpc, mintA, mintB, price);
+
+    const balanceBefore = await rpc.getBalance(signer.address).send();
+    const maybePool = await fetchMaybeWhirlpool(rpc, poolAddress);
```
wjthieme, line 70: nit: I would name them poolBefore and poolAfter

ts-sdk/whirlpool/src/token.ts
```diff
@@ -33,6 +32,8 @@ import {
   getCreateAccountWithSeedInstruction,
   getTransferSolInstruction,
 } from "@solana-program/system";
+import { getTokenSize } from "@solana-program/token";
+import { getTokenSize as getToken22Size } from "@solana-program/token-2022";
```
wjthieme, line 36: nit: maybe getTokenWithExtensionsSize since if it is a t22 token with no extensions we still the other one

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,301 @@
...
+
+    await sendTransaction(instructions);
+
+    const pool = await fetchMaybeWhirlpool(rpc, poolAddress);
+    const balanceAfter = await rpc.getBalance(signer.address).send();
+    const balanceChange = balanceBefore.value - balanceAfter.value;
+    const txFee = 15000n; // 3 signing accounts * 5000 lamports
+    const minRentExempt = balanceChange - txFee;
+
+    assertAccountExists(pool);
```
wjthieme, line 108: Interesting! Never knew this existed

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -1,5 +1,301 @@
...
+    assert.strictEqual(estInitializationCost, minRentExempt);
+    assert.strictEqual(sqrtPrice, pool.data.sqrtPrice);
+    assert.strictEqual(mintTEA, pool.data.tokenMintA);
+    assert.strictEqual(mintTEB, pool.data.tokenMintB);
+    assert.strictEqual(SPLASH_POOL_TICK_SPACING, pool.data.tickSpacing);
+  });
+
+  it("Should create splash pool with 1 TE tokens (with extension)", async () => {
```
wjthieme, line 143: nit: maybe instead of `with extension` be explicit about that it is the TE extension

ts-sdk/whirlpool/tests/sysvar.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  const OVERHEAD = 128n;
+
+  it("Should calculate the correct minimum balance for a token account", () => {
+    const tokenSize = 165;
+    const calcultatedMinimumBalance = calculateMinimumBalanceForRentExemption(
+      rent,
+      tokenSize,
+    );
+
+    const expectedMinimumBalance = lamports((165n + OVERHEAD) * 10n);
```
wjthieme, line 22: I wouldn't do the calculation here and just assert agains the actual value

wjthieme: So just 1445n

ts-sdk/whirlpool/tests/token.test.ts
```diff
@@ -476,4 +477,23 @@ describe("Token Account Creation", () => {
...
+  });
+
+  it("Should get the correct token size for TOKEN_2022_PROGRAM mint", async () => {
+    const mint = await setupMintTE();
+    const mintAccount = await fetchMint(rpc, mint);
+    const tokenSize = getTokenSizeForMint(mintAccount);
+    assert.strictEqual(tokenSize, 165);
+  });
+
+  it("Should get the correct token size for TOKEN_2022_PROGRAM mint with", async () => {
```
wjthieme, line 494: nit: `mint with ...`

wjthieme (approved):

PR #502, opened 2024-11-14
Add set reward authority cli commands

yugure-orca:
Added the following cli commands:

- setRewardEmissionsSuperAuthority
- setRewardAuthority

Those commands have been used to adjust authorities on the config and pools in Eclipse.
(It worked as expected)

### DIFF
--- a/legacy-sdk/cli/src/utils/transaction_sender.ts
+++ b/legacy-sdk/cli/src/utils/transaction_sender.ts
@@ -7,7 +7,7 @@ import {
 } from "@orca-so/common-sdk";
 import Decimal from "decimal.js";
 import base58 from "bs58";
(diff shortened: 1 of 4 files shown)


legacy-sdk/cli/src/utils/transaction_sender.ts
```diff
@@ -31,7 +31,7 @@ export async function sendTransaction(
     let priorityFeeInLamports = 0;
 
     while (true) {
-      const priorityFeeInSOL = await promptNumber("priorityFeeInSOL");
+      const priorityFeeInSOL = await promptText("priorityFeeInSOL");
```
yugure-orca, line 34: I cannot type "0.000001" SOL, so changed to `promptText`.

PR #510, opened 2024-11-15
Add Example Usage to Fetch All Whirlpool Accounts with Specified Filters

calintje:
**Title**
Add Example Usage to Fetch All Whirlpool Accounts with Specified Filters

**Details**
Fetching all pools is a much requested feature by developers. Adding an example of how this can be done using our SDK makes it easier for DAs to point developers in the right direction.


yugure-orca (approved):
Thanks! 🚀

PR #513, opened 2024-11-16
Rustdocs high-level Rust SDK

calintje:
**Title**
Rustdocs high-level Rust SDK

**Details**
- Rustdocs
- Update `set_whirlpools_config_address`, including test
- Update typedocs
  - removed new wallet creation from code examples like decrease liquidity and close position.
  - other fixes

Screenshot 2024-11-18 100356

Screenshot 2024-11-18 100439


rust-sdk/whirlpool/src/config.rs
```diff
@@ -22,8 +57,19 @@ pub static WHIRLPOOLS_CONFIG_ADDRESS: Mutex<Pubkey> = Mutex::new(DEFAULT_WHIRLPO
 pub static WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Mutex<Pubkey> =
     Mutex::new(DEFAULT_WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS);
 
+/// Input type for setting the Whirlpools configuration.
+pub enum WhirlpoolConfigInput {
```
wjthieme, line 61: nit:

Functionally the same, but instead of two enums you can do a single one:

```rust
pub enum WhirlpoolsConfigAddress {
  SolanaMainnet,
  SolanaDevnet,
  EclipseMainnet,
  EclipseTestnet
}

impl Into<Pubkey> for WhirlpoolsConfigAddress {
  pub fn into(val: self) -> Pubkey {
    ...
  }
```

Or can do it the other way around
```rust
pub enum WhirlpoolsConfigInput {
  Address(Pubkey)
  SolanaMainnet,
  SolanaDevnet,
  EclipseMainnet,
  EclipseTestnet
}

impl From<Pubkey> for WhirlpoolsConfigInput {
  fn from(val: Pubkey) -> Self {
    WhirlpoolsConfigInput::Address(val)
  }
}
```

This looks nicer for the caller as they can just do:
```rust
pub fn set_whirlpools_config_addres(address: Pubkey)

set_whirlpools_config_addres(WhirlpoolConfigInput::SolanaMainnet.into())
set_whirlpools_config_addres(Pubkey.from_str("..."))
```
Or if you choose option two:
```rust
pub fn set_whirlpools_config_addres(input: WhirlpoolConfigInput)

set_whirlpools_config_addres(WhirlpoolConfigInput::SolanaMainnet)
set_whirlpools_config_addres(Pubkey.from_str("...").into())
```

Instead, the current implementation results in:
```rust
set_whirlpools_config_addres(WhirlpoolConfigInput::Network(Network::SolanaMainnet))
set_whirlpools_config_addres(WhirlpoolConfigInput::Address(Pubkey.from_str("..."))

File: rust-sdk/whirlpool/src/config.rs [hunk 2]
```diff
@@ -9,6 +9,41 @@ pub const DEFAULT_WHIRLPOOLS_CONFIG_ADDRESS: Pubkey = Pubkey::new_from_array([
...
+    SolanaDevnet,
+    EclipseMainnet,
+    EclipseTestnet,
+}
+
+impl Network {
+    pub fn config_address(&self) -> Pubkey {
+        match self {
+            Network::SolanaMainnet => DEFAULT_WHIRLPOOLS_CONFIG_ADDRESS,
+            Network::SolanaDevnet => Pubkey::new_from_array([
```
wjthieme, line 24: In this case probably not that big of a deal in this case, but since rust is a compiled language there is a performance improvement we can make by defining the addresses as `const`. A `const` is evaluated at compile-time instead of at runtime meaning that they are faster at runtime.
```rust
const SOLANA_MAINNET_WHIRLPOOLS_CONFIG_ADDRESS = Pubkey::new_from_array([...]);
const SOLANA_DEVNET_WHIRLPOOLS_CONFIG_ADDRESS = Pubkey::new_from_array([...]);
const ECLIPSE_MAINNET_WHIRLPOOLS_CONFIG_ADDRESS = Pubkey::new_from_array([...]);
const ECLIPSE_TESTNET_WHIRLPOOLS_CONFIG_ADDRESS = Pubkey::new_from_array([...]);

impl Network {
  pub fn config_address(&self) -> Pubkey {
    match self {
      Network::SolanaMainnet => SOLANA_MAINNET_WHIRLPOOLS_CONFIG_ADDRESS,
      Network::SolanaDevnet => SOLANA_DEVNET_WHIRLPOOLS_CONFIG_ADDRESS,
      Network::EclipseMainnet => ECLIPSE_MAINNET_WHIRLPOOLS_CONFIG_ADDRESS,
      Network::EclipseTestnet => ECLIPSE_TESTNET_WHIRLPOOLS_CONFIG_ADDRESS,
    }
  }
}
```

rust-sdk/whirlpool/src/config.rs
calintje: This is nice! From a users perspective, I like the second option better, since I think most people will use Orca's Whirlpools when using the SDK.

rust-sdk/whirlpool/src/config.rs
```diff
@@ -22,8 +22,52 @@ pub static WHIRLPOOLS_CONFIG_ADDRESS: Mutex<Pubkey> = Mutex::new(DEFAULT_WHIRLPO
...
+/// Input type for setting the Whirlpools configuration.
+pub enum WhirlpoolsConfigInput {
+    Address(Pubkey),
+    SolanaMainnet,
+    SolanaDevnet,
+    EclipseMainnet,
+    EclipseTestnet,
+}
+
+impl From<Pubkey> for WhirlpoolsConfigInput {
```
wjthieme, line 55: Might be nice to also implement Into<Pubkey> for WhirlpoolsConfigInput. That way you can put the `match` statement in there (making it public)

rust-sdk/whirlpool/src/create_pool.rs
```diff
@@ -41,6 +41,60 @@ pub struct CreatePoolInstructions {
...
+///
+/// # Example
+///
+/// ```rust
+/// use solana_client::rpc_client::RpcClient;
+/// use solana_sdk::{pubkey::Pubkey, signer::Keypair};
+/// use orca_whirlpools::create_splash_pool_instructions
+/// use std::str::FromStr;
+///
+/// set_whirlpools_config_address(WhirlpoolsConfigInput::SolanaDevnet).unwrap()
```
wjthieme, line 78: `.unwrap` in rust results in a `panic` which works different from throwing in ts. A panic stops the code and cannot be recovered.

I guess for script-style code this is fine (since you just want your script to stop on an error) but for production-environments you generally want to handle errors.

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -27,20 +27,96 @@ use crate::{
 
 // TODO: support transfer hooks
 
+/// Represents the parameters for decreasing liquidity in a pool.
+///
+/// You must specify only one of the parameters (`TokenA`, `TokenB`, or `Liquidity`).
+/// Based on the provided value, the SDK computes the other two parameters.
+///
+/// # Variants
```
wjthieme, line 35: Here you can put the rustdoc comment next to the Variant itself I think and it will then show up nicely in the rustodc

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -27,20 +27,96 @@ use crate::{
...
 #[derive(Debug, Clone)]
 pub enum DecreaseLiquidityParam {
     TokenA(u64),
     TokenB(u64),
     Liquidity(u128),
 }
 
+/// Represents the instructions and quote for decreasing liquidity in a position.
+///
+/// # Fields
```
wjthieme, line 49: Same here I potentially you can put the comment next to the field itself

ts-sdk/whirlpool/src/position.ts
```diff
@@ -85,8 +85,7 @@ function getPositionInBundleAddresses(
  * import { generateKeyPairSigner, createSolanaRpc, devnet } from '@solana/web3.js';
  *
  * const devnetRpc = createSolanaRpc(devnet('https://api.devnet.solana.com'));
- * const wallet = await generateKeyPairSigner();
- * await devnetRpc.requestAirdrop(wallet.address, lamports(1000000000n)).send();
+ * const wallet = await generateKeyPairSigner(); // USE YOUR OWN WALLET
```
wjthieme, line 88: Think here you can just do address("INSERT_WALLET_ADDRESS"). Since it doesn't have to be a signer

rust-sdk/whirlpool/src/decrease_liquidity.rs
calintje: 👌🏻

## Before
Screenshot 2024-11-18 154537

## After
Screenshot 2024-11-18 154611

rust-sdk/whirlpool/src/config.rs
```diff
@@ -3,27 +3,73 @@ use std::{error::Error, sync::Mutex};
...
+
+/// The Whirlpools program's config account address for Eclipse Testnet.
+const ECLIPSE_TESTNET_WHIRLPOOLS_CONFIG_ADDRESS: Pubkey = Pubkey::new_from_array([
+    213, 230, 107, 150, 137, 123, 254, 203, 164, 137, 81, 181, 70, 54, 172, 140, 176, 39, 16, 72,
+    150, 84, 130, 137, 232, 108, 97, 236, 197, 119, 201, 83,
+]);
+
 /// The default address for the Whirlpools program's config extension account.
-pub const DEFAULT_WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Pubkey = Pubkey::new_from_array([
+pub const SOLANA_MAINNET_WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Pubkey = Pubkey::new_from_array([
```
wjthieme, line 31: nit: don't think this has to be public

wjthieme (approved):
🚢  but let's hold off merging until #428 is merged as to not interfere with reviews there

PR #522, opened 2024-11-19
Fix lint staged

wjthieme:
Replaced lint-staged with just checking for both lint and format in CI


yugure-orca (approved):
LGTM

changes on programs/whirlpool are no logic change, but compiler generates different binary. (so solana-verify hash changed)

PR #532, opened 2024-11-21
Add unit tests for increaseLiquidity.ts

calintje:
**Title**
Add unit tests for increaseLiquidity.ts

**Details**
- Add setupPosition and setupTEPosition functions to program.ts
- Write unit tests for external functions

### DIFF
--- a/ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
+++ b/ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, beforeAll, vi, expect } from "vitest";
+import { describe, it, beforeAll } from "vitest";
 import {
   increaseLiquidityInstructions,
   openPositionInstructions,
@@ -9,15 +9,26 @@ import { setupMint, setupAta } from "./utils/token";
 import {
   fetchMaybePosition,
   fetchPosition,
-  fetchWhirlpool,
   getPositionAddress,
 } from "@orca-so/whirlpools-client";
 import { fetchToken } from "@solana-program/token-2022";
 import { address, assertAccountExists, type Address } from "@solana/web3.js";
 import assert from "assert";
-import { setupPosition, setupTEPosition, setupWhirlpool } from "./utils/program";
-import { DEFAULT_FUNDER, setDefaultFunder, SPLASH_POOL_TICK_SPACING } from "../src/config";
-import { setupAtaTE, setupMintTE, setupMintTEFee } from "./utils/tokenExtensions";
+import {
+  setupPosition,
+  setupTEPosition,
+  setupWhirlpool,
+} from "./utils/program";
+import {
+  DEFAULT_FUNDER,
+  setDefaultFunder,
+  SPLASH_POOL_TICK_SPACING,
+} from "../src/config";
+import {
+  setupAtaTE,
+  setupMintTE,
+  setupMintTEFee,
+} from "./utils/tokenExtensions";
 
 describe("Increase Liquidity Instructions", () => {
   const tickSpacing = 64;
@@ -48,12 +59,12 @@ describe("Increase Liquidity Instructions", () => {
     whirlpools["Token-TE Token"] = await setupWhirlpool(
       mintA,
       mintTEA,
-      tickSpacing
+      tickSpacing,
     );
     whirlpools["TE Token-TE Token"] = await setupWhirlpool(
       mintTEA,
       mintTEB,
-      tickSpacing
+      tickSpacing,
     );
     whirlpools["Token-TE Token with Transfer Fee extension"] =
       await setupWhirlpool(mintA, mintTEFee, tickSpacing);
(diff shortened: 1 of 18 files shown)


ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
```diff
@@ -1,5 +1,480 @@
...
+describe("Increase Liquidity Instructions", () => {
+  const tickSpacing = 64;
+  const tokenBalance = 1_000_000n;
+  let mintA: Address;
+  let mintB: Address;
+  let mintTEA: Address;
+  let mintTEB: Address;
+  let mintTEFee: Address;
+  let ataMap: Record<string, Address> = {};
+  let whirlpools: Record<string, Address> = {};
```
wjthieme, line 42: Took me a little bit to figure out what is what here. Maybe it makes sense to do:
```
let ataMap: Record<string, Address> = {}; // key is the mint address
let poolAToB: Address
let poolAToTEA: Address
// etc.
let equallySpacedPositions:  Record<string, Address> = {}; // key is the whirlpool address
let oneSidedPositionsA:  Record<string, Address> = {}; // key is the whirlpool address
let oneSidedPositionsB:  Record<string, Address> = {}; // key is the whirlpool address
```

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
```diff
@@ -1,5 +1,480 @@
...
+
+    assert.strictEqual(quote.tokenEstA, balanceChangeTokenA);
+    assert.strictEqual(quote.tokenEstB, balanceChangeTokenB);
+    assert.strictEqual(quote.liquidityDelta, position.data.liquidity);
+  };
+
+  it("Should handle liquidity increase for whirlpool with tokenA=Token and tokenB=Token, and Centered Position", async () => {
+    await testLiquidityIncrease(
+      positions["Token-Token"][0],
+      address(mintA),
```
wjthieme, line 168: nit: Aren't these already and address? Why do you need to convert to address again?

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
```diff
@@ -1,5 +1,480 @@
...
+    positionMint: Address,
+    tokenA: Address,
+    tokenB: Address,
+  ) => {
+    const amount = 10_000n;
+
+    const { quote, instructions } = await increaseLiquidityInstructions(
+      rpc,
+      positionMint,
+      { tokenA: amount },
```
wjthieme, line 145: nit: Maybe we should do `liquidity` instead of tokenA?

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
```diff
@@ -1,5 +1,480 @@
...
+    await assert.rejects(
+      increaseLiquidityInstructions(rpc, positionMint, {
+        tokenA: tokenAAmount,
+      }),
+    );
+    setDefaultFunder(signer);
+  });
+});
+
+describe("Open Position Instructions", () => {
```
wjthieme, line 320: nit: would it make sense to put this in a separate file?

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
wjthieme: The pools you are only ever getting with a static string (not dynamic) so there is no strict need for them to be in an object

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
```diff
@@ -1,5 +1,480 @@
...
+        {
+          tickLower: 100,
+          tickUpper: 200,
+        },
+      ),
+    ];
+  });
+
+  const testLiquidityIncrease = async (
+    positionMint: Address,
```
wjthieme, line 136: idea: since this actually only needs the positionMint (tokenA/tokenB address we can derive from the account data) we could potentially do the `for (mint in PositionMints) { it("abcee") }` approach.

For example in the before all if you do something like this:

```ts
const positions: Map<string, Address> = new Map(); // key is a human-readable name

beforeAll(async () => {
  const mintA = ...
  const mintB = ...
  // etc.
  
  const mintPairs = [
    [mintA, mintB]
    // etc.
  ]

  const pools = mintPairs.map(x => /* init pool */);

  const positionTypes = [
    { lowerTick: -100, upperTick: 100 },
    { lowerTick: -100, upperTick: -1 },
    { lowerTick: 1, upperTick: 100 }
  ] 

  for (const [name, pool] in pools) {
    for (const positionType in positionTypes) {
      positions.set(`${name} #{positionType}`, /* setupPosition */) 
    }
  }
})

for (const [name, positionMint] of positions.entries()) {
  it(`Increase liquidity ${name}`, () => {
    // write the test flow here
  })
}

// add the other tests
```

wjthieme (commented):
Nice! Could nits and ideas to think about but the suite itself looks solid

ts-sdk/whirlpool/tests/increaseLiquidity.test.ts
calintje: Makes sense

wjthieme (approved):
🚢 🚢 🚢

PR #535, opened 2024-11-21
Make `compute_swap` function public

### LINKED ISSUE #534: Weird `TickArraySequence::new` crash in multithread ffi context
We're integrating `orca_whirlpools_core` in an iOS project via rust ffi; To avoid blocking UI, we call `swap_quote_by_input_token` in a background thread (concurrent queue).

image

These are the findings when I debug:

1. It won't crash when compile with release mode, by default we build and debug with debug mode
2. It won't crash if we call it synchronously in main thread
3. It won't crash if we replace `TickArraySequence::new` with `TickArraySequence {}` (we sorted tick array outside and make `tick_arrays` and `tick_spacing` public)

```diff
- let tick_sequence = TickArraySequence::new(tick_arrays.into(), whirlpool.tick_spacing)?;
+ let tick_sequence = TickArraySequence {
+        tick_arrays: tick_arrays.into(),
+        tick_spacing: whirlpool.tick_spacing,
+    };
```

I tried version 0.3.1 with this fix https://github.com/orca-so/whirlpools/pull/521 included. could you please take a look? what could be the issue? my guess is the memory layout or something breaks in multithread + ffi @wjthieme @yugure-orca 

If possible, can I make `compute_swap` function public? this gives 3rd party developer more flexibilities.

0xh3rman:
resolves #534

Make `TickArraySequence` and `compute_swap` public, workaround for #534, this also gives flexibility to 3rd party developer to compute swap quote.


rust-sdk/core/src/quote/swap.rs
```diff
@@ -150,13 +150,31 @@ pub fn swap_quote_by_output_token(
...
+///    - `true`: Swap from token A to token B.
+///    - `false`: Swap from token B to token A.
+/// - `specified_input`: Determines if the input amount is specified:
+///    - `true`: `token_amount` represents the input amount.
+///    - `false`: `token_amount` represents the output amount.
+/// - `_timestamp`: A placeholder for future full swap logic, currently ignored.
+///
+/// # Returns
+/// A `Result` containing a `SwapResult` struct if the swap is successful, or an `ErrorCode` if the computation fails.
+pub fn compute_swap<const SIZE: usize>(
```
wjthieme, line 177: Not necessarily opposed to making this public but `compute_swap` misses two important parts. Might be nice to highlight that in the rustdoc comment:
* No slippage is being taken in to account (does not compute the maxSent or minReceived amounts)
* Transfer fee extension is not being taken in to account

0xh3rman: Added notes to comments

0xh3rman:
Should I add this root `Cargo.toml`? rust complains when I build it
```toml
[profile.release]
overflow-checks = true
```

wjthieme:
> overflow-checks = true

What is the error that you are getting? The math in this lib is setup up using checked_math if there is a chance of overflowing. Adding overflow-checks in prod makes the code less efficient

0xh3rman:
ah ok if this is an intended behavior, I saw a warning when I built it

0xh3rman:
@wjthieme @yugure-orca let's merge this?

wjthieme (approved):

0xh3rman:
Thanks for the merge, could you please tag a new release? 🙏

PR #557, opened 2024-11-29
Update github docs workflow

calintje:
**Title**
Update github `docs` workflow so that it builds preview sites for new branches

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,34 +2,52 @@ name: Docs
 
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - "*"
 
 concurrency:
-  group: "docs"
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -2,34 +2,52 @@ name: Docs
...
+      if: github.ref_name == 'main'
+      id: deployment
+      uses: actions/deploy-pages@v4
+      with:
+        branch: gh-pages
+
+    - name: Deploy preview
+      if: github.ref_name != 'main'
+      run: |
+        PREVIEW_DIR="previews/${GITHUB_REF_NAME##*/}"
```
wjthieme, line 50: How exactly does this work? All I see this doing is creating a new folder and moving the docs output folder there

PR #563, opened 2024-12-02
Add Token Testing for Rust Sdk

pauldragonfly:
# Token Testing Infrastructure Refactor

## Overview
Refactored token testing infrastructure to support Token-2022 program and improve test maintainability. Split token-related functionality into separate modules to handle different token standards and testing scenarios more effectively.

### Test Structure Reorganization
- Moved utility code to `tests/utils/` for better separation of concerns
- Split token operations into:
  - `token.rs`: Standard SPL Token operations
  - `token_extensions.rs`: Token-2022 specific features
- Introduced dedicated `token_test.rs` for comprehensive token testing

### Token Testing Enhancements
- Added support for different native SOL wrapping strategies:
  - ATA-based wrapping
  - Keypair-based temporary accounts
  - Seed-based deterministic accounts
- Implemented Token-2022 extension testing:
  - Transfer fee configuration
  - Extension type management
  - Custom mint configurations

### RPC Mock Improvements
- Enhanced mock RPC to track token operations
- Added support for Token-2022 program interactions
- Improved account state management for testing


rust-sdk/whirlpool/Cargo.toml
```diff
@@ -28,6 +28,7 @@ orca_whirlpools_client = { path = "../client", features = ["core-types"] }
 bincode = { version = "^1.3" }
 serde = { version = "^1.0" }
 serde_json = { version = "^1.0" }
+tokio = { version = "1.0", features = ["sync"] }
```
wjthieme, line 31: Is this needed as a normal dep or can it be a dev-dep?

rust-sdk/whirlpool/src/tests/utils/rpc.rs
```diff
@@ -188,6 +213,66 @@ impl RpcContext {
         let signature = self.rpc.send_transaction(&transaction).await?;
         Ok(signature)
     }
+
+    pub async fn create_token_2022_mint(
```
wjthieme, line 217: Might be nice to put this function in a separate `token_extensions` file (just like we have the `token` file). If we add everything here this file will become very bloated

rust-sdk/whirlpool/src/tests/token_test.rs
```diff
@@ -0,0 +1,126 @@
+use super::utils::{rpc::RpcContext, token::*, token_extensions::*};
+use crate::NativeMintWrappingStrategy;
+use serial_test::serial;
+use solana_sdk::signer::Signer;
+use spl_token::native_mint;
+use spl_token_2022::{extension::ExtensionType, ID as TOKEN_2022_PROGRAM_ID};
+
+#[tokio::test]
```
wjthieme, line 8: I think preference would be to have tests in the same file as the code itself (which is what we are doing for the other rust packages as well). One exception of course would be the e2e test since that doesn't really relate to a single file/function

rust-sdk/whirlpool/src/tests/token_test.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    let result = prepare_token_accounts_instructions(
+        &ctx,
+        ctx.signer.pubkey(),
+        vec![TokenAccountStrategy::WithoutBalance(native_mint::ID)],
+    )
+    .await
+    .unwrap();
+
+    assert_eq!(result.create_instructions.len(), 1); // Create ATA
```
wjthieme, line 35: This currently only asserts the length of the instructions (but the instructions might be wrong). Instead we can actually send the instructions to BanksClient (like we do in the e2e tests) we can assert the pre- and post- state of BanksClient. That should give us more confidence that the instructions created by the prepare_token_accunts_instructions are actually correct/valid

rust-sdk/whirlpool/src/tests/token_test.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    // Verify extension creation
+    let calls = ctx.token_calls.read().await;
+    assert!(calls.contains(&"create_token_2022_mint".to_string()));
+
+    // Verify account data
+    let account = ctx.rpc.get_account(&mint).await.unwrap();
+    assert!(account.data.len() > 82); // Size is larger due to extension
+    assert_eq!(account.owner, TOKEN_2022_PROGRAM_ID);
+}
```
wjthieme, line 126: Might be nice to also Add a few tests here:
* Test if the withBalance asserts the balance for non-native tokens
* Case if the token account already exists (we want to make sure it doesn't try to create it again)

You can have a look at `ts-sdk/whirlpool/tests/token.test.ts` as that is pretty extensive

rust-sdk/whirlpool/Cargo.toml
pauldragonfly: Good point, need to delete

rust-sdk/whirlpool/src/tests/utils/rpc.rs
pauldragonfly: Yep, I think that's more native rust way, just tried to follow ts-sdk structure but I can fix

rust-sdk/whirlpool/src/tests/token_test.rs
pauldragonfly: Ok update to put it into token.rs

rust-sdk/whirlpool/src/tests/token_test.rs
pauldragonfly: Ok added more sophisticated asserts

rust-sdk/whirlpool/src/tests/token_test.rs
pauldragonfly: Added more tests similar to token.test.ts

rust-sdk/whirlpool/src/token.rs
```diff
@@ -333,4 +342,233 @@ mod tests {
...
+    #[tokio::test]
+    #[serial]
+    async fn test_token_2022_account() {
+        let ctx = RpcContext::new().await;
+
+        // Create basic Token-2022 mint (without transfer fee)
+        let mint = setup_mint_te(&ctx, &[]).await.unwrap();
+        let mint_account = ctx.rpc.get_account(&mint).await.unwrap();
+
+        // Verify account data
```
wjthieme, line 379: Does this test any of the sdk functions or only the test_utils function? If the second don't think we actually need this test since BanksClient will just fail if the instructions in `setup_mint_te` are bad

rust-sdk/whirlpool/src/token.rs
```diff
@@ -333,4 +342,233 @@ mod tests {
...
+
+        let result = prepare_token_accounts_instructions(
+            &ctx.rpc,
+            ctx.signer.pubkey(),
+            vec![TokenAccountStrategy::WithoutBalance(native_mint::ID)],
+        )
+        .await
+        .unwrap();
+
+        assert_eq!(result.create_instructions.len(), 1); // Create ATA
```
wjthieme, line 425: This right here is only asserting the length of the instructions which only checks if there are instructions but does not check if said instructions are correct.

What I would do:
* Take the BanksClient state before `does the token account for NATIVE_MINT exist?`
* Simulate the create instructions using BanksClient (`ctx.send_transaction_with_signers` or `ctx.send_transaction`).
* Take the BanksClient state after `does the token account for NATIVE_MINT exist now?`
* Simulate the cleanup instructions using BanksClient
* Take the BanksClient state after `does the token account for NATIVE_MINT exist now?`

This way:
* We know that the instructions are valid cause they can be simulated by BanksClient
* Know that the instructions do what we want them to do based on the outcome of the instruction: `initialize token account(s) with or without balance checks`.

pauldragonfly: Ok updated!

rust-sdk/whirlpool/src/token.rs
pauldragonfly: Great point, it is testing only utils function indeed. Updated to make work for actual functionality.

wjthieme (commented):
Great start! Got a couple nits but nothing major related to the setup.

We should still add more tests cases:
* Other native token wrapping methods (with and without a balance specified cause they  they actuall wrap/unwrap tokens based on the amount specified)
* Token without a balance specified
* Token 2022
* etc.

rust-sdk/whirlpool/src/tests/token_extensions.rs
```diff
@@ -1,19 +1,124 @@
...
+    instructions.push(system_instruction::create_account(
+        &ctx.signer.pubkey(),
+        &mint.pubkey(),
+        rent,
+        space as u64,
+        &TOKEN_2022_PROGRAM_ID,
+    ));
+
+    // 2. Initialize extensions
+    if extensions.contains(&ExtensionType::TransferFeeConfig) {
```
wjthieme, line 51: nit: Loop and match?

rust-sdk/whirlpool/src/tests/token_extensions.rs
```diff
@@ -1,19 +1,124 @@
...
+        &TOKEN_2022_PROGRAM_ID,
+        &mint.pubkey(),
+        &ctx.signer.pubkey(),
+        None, // freeze_authority
+        6,    // decimals
+    )?);
+
+    // 4. Set newer transfer fee if needed
+    if extensions.contains(&ExtensionType::TransferFeeConfig) {
+        instructions.push(set_transfer_fee(
```
wjthieme, line 73: loop and match?

rust-sdk/whirlpool/src/token.rs
```diff
@@ -105,7 +108,14 @@ pub(crate) async fn prepare_token_accounts_instructions(
         }
 
         let existing_balance = if let Some(account_info) = &ata_account_infos[i] {
-            Account::unpack(&account_info.data)?.amount
+            if mint_account_infos[i].owner == TOKEN_2022_PROGRAM_ID {
+                // For Token-2022 accounts, use StateWithExtensions
+                let account_state = StateWithExtensions::<Account>::unpack(&account_info.data)?;
```
wjthieme, line 113: Nice!

rust-sdk/whirlpool/src/token.rs
```diff
@@ -333,4 +352,393 @@ mod tests {
...
+        )
+        .await
+        .unwrap();
+
+        // Verify account wasn't modified
+        let final_account = ctx.rpc.get_account(&ata).await.unwrap();
+        assert_eq!(initial_account.data, final_account.data);
+    }
+
+    // 3. Native Token Tests
```
wjthieme, line 494: Only thing I would still add is Native Token Tests with Balance (since they wrap/unwrap tokens from SOL > wSOL)

wjthieme (commented):
Looking good! Almost there!

rust-sdk/whirlpool/src/token.rs
```diff
@@ -664,15 +664,34 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    #[serial]
+    async fn test_native_token_balance() -> Result<(), Box<dyn Error>> {
+        let ctx = RpcContext::new().await;
+        let amount = 1_000_000_000; // 1 SOL
+
+        // Setup native token (wSOL) account with balance using helper function
+        let wsol_ata = setup_ata_with_amount(&ctx, native_mint::ID, amount).await?;
```
wjthieme, line 674: Is this testing a SDK function or a function from the test utils? In case the second I think we can skip this test

rust-sdk/whirlpool/src/token.rs
```diff
@@ -664,15 +664,34 @@ mod tests {
...
+        assert_eq!(token_account.owner, ctx.signer.pubkey());
+
+        Ok(())
+    }
+
     // 4. Token-2022 Tests
     #[tokio::test]
     #[serial]
-    async fn test_token_2022_extensions() {
+    async fn test_token_2022_extensions() -> Result<(), Box<dyn Error>> {
```
wjthieme, line 689: Didn't know you could return result from tokio tests that's kinda nice

wjthieme (approved):

PR #564, opened 2024-12-02
Add Pool Test for Rust-sdk

pauldragonfly:
# Add pool test utilities and test cases

## Overview
Add pool initialization test utilities and corresponding test cases to validate core pool functionality, mirroring the TypeScript SDK's test coverage.

## Core changes
Implement setup_whirlpool utility function that:
- Creates token vaults and initializes pool with specified token pair and tick spacing
- Handles both standard and concentrated liquidity pool configurations
- Added mock implementation of `getProgramAccounts` to handle both extensions and normal PDA

Add test cases to verify:

- Splash pool initialization and configuration
- Concentrated liquidity pool initialization and configuration
- Pool state validation including liquidity, tick spacing, and fee rates

Implement test_fetch_all_pools_for_pair test case:

- Add mock implementation of getProgramAccounts RPC method with filter support
- Enable testing of pool discovery functionality previously unavailable in ts-sdk
- Verify correct handling of both initialized and uninitialized pools

Fix bug in pool discovery:

- Update pool fetching logic to properly iterate through fee tiers
- Ensure uninitialized pools are correctly identified and returned


rust-sdk/whirlpool/src/pool.rs
```diff
@@ -380,3 +380,186 @@ pub async fn fetch_whirlpools_by_token_pair(
...
+
+    impl TestContext {
+        async fn new() -> Result<Self, Box<dyn Error>> {
+            let ctx = RpcContext::new().await;
+            let mint_a = setup_mint_with_decimals(&ctx, 9).await?;
+            let mint_b = setup_mint_with_decimals(&ctx, 9).await?;
+
+            setup_ata_with_amount(&ctx, mint_a, 500_000_000_000).await?;
+            setup_ata_with_amount(&ctx, mint_b, 500_000_000_000).await?;
+
```
wjthieme, line 408: nit: think if we set up the two pools (splash pool + cl pool) from here we don't have to repeat the logic every time

rust-sdk/whirlpool/src/tests/rpc.rs
```diff
@@ -306,14 +351,108 @@ async fn send(
                 feature_set: Some(version.feature_set),
             })?
         }
+        "getProgramAccounts" => {
```
wjthieme, line 354: I think this might be a slippery slope to do our own implementation of `getProgramAccounts`. This opens us up to several vectors where we might mess up if we forget something:
* Forget to add an account to the accounts hashmap (which is currently already happening for all the accounts initialized in tests/token.rs and tests/program.rs)
* Our implementation of memcmp might differ from the actual rpc-implementation (and there is currently no way of knowing if it does).

With this I am leaning to just doing what we did in in the ts tests: skip the tests that require gpa and add a comment saying that BanksClient doesn't support gpa and therefor we decided to leave that function untested

pauldragonfly: Good point, I get where you're coming from. I initially thought it’d be fine since this is just for testing, but yeah, doing our own GPA implementation does feel a bit risky—especially if we accidentally miss things or introduce inconsistencies with how the actual RPC works.

I’m okay with the idea of doing what the TS tests do for now: skip the GPA-dependent tests and add a comment explaining the limitation with BanksClient. It’s not ideal, but it’s clear and avoids the hassle of maintaining a custom implementation.

rust-sdk/whirlpool/src/pool.rs
pauldragonfly: Yep, fixed to use presetup pools

wjthieme (approved):
🚢🚢🚢

PR #588, opened 2024-12-11
Users unable to close positions on ME/SOL pool without harvesting first

pauldragonfly:
Description:
This PR fixes a constraint error that occurs when calling closePosition on a position that has uninitialized rewards. The error message indicates that the reward_owner_account.mint does not match the expected mint address stored in whirlpool.reward_infos[reward_index].mint. This issue arises because closePosition attempts to collect rewards for all possible reward indexes, including those that are not initialized.

**Problem Statement:**
When attempting to close a position using closePosition, a constraint error occurs if the whirlpool has uninitialized rewards. The error is due to the program trying to collect rewards for uninitialized reward indexes, resulting in mismatched mint addresses.

**Solution:**
Modify the getClosePositionIx method in whirlpool-impl.ts to skip any uninitialized rewards when constructing the instructions for collectReward. This is achieved by filtering out reward indexes where the mint is PublicKey.default, which indicates that the reward is uninitialized.

**Steps:**
Filtering Uninitialized Rewards: The code now filters out any rewards where info.mint equals PublicKey.default, which signifies that the reward has not been initialized. This ensures that only initialized rewards are processed.
Maintaining Correct Indexing: By mapping over rewardInfos with both info and index, we preserve the correct reward index when processing the rewards.

**Edge Case Handling:**
Uninitialized Rewards: Skipping uninitialized rewards prevents the constraint error caused by attempting to collect rewards that do not exist.
Initialized Rewards with Zero Emission: Rewards that are initialized but have zero emission are still processed. If there's any accumulated reward (e.g., from past emissions), it will be collected. This ensures that no rewards are missed.

**Why Edge Cases Are Not a Problem:**
No Loss of Rewards: Since we only skip uninitialized rewards, there is no risk of missing out on any accrued rewards. All initialized rewards, regardless of their emission status, are still collected.
Consistency with collectReward Behavior: The collectReward function already handles uninitialized rewards gracefully by not attempting to collect them. By aligning closePosition with this behavior, we ensure consistency across the SDK.
Successful Execution: With this change, calling closePosition no longer results in a constraint error, even when some rewards are uninitialized.

**Testing:**
Manual Testing: Verified that closePosition works as expected when:
All rewards are initialized.
Some rewards are uninitialized.
Rewards have zero emission.

### DIFF
--- a/legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
+++ b/legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
@@ -13,8 +13,7 @@ import {
   TOKEN_2022_PROGRAM_ID,
   TOKEN_PROGRAM_ID,
 } from "@solana/spl-token";
-import type { PublicKey } from "@solana/web3.js";
-import { Keypair } from "@solana/web3.js";
+import { PublicKey, Keypair } from "@solana/web3.js";
 import invariant from "tiny-invariant";
 import type { WhirlpoolContext } from "../context";
 import type {
@@ -599,14 +598,16 @@ export class WhirlpoolImpl implements Whirlpool {
     const shouldDecreaseLiquidity = positionData.liquidity.gtn(0);
 
     const rewardsToCollect = this.data.rewardInfos
-      .filter((_, i) => {
-        return (
-          (rewardsQuote.rewardOwed[i] ?? ZERO).gtn(0) ||
-          // we need to collect reward even if all reward will be deducted as transfer fee
-          (rewardsQuote.transferFee.deductedFromRewardOwed[i] ?? ZERO).gtn(0)
-        );
-      })
-      .map((info) => info.mint);
+        .map((info, i) => ({ info, index: i }))
+        .filter(({ info, index }) => {
+            if (info.mint.equals(PublicKey.default)) {
+                return false;
+            }
+            return (
+                (rewardsQuote.rewardOwed[index] ?? ZERO).gtn(0) ||
+                (rewardsQuote.transferFee.deductedFromRewardOwed[index] ?? ZERO).gtn(0)
+            );
+        });
 
     const shouldCollectRewards = rewardsToCollect.length > 0;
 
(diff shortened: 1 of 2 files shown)


legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
```diff
@@ -599,14 +598,18 @@ export class WhirlpoolImpl implements Whirlpool {
     const shouldDecreaseLiquidity = positionData.liquidity.gtn(0);
 
     const rewardsToCollect = this.data.rewardInfos
-      .filter((_, i) => {
+      .map((info, i) => ({ info, index: i }))
+      .filter(({ info, index }) => {
+        if (info.mint.equals(PublicKey.default)) {
```
odcheung, line 603: we can use PoolUtil.isRewardInitialized here

legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
```diff
@@ -599,14 +598,18 @@ export class WhirlpoolImpl implements Whirlpool {
     const shouldDecreaseLiquidity = positionData.liquidity.gtn(0);
 
     const rewardsToCollect = this.data.rewardInfos
-      .filter((_, i) => {
+      .map((info, i) => ({ info, index: i }))
+      .filter(({ info, index }) => {
```
odcheung, line 602: looks like we are mapping to get the index to provide filter with the index. Does the filter's index paramter help in this case?

odcheung (commented):
nice! you found it

legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
pauldragonfly: Good point, fixed!

legacy-sdk/whirlpool/src/impl/whirlpool-impl.ts
pauldragonfly: Also fixed

wjthieme (approved):
🚢🚢🚢

PR #592, opened 2024-12-13
Add support for anchor v0.28 to rust client

wjthieme:
Two issues:
* anchor v0.28 was not supported even though there is no reason why it couldn't be. I relaxed the version to also allow 0.28
* Because anchor is still in pre-release (0.x.x) versions 0.28, 0.29 and 0.30 are not semver compatible. The `^0.29` that was specified therefore only allowed version 0.29. I changed this to go from 0.28 to 0.30 (including).
* Updated the integration test to make sure the versions are pinned correctly (and not have two versions of the dependencies in question)


rust-sdk/core/Cargo.toml
```diff
@@ -18,7 +18,7 @@ floats = ["dep:libm"]
 
 [dependencies]
 ethnum = { version = "^1.1" }
-libm = { version = "^0.1", optional = true }
+libm = { version = "^0.2", optional = true }
```
yugure-orca, line 21: quick Q: Why we requre 0.2 ? (old is 0.1).
I don't think 0.2 is a bad idea, but I am concerned that the requirement has been raised here while other areas have been relaxed.

yugure-orca (approved):
with 1 quick Q.

odcheung (approved):

rust-sdk/core/Cargo.toml
wjthieme: Yeah this is the same issue as with the anchor dep. 0.1 and 0.2 are not semver compatible (even though both would work with the library). I should be able to change it to `>=0.1, <0.3` which might be better

PR #602, opened 2024-12-18
[DO NOT MERGE] batchUpdateProtocolFeeRate

yugure-orca:
add temporary CLI command to update protocol fee setting


yugure-orca:
The script worked (but will not merge as standard CLI commands)

PR #601, opened 2024-12-17
Add rust-sdk position test

pauldragonfly:
**Add Position and Position Bundle Test Coverage**

**Overview**
- Introduces additional position and position bundle tests, aligned with the TypeScript SDK tests.

**Changes**
- Implemented `setup_position` and `setup_position_bundle` for SPL and Token-2022.
- Added tests for:
  - Position ownership and retrieval
  - Position bundle creation and retrieval

**Notes**
- Tests are currently `#[ignore]` due to missing `getProgramAccounts` support in solana-bankrun.
- Once support is added, these tests can be enabled by removing the `#[ignore]` attributes.

PR #605, opened 2024-12-19
Add increase liquidity test for rust sdk

pauldragonfly:
# Add tests for increase liquidity and update RPC configurations

## Changes
1. Added tests for increase liquidity functionality
2. Added configuration to RPC setup
3. Updated existing position and pool tests to use configuration-based setup
4. Updated program to ensure proper functioning of setup position and setup te position operations

## Details
- Implemented test cases covering various scenarios for increase liquidity
- Refactored existing tests to align with new configuration pattern
- Fixed program logic for position setup operations


rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -695,3 +695,82 @@ pub async fn open_position_instructions(
...
+            Some(ctx.signer.pubkey()),
+        )
+        .await?;
+
+        // send transaction
+        let signers: Vec<&Keypair> = increase_ix.additional_signers.iter().collect();
+
+        ctx.send_transaction_with_signers(increase_ix.instructions, signers)
+            .await?;
+
```
wjthieme, line 773: Here I would assert that the quote matches whatever happened on-chain (same way we have the test in ts sdk)

rust-sdk/whirlpool/src/tests/program.rs
```diff
@@ -160,11 +265,12 @@ pub async fn setup_te_position(whirlpool: Pubkey) -> Result<Pubkey, Box<dyn Erro
...
     }
     .instruction(OpenPositionInstructionArgs {
-        tick_lower_index,
-        tick_upper_index,
+        tick_lower_index: lower_tick_index,
+        tick_upper_index: upper_tick_index,
         position_bump,
     });
 
+    println!("Sending transaction with instructions...");
```
wjthieme, line 273: This print can be removed?

rust-sdk/whirlpool/src/tests/program.rs
```diff
@@ -237,3 +343,13 @@ pub async fn setup_position_bundle(
 
     Ok(position_bundle_address)
 }
+
+pub async fn setup_config_and_fee_tiers(ctx: &RpcContext) -> Result<(), Box<dyn Error>> {
```
wjthieme, line 347: Is this used anywhere? it is not actually setting up any fee tiers and/or configs

rust-sdk/whirlpool/src/tests/rpc.rs
```diff
@@ -40,6 +40,7 @@ use crate::{SPLASH_POOL_TICK_SPACING, WHIRLPOOLS_CONFIG_ADDRESS};
 pub struct RpcContext {
     pub rpc: RpcClient,
     pub signer: Keypair,
+    pub config: Pubkey,
```
wjthieme, line 43: Is this param used anywhere?

rust-sdk/whirlpool/src/increase_liquidity.rs
pauldragonfly: the issue is, to assert the quote match, need to fetch the position from chain and need GPA implementation. Actually in the ts it is not checking actual quote match but just checking adding successfully or any errors.

rust-sdk/whirlpool/src/tests/program.rs
pauldragonfly: Removed

rust-sdk/whirlpool/src/tests/program.rs
pauldragonfly: Right, Removed

rust-sdk/whirlpool/src/tests/rpc.rs
pauldragonfly: Yes, it is used in the test context to setup whirlpool. It needs config.

pauldragonfly: actually you are right, it is not necessary, removing all

rust-sdk/whirlpool/src/increase_liquidity.rs
wjthieme: Why would it need GPA? GPA is only for if you don't know the addresses of the accounts you are fetching. In this case you do know the addresses right?

pauldragonfly: Oh... you are right. I updated one more test set to check actual quote. I separated those to make sure both function itself and inner is properly checked

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -695,3 +695,82 @@ pub async fn open_position_instructions(
...
+    const TEST_SLIPPAGE_TOLERANCE: u16 = 100; // 1%
+
+    #[tokio::test]
+    #[serial]
+    async fn test_increase_liquidity() -> Result<(), Box<dyn Error>> {
+        let ctx = RpcContext::new().await;
+
+        // setup and initialize
+        setup_config_and_fee_tiers(&ctx).await?;
```
wjthieme: Is this needed? I believe gthat the config and fee tiers get written already when bankrun starts

rust-sdk/whirlpool/src/increase_liquidity.rs
wjthieme: Where?

rust-sdk/whirlpool/src/tests/program.rs
wjthieme: It's still here?

rust-sdk/whirlpool/src/tests/rpc.rs
wjthieme: It is still here?

rust-sdk/whirlpool/src/increase_liquidity.rs
pauldragonfly: test_increase_liquidity_with_quote_verification this function

pauldragonfly: Actually, cleaned up all tests and merge to consolidate the multiple tests. Verify quote, liquidity now.

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -695,3 +695,157 @@ pub async fn open_position_instructions(
...
+    }
+
+    async fn get_token_balance(rpc: &RpcClient, address: Pubkey) -> Result<u64, Box<dyn Error>> {
+        // For an SPL account that we already know, just fetch + unpack
+        let account_data = rpc.get_account(&address).await?;
+        let token_account = spl_token::state::Account::unpack(&account_data.data)?;
+        Ok(token_account.amount)
+    }
+
+    const TEST_SLIPPAGE_TOLERANCE: u16 = 100; // 1%
```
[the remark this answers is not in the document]
pauldragonfly: Yes updated to check full combo of pools

rust-sdk/whirlpool/src/tests/program.rs
```diff
@@ -149,7 +124,7 @@ pub async fn setup_position(
         owner: owner,
         position: position_pubkey,
         position_mint: position_mint.pubkey(),
-        position_token_account,
+        position_token_account: Pubkey::default(), // instruction will create
```
wjthieme: Same comment here as in #601

pauldragonfly: Fixed!

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -695,3 +695,205 @@ pub async fn open_position_instructions(
...
+        minted.insert("TEFee", mint_te_fee);
+
+        let mut user_atas: HashMap<&str, Pubkey> = HashMap::new();
+        user_atas.insert("A", user_ata_a);
+        user_atas.insert("B", user_ata_b);
+        user_atas.insert("TEA", user_ata_te_a);
+        user_atas.insert("TEB", user_ata_te_b);
+        user_atas.insert("TEFee", user_ata_tefee);
+
+        let pool_combos = vec![
```
wjthieme, line 850: Would it make sense to split this up in separate tests like we are doing in TS? There is probably some macro that we can use for that

pauldragonfly: I can separate the tests by pool type or combos, but unlike typescript tests, I don't see simple way to generate full combo tests with macros. Do we have any similar approach somewhere?

pauldragonfly: Actually found some ways. I added rstest for cases (still cannot loop-based generate, but much better now)

pauldragonfly: If you think this way looks good, I'll fix all other tests also

wjthieme (approved):
🚀

PR #629, opened 2025-01-03
Fix a bug with the cyclic fee/rewards quote calculation

wjthieme:
Right now the new core sdk mishandles the cyclic aspect of fee and rewards quotes. Currently it for the checkpoint calculation, it wraps around u256 while it should be wrapping around u128.

I've fixed the issue and added a test case

### DIFF
--- a/rust-sdk/core/src/quote/fees.rs
+++ b/rust-sdk/core/src/quote/fees.rs
@@ -225,7 +225,7 @@ mod tests {
     }
 
     #[test]
-    fn test_cyclic_checkpoint_growth() {
+    fn test_cyclic_growth_checkpoint() {
         let position = PositionFacade {
             liquidity: 91354442895,
             tick_lower_index: 15168,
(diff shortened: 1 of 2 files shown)


rust-sdk/core/src/quote/fees.rs
```diff
@@ -63,14 +63,16 @@ pub fn collect_fees_quote(
         .wrapping_sub(fee_growth_below_b)
         .wrapping_sub(fee_growth_above_b);
 
-    let fee_owed_delta_a: U256 = <U256>::from(fee_growth_inside_a)
-        .wrapping_sub(position.fee_growth_checkpoint_a.into())
+    let fee_growth_delta_a = fee_growth_inside_a.wrapping_sub(position.fee_growth_checkpoint_a);
+
+    let fee_growth_delta_b = fee_growth_inside_b.wrapping_sub(position.fee_growth_checkpoint_b);
+
+    let fee_owed_delta_a: U256 = <U256>::from(fee_growth_delta_a)
```
odcheung, line 70: in the contract, if this overflows (it does a u128.mul(u128)), we just default to 0. Is this something we can do here to avoid the u256 wrapping?

wjthieme: It is supposed to wrap (on u128) cause the fee growth is cyclical. This implies should match legacy sdk

odcheung (approved):

Matar-Masw21 (commented):
Check out

PR #630, opened 2025-01-06
docs: rename function get_positions_for_owner to fetch_positions_for_owner

srgchrksv:
The get_positions_for_owner is not available in orca_whirlpools and fetch_positions_for_owner is working.


wjthieme (approved):
Thanks for the contribution!

PR #648, opened 2025-01-13
Add harvest test for Rust sdk

pauldragonfly:
Add Harvest Position Tests for Multi-Pool + Rewards

**Scope**: Added tests in harvest.rs to validate fee and reward collection for multi-pool setups.

**Key Features:**
Multi-Pool Coverage: Tested various pool types (e.g., A-B, A-TE) with different position ranges (centered, one-sided) for both SPL & Token-2022 mints.
Fee Generation: Incorporated swaps to accumulate fees, then verified fee_owed_a, fee_owed_b, and rewards during harvest.
Reward Collection: Validated time/emission-based reward accumulation and balances after harvesting.
Implementation:


wjthieme (commented):
Can we do the same thing with `rstest`? Otherwise looks good to me!

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -734,115 +736,171 @@ mod tests {
...
+                &ctx,
+                &dec_ix,
+                user_ata_for_token_a,
+                user_ata_for_token_b,
+                position_mint,
+            )
+            .await
+            .unwrap();
+
+            println!(
```
wjthieme, line 900: Do we still need these printlns? They different configs now show up as separate tests right?

wjthieme (approved):
One small nit (that also applies to the other rstests) but LGTM!

PR #652, opened 2025-01-14
Add Swap test for Rust SDK

pauldragonfly:
**Add Swap Test for Whirlpool**
**Key Features:**

Pool Creation:
Created a concentrated liquidity pool.
Created a full-range liquidity pool.

Swap Functionality:
Implemented tests for multiple swap routes.

Balance Validation:
Verified balances for both input and output tokens.


rust-sdk/whirlpool/src/swap.rs
```diff
@@ -323,3 +323,267 @@ pub async fn swap_instructions(
...
+                Ok(parsed.base.amount)
+            } else {
+                let parsed = TokenAccount::unpack(&account_data.data)?;
+                Ok(parsed.amount)
+            }
+        }
+
+        pub async fn init_pool(&self, is_splash: bool) -> Result<Pubkey, Box<dyn Error>> {
+            if is_splash {
+                let pool = create_splash_pool_instructions(
```
wjthieme, line 386: Better to use the `setup_whirlpool` instead of the sdk function here?

The risk is that if we use the sdk function in setup the test is not fully isolated (meaning errors in create_splash_pool_instructions, etc. might show up as errors in the swap test)

rust-sdk/whirlpool/src/swap.rs
```diff
@@ -323,3 +323,267 @@ pub async fn swap_instructions(
...
+            println!(
+                "swap result => a_to_b={}, used_a={}, used_b={}, gained_a={}, gained_b={}",
+                a_to_b, used_a, used_b, gained_a, gained_b
+            );
+
+            Ok(())
+        }
+    }
+
+    #[tokio::test]
```
wjthieme, line 514: Can we do rstest here?

rust-sdk/whirlpool/src/swap.rs
wjthieme: Same goes for setup_position

rust-sdk/whirlpool/src/swap.rs
pauldragonfly: Updated!

rust-sdk/whirlpool/src/swap.rs
pauldragonfly: Yep make sense. I copied e2e test but I think valid. Should we update e2e also?

wjthieme (approved):
🚀

PR #660, opened 2025-01-17
Add close position test for rust sdk

pauldragonfly:
Add close position test for rust sdk

Key feature :
- Add multiple combo test for close position


wjthieme (commented):
Nice! Think only the TS close position test does not have fees yet. Can you add as well?

wjthieme (approved):
🚀

PR #667, opened 2025-01-20
tx-sender public package

parhim:
ts package for building and sending transactions using web3.js 2.0
tests can be ran with yarn test

I tried a bunch of different building approaches but cant seem to figure out how to preserve docstring fn descriptions

tested with whirlpools instructions and works pretty well, though for some reason, during testing websocket based tx confirmations are like 5 to 10x slower.

the main thing to review is package structure and configs

added a readme with an example

https://linear.app/orca-so/issue/DX-152/send-transaction-library


ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,230 @@
...
+  last150Slots.sort((a, b) => Number(a.slot - b.slot));
+  const last50Slots = last150Slots.slice(-50);
+  const nonZeroFees = last50Slots.filter((slot) => slot.prioritizationFee > 0);
+  if (nonZeroFees.length === 0) return BigInt(0);
+  const sum = nonZeroFees.reduce(
+    (acc, slot) => acc + slot.prioritizationFee,
+    BigInt(0)
+  );
+
+  return sum / BigInt(nonZeroFees.length);
```
wjthieme, line 216: Median might be better than average as it is less susceptible to outliers

ts-sdk/tx-sender/src/jito.ts
```diff
@@ -0,0 +1,51 @@
...
+  const response = await fetch(
+    "https://bundles.jito.wtf/api/v1/bundles/tip_floor"
+  );
+
+  if (!response.ok) {
+    throw new Error("Failed to fetch recent Jito tips");
+  }
+  const data = await response.json().then((res) => res[0]);
+  cachedJitoTip = lamports(
+    BigInt(Math.floor(Number(data.landed_tips_50th_percentile) * 10 ** 9))
```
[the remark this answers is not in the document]
wjthieme: Can we have it so that the user can also specify `50ema` which is also in the response. Might be nice for some use-cases

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @throws {Error} If signing fails
+ *
+ * @example
+ * const signedTx = await signTransactionMessage(
+ *   transactionMessage,
+ *   [signer1, signer2]
+ * );
+ */
+
+async function signTransactionMessage(
```
wjthieme, line 178: All singers need to already be in the instructions. If a singer is supplied as a noopSigner it basically means that the user needs to add the signature manually. I suspect an extra sign function like this is possibly redundant

wjthieme (commented):
First draft looks great!

ts-sdk/tx-sender/src/config.ts
```diff
@@ -0,0 +1,156 @@
...
+const DEFAULT_PRIORITIZATION: TransactionConfig = {
+  priorityFee: {
+    type: "dynamic",
+    maxCapLamports: BigInt(4_000_000), // 0.004 SOL
+  },
+  jito: {
+    type: "dynamic",
+    maxCapLamports: BigInt(4_000_000), // 0.004 SOL
+  },
+  chainId: "solana",
```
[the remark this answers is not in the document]
parhim: that wouldnt be robust, especially with proxies. theres no way to know unless theres some frozen account with some value that never changes that can be fetched

ts-sdk/tx-sender/src/buildTransaction.ts
```diff
@@ -0,0 +1,181 @@
...
+  let message = await generateTransactionMessage(instructions, rpc, signer);
+
+  if (lookupTableAddresses?.length) {
+    const lookupTableAccounts = await fetchAllMaybeAddressLookupTable(
+      rpc,
+      lookupTableAddresses
+    );
+    const tables = lookupTableAccounts.reduce(
+      (prev, account) => {
+        assertAccountExists(account);
```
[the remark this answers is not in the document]
parhim: wouldnt that tx still fail later on ? or do we expect users to supply lookup tables even when theyre not needed ?

ts-sdk/tx-sender/src/config.ts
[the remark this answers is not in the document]
wjthieme: genesisHash?

wjthieme:
Another thing I thought of that would be nice to have:
* if you try to send a tx that is not fully singed. Instead of failing directly we can simulate the tx and return that result. This behavior can possibly be turned on/off from config. A few internal scripts currently have a mechanism like this and we call it simulationOnly or dryRun mode

ts-sdk/tx-sender/src/config.ts
```diff
@@ -0,0 +1,153 @@
...
+    };
+
+export type TransactionConfig = {
+  jito: FeeSetting;
+  priorityFee: FeeSetting;
+  computeUnitMarginMultiplier: number;
+  priorityFeePercentile: Percentile;
+};
+
+export type Percentile = "25" | "50" | "50ema" | "75" | "95" | "99";
```
wjthieme, line 146: JitoPercentile? For priority fee percentile it can take any number

ts-sdk/tx-sender/src/config.ts
```diff
@@ -0,0 +1,153 @@
...
+  connectionContext?: ConnectionContext;
+}) => {
+  globalConfig = {
+    transactionConfig: config.transactionConfig || DEFAULT_PRIORITIZATION,
+    connectionContext: config.connectionContext,
+  };
+};
+
+export async function setRpc(
+  url: string,
```
wjthieme, line 50: nit: Maybe nice to do `string | Rpc`?

ts-sdk/tx-sender/src/config.ts
```diff
@@ -0,0 +1,153 @@
...
+  supportsPriorityFeePercentile: boolean = false,
+  chain?: ChainId
```
wjthieme, line 52: nit: I wouldn't let a user overwrite the chainId as that can only lead to problems. Genesis hash should always give the correct chainId

ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,272 @@
...
+  transactionConfig: TransactionConfig,
+  connectionContext: ConnectionContext,
+  signer: TransactionSigner
+) {
+  const { rpcUrl, chainId } = connectionContext;
+  const { jito, priorityFee, priorityFeePercentile } = transactionConfig;
+  const rpc = rpcFromUrl(rpcUrl);
+
+  if (jito.type !== "none") {
+    if (chainId === "solana") {
```
wjthieme, line 59: I would do the chainid check in the `processJitoTipForTxMessage` function itself

ts-sdk/tx-sender/src/compatibility.ts
```diff
@@ -0,0 +1,23 @@
...
+  const transport = createDefaultRpcTransport({ url });
+  const rpc = createRpc({ api, transport });
+  return rpc;
+}
+
+function normalizeAddresses(addresses?: (string | Address)[]): Address[] {
+  return addresses?.map((addr) => address(addr)) ?? [];
+}
+
+export { normalizeAddresses, rpcFromUrl };
```
wjthieme, line 23: Can you remove all the export statements to just directly export the functions?

ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,272 @@
...
+    } else {
+      console.warn(
+        "Jito tip is not supported on this chain. Skipping jito tip."
+      );
+    }
+  }
+  let computeUnits = await getComputeUnitsForTxMessage(rpc, message);
+
+  if (!computeUnits) throw new Error("Transaction simulation failed");
+  // add margin to compute units
```
wjthieme, line 75: Still needs to be implemented?

ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,272 @@
...
+      );
+    } else {
+      console.warn(
+        "Jito tip is not supported on this chain. Skipping jito tip."
+      );
+    }
+  }
+  let computeUnits = await getComputeUnitsForTxMessage(rpc, message);
+
+  if (!computeUnits) throw new Error("Transaction simulation failed");
```
wjthieme, line 74: Should we just default to the max (= 1.4M) if simulation fails? There might be scenarios where simulation would fail and we still want to build transaction (i.e. when firing multiple txs that depend on each other)

ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,272 @@
...
+  return processPriorityFeeForTxMessage(
+    message,
+    computeUnits,
+    transactionConfig,
+    connectionContext,
+    priorityFeePercentile
+  );
+}
+
+async function processPriorityFeeForTxMessage(
```
wjthieme, line 86: nit :maybe this can be called processComputeBudgetForTxMessage and move it to a file called computeBudget.ts. That way:
* priorityFee = either jito tip or computeBudget
* jitoTip = just jito tip
* computeBudget = just computeBudget

ts-sdk/tx-sender/src/priorityFees.ts
```diff
@@ -0,0 +1,272 @@
...
+    );
+  } else {
+    const rpc = rpcFromUrl(rpcUrl);
+    const recent = await rpc
+      .getRecentPrioritizationFees(writableAccounts)
+      .send();
+    const nonZero = recent
+      .filter((pf) => pf.prioritizationFee > 0)
+      .map((pf) => pf.prioritizationFee);
+    const sorted = nonZero.sort((a, b) => Number((a - b) / BigInt(1_000_000)));
```
wjthieme, line 186: Why `(a - b) / BigInt(1_000_000)`? In a lot of cases this would become 0

ts-sdk/tx-sender/src/priorityFees.ts
parhim: forgot to remove comment

ts-sdk/tx-sender/src/priorityFees.ts
parhim: good idea

ts-sdk/tx-sender/src/config.ts
parhim: with the new rpc, we're unable to get the url from the object and for some use cases the url is needed (triton priority fees)

parhim:
todo from calins comment:
config.ts
it would make more sense if the types, getters and setters are named the same way.
Example 1: We now have ConnectionContext, setRpc and getConnectionContext. I think I prefer RpcConfig, setRpcConfig and getRpcConfig better (since we also have globalConfig and not globalContext.
Example 2: we have getPriorityConfig which gets the TransactionConfig . but we have 4 different setters for the TransactionConfig: setPriorityFeePercentile, setPriorityFeeSetting,  setJitoTipSetting, setComputeUnitMarginMultiplier . I think exposing TransactionConfig is probably not very descriptive, so maybe replacing with 4 getters is better?
buildTransaction.ts and sentTransaction.ts (a bit off-topic)
I noticed that the closePosition method from the Legacy SDK returns an array of transactions, in case all the instructions do not fit in a single transaction. Is this something we can add to the tx-sender?

ts-sdk/tx-sender/src/config.ts
wjthieme: Hmm. That checks out. The idea of the new RPC object is that it can do more complex things like auto-retries, round-robin across multiple urls, etc. Not sure how many people will use it like that though. Can't we use the rpc object to construct the triton prio fee call (using one of the lower level functions `createRpcMessage` from `@solana/rpc-spec-types`). IMO not a blocker so you can also add a ticket to linear for this (changing it from `string` to `string | Rpc` is non-breaking change)

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -0,0 +1,82 @@
...
+ * assertTransactionIsFullySigned(signedTransaction);
+ *
+ * const signature = await sendSignedTransaction(
+ *   signedTransaction,
+ * );
+ */
+export async function sendSignedTransaction(
+  transaction: FullySignedTransaction
+) {
+  const { rpcUrl } = getRpcConfig();
```
wjthieme, line 63: Can you add a follow-up item to add a function sendTranscations which sends a list of the txs concurrently ?

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @throws {Error} If transaction sending fails or RPC connection fails
+ *
+ * @example
+ * assertTransactionIsFullySigned(signedTransaction);
+ *
+ * const signature = await sendSignedTransaction(
+ *   signedTransaction,
+ * );
+ */
+export async function sendSignedTransaction(
```
wjthieme, line 60: nit: maybe just `sendTransaction`?

ts-sdk/tx-sender/src/computeBudget.ts
```diff
@@ -0,0 +1,193 @@
...
+    const rpc = rpcFromUrl(rpcUrl);
+    const recent = await rpc
+      .getRecentPrioritizationFees(writableAccounts)
+      .send();
+    const nonZero = recent
+      .filter((pf) => pf.prioritizationFee > 0)
+      .map((pf) => pf.prioritizationFee);
+    const sorted = nonZero.sort((a, b) => Number(a - b));
+
+    if (percentile === "50") {
```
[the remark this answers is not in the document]
parhim: depends on whether the length of nonZero is odd or even
for true median it takes the average of two middle elements

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -57,22 +61,49 @@ export async function buildAndSendTransaction(
...
+        .sendTransaction(encodedTransaction, {
+          maxRetries: BigInt(0),
+          skipPreflight: true,
+          encoding: "base64",
+        })
+        .send();
+
+      const { value } = await rpc.getSignatureStatuses([txHash]).send();
+
+      if (value[0]?.confirmationStatus === commitment) {
```
wjthieme, line 98: Might want to inspect the error here and throw if there is one. Confirmed != tx success cause it can also mean tx landed in a block but failed

ts-sdk/tx-sender/src/computeBudget.ts
```diff
@@ -0,0 +1,183 @@
...
+   * landed transaction, specified in increments of
+   * micro-lamports (0.000001 lamports).
+   */
+  prioritizationFee: MicroLamports;
+  /** Slot in which the fee was observed */
+  slot: Slot;
+};
+
+function percentileNumber(percentile: Percentile) {
+  switch (percentile) {
```
wjthieme, line 171: nit: `parseInt`?

wjthieme (approved):
Only thing we would need before we can publish is the name change but LFG

PR #673, opened 2025-01-23
chore: upgrade bytemuck to 1.19.0

154562edmck:


wjthieme:
Thanks for the contribution. The next version update we have planned for the program is when anchor 0.31 gets out. Feel free to reach out in our discord and we can help solve your specific use case

PR #680, opened 2025-01-23
Integrate changesets into monorepo

wjthieme:
Right now updating the packages in the monorepo is either update all of them at once or update none. This is not very efficient and annoying because we get lots of new versions without any changes.

This also allows us to publish the legacy-sdk packages using the same process as the new sdks.

Furthermore we don't really have proper changelogs for the libraries. This enforces PRs to contain a changeset (which will be built into a changelog and determines which packages should be published)


scripts/publish-npm
```diff
@@ -0,0 +1,29 @@
...
+
+# Patch version numbers for local dependencies
+local_deps=$(jq -r '.dependencies // {} | to_entries | .[] | select(.value == "*") | .key' package.json)
+if [ -n "$local_deps" ]; then
+  for dep in $local_deps; do
+    latest_version=$(npm view $dep version)
+    jq --arg dep "$dep" --arg version "$latest_version" \
+      '.dependencies[$dep] = $version' package.json > tmp.json && mv tmp.json package.json
+  done
+fi
```
odcheung, line 25: say you updated common-sdk, i see that this will increment legacy-sdk's version, but will it publish legacy-sdk as well?

odcheung (approved):
overall looks good. just 1 question on the dependency publish workflow (but not blocking)

scripts/publish-npm
wjthieme: Only packages that have one ore more changesets will be updated

wjthieme: Only packages that have one ore more changesets will be updated

PR #695, opened 2025-01-30
fixed filenames to be windows compatible

jg1011:
Due to inability to gain local access had to create a new folder and move stuff around, but fixed the annoying file name error one gets trying to pull on windows. 

I could only spot one incompatible filename by eye, more PRs may be due if I missed any,


yugure-orca (approved):
Thanks for the fix!

Note: The problematic character was a colon (":").

PR #726, opened 2025-02-06
Add support for solana v2 crates to rust client

wjthieme:
Since this is a breaking change, I will bump the major version of the orca_whirlpools_client crate. This version is still backwards compatible with solana v1 crates but the consumer will need to enable the `solana-v1` feature.

Theoretically I could have also done non-breaking:
* Use a solana-v2 feature that a user explicitly needs to enable. Imo this makes it more difficult down the line to make solana v2 the default/recommended
* Add solana-v1 to the default features. This would be non-breaking but would just add the breaking-change and major version bump for a later date. This would make the consumers cargo.toml less clean as having to disable the default features is less nice than having to enable a feature

### DIFF
--- a/.changeset/stupid-rats-laugh.md
+++ b/.changeset/stupid-rats-laugh.md
@@ -1,0 +1,6 @@
+---
+"@orca-so/whirlpools-rust-client": major
+"@orca-so/whirlpools-rust": major
+---
+
+Add support for solana v2 crates, to use solana v1 with orca_whirlpools_client please use the `solana-v1` feature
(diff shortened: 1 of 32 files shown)


ara-selini:
Hey folks, is there an update on this?

motlaq95 (commented):
Ok

.changeset/stupid-rats-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+"@orca-so/whirlpools-rust-client": major
+"@orca-so/whirlpools-rust": major
+---
+
+Add support for solana v2 crates, to use solana v1 with orca_whirlpools_client please use the `solana-v1` feature
```
Janabreuning67, line 6: H

Janabreuning67 (commented):
Need helps to recover all my losses Crypto

.changeset/stupid-rats-laugh.md
Janabreuning67: I need help recovering all my losses coins

Issue #727, opened 2025-02-06
🐛 Bug Report: `fetch_positions_for_owner` fails due to invalid token program parsing


pplanel:
#### **Description**  
When running the example to fetch pool positions by owner using `orca_whirlpools`, an error occurs:  

```
called `Result::unwrap()` on an `Err` value: Invalid
```

After debugging, in this call stack
1. [fetch_positions_for_owner](https://github.com/orca-so/whirlpools/blob/cde4bece3b9fd89d56269fca65ac0ef5bd56ba82/rust-sdk/whirlpool/src/position.rs#L139) calls
 1.1. [`get_token_accounts_for_owner`](https://github.com/orca-so/whirlpools/blob/cde4bece3b9fd89d56269fca65ac0ef5bd56ba82/rust-sdk/whirlpool/src/account.rs#L39) calls
1.2. [`rpc.get_token_accounts_by_owner`](https://github.com/orca-so/whirlpools/blob/cde4bece3b9fd89d56269fca65ac0ef5bd56ba82/rust-sdk/whirlpool/src/account.rs#L44)

The issue occurs while trying to parse data.program as a Pubkey in this [line](https://github.com/orca-so/whirlpools/blob/cde4bece3b9fd89d56269fca65ac0ef5bd56ba82/rust-sdk/whirlpool/src/account.rs#L52)

However, looking at the response from [`rpc.get_token_accounts_by_owner`](https://github.com/orca-so/whirlpools/blob/cde4bece3b9fd89d56269fca65ac0ef5bd56ba82/rust-sdk/whirlpool/src/account.rs#L44), the `program` field contains `"spl-token"`, which is not a valid public key:  

```rust
[src/main.rs:28:5] &accounts[0] = RpcKeyedAccount {
    pubkey: "51qZjmU4G3y19NgHzYusNxufjLRvro7QPUDJRY4TgMzP",
    account: UiAccount {
        lamports: 2039280,
        data: Json(
            ParsedAccount {
                program: "spl-token",
                parsed: Object {
                    "info": Object {
                        "isNative": Bool(false),
                        "mint": String("Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB"),
                        "owner": String("HhZ7ofbr1JgrM6uEJXgWs8VzjHmHL3Tm7ke45ntLFkZs"),
                        "state": String("initialized"),
                        "tokenAmount": Object {
                            "amount": String("0"),
                            "decimals": Number(6),
[...]

PR #729, opened 2025-02-07
fix: modify parsing logic in get_token_accounts_for_owner

pplanel:
Fixes https://github.com/orca-so/whirlpools/issues/727

Modify the parsing logic to check if data.program is "spl-token" and use spl_token::ID, rather than attempting to parse it as a Pubkey.


wjthieme:
Thanks for the contribution! Can you add a patch changeset for this fix?

wjthieme (approved):

PR #735, opened 2025-02-10
fix: allow whirlpool_repositioning_bot work on pools paired with SOL

pplanel:
## Description
This change extends the `fetch_token_balance` function to support fetching the balance of the native SOL mint in addition to existing SPL token balance retrieval.

## Changes
- Added a special case for the native SOL mint address
- Uses `get_balance()` RPC method to fetch SOL balance
- Converts lamports to SOL using `lamports_to_sol()` function
- Adds a constant `NATIVE_MINT_ADDRESS` for clarity

## Rationale
Previously, the utility function only supported fetching balances for SPL tokens. This change allows fetching the SOL balance using the same method, improving the utility's flexibility.

### DIFF
--- a/.changeset/mean-cows-flow.md
+++ b/.changeset/mean-cows-flow.md
@@ -1,0 +1,5 @@
+---
+"@orca-so/whirlpools-example-rust-repositioning-bot": patch
+---
+
+Update to work on SOL paired pools
--- a/examples/rust-sdk/whirlpool_repositioning_bot/Cargo.toml
+++ b/examples/rust-sdk/whirlpool_repositioning_bot/Cargo.toml
@@ -13,7 +13,8 @@ serde_json = { version = "^1.0" }
 solana-client = { version = "^1.18" }
 solana-sdk = { version = "^1.18" }
 spl-token-2022 = { version = "^3.0" }
+spl-token = { version = "^3.0" }
 spl-associated-token-account = { version = "^3.0" }
 tokio = { version = "^1.41.1" }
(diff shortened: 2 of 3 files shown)


calintje:
There are quite a few cases where LPers/traders already have a wSOL token account which they use to perform DeFi actions. This reduces the need to create and close token accounts with each transaction, which reduces transaction size and compute units used.

With this fix, they would see their SOL balance, and not their wSOL balance.

Do you think we can come up with a clean solution for both scenarios? Otherwise I propose to clarify this issue in the README, and let users build their own implementation on top of this.

pplanel:
You're absolutely right—the initial implementation was a bit naive.

In my case, I have a SOL/USDC LP position, which is actually WSOL/USDC under the hood. However, I don't have an ATA for WSOL. The bot was checking the balance of the pool’s mint, and since I didn’t have a WSOL account, the check failed.

I've updated the PR to handle this scenario more gracefully. Now, it correctly differentiates between SOL and WSOL cases, ensuring that users see the expected balance based on their token accounts. Let me know if you have any further suggestions!

calintje:
Thank you!

Could you please share how you tested this, including transaction hashes? From my experience, in the current state of the example, the transaction size exceeds the transaction size limit when the SDK adds the instructions to handle token account creation.

pplanel:
Yeah, I'm getting the same error. A Address Lookup Table would help here?

calintje:
Let me know if you find a better solution for this that works for all cases and we'll reopen the PR!

PR #743, opened 2025-02-13
Increase MAX_FEE_RATE

yugure-orca:
## Motivation
To support more high fee rate for volatile token pairs.

Current max: 30_000 (3%)
Next max: 60_000 (6%)

### DIFF
--- a/legacy-sdk/whirlpool/tests/integration/initialize_fee_tier.test.ts
+++ b/legacy-sdk/whirlpool/tests/integration/initialize_fee_tier.test.ts
@@ -87,7 +87,7 @@ describe("initialize_fee_tier", () => {
       configInitInfo,
       configKeypairs.feeAuthorityKeypair,
       testTickSpacing,
-      30_000, // 3 %
+      60_000, // 6 %
     );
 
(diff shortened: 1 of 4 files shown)


.changeset/healthy-chicken-tease.md
```diff
@@ -0,0 +1,2 @@
+---
```
philcchen, line 1: sorry, what is this?

philcchen (approved):

.changeset/healthy-chicken-tease.md
yugure-orca: @philcchen 
Thanks for the review.

Changeset have been introduced recently.
It is used to publish SDK packages with auto-increment on semver.
https://github.com/orca-so/whirlpools?tab=readme-ov-file#changesets

This change doesn't affect client-side code, so I executed `yarn changeset --empty` and this file is generated. 🙂

wjthieme:
Nit: changelogs are also created for the program so an actual changeset would be nice!

wjthieme (approved):

PR #768, opened 2025-02-22
LL

yugure-orca:
For LL release


.changeset/purple-buses-move.md
```diff
@@ -0,0 +1,6 @@
+---
+"@orca-so/whirlpools-sdk": patch
+"@orca-so/whirlpools-program": patch
```
wjthieme, line 3: nit: Can you also bump `@orca-so/whirlpools-client` and `@orca-so/whirlpools-client-rust`. That way the clients will be regenerated from the new IDL and published as well (without the gpa and pda helpers but we can add those later)

wjthieme (approved):
LGTM

.changeset/purple-buses-move.md
yugure-orca: Thanks for the comment!
added!

philcchen (approved):
Audit already conducted, will be added in a follow-up

PR #778, opened 2025-02-27
Event emission

yugure-orca:
For event emission release

note: PR59

### DIFF
--- a/programs/whirlpool/src/instructions/v2/two_hop_swap.rs
+++ b/programs/whirlpool/src/instructions/v2/two_hop_swap.rs
@@ -383,74 +383,36 @@ pub fn handler<'info>(
     */
 
     let pre_sqrt_price_one = whirlpool_one.sqrt_price;
-    let (input_amount_one, input_transfer_fee_one, output_amount_one, output_transfer_fee_one) =
-        if a_to_b_one {
-            (
-                swap_update_one.amount_a,
-                calculate_transfer_fee_excluded_amount(
-                    &ctx.accounts.token_mint_input,
-                    swap_update_one.amount_a,
-                )?
-                .transfer_fee,
-                swap_update_one.amount_b,
-                calculate_transfer_fee_excluded_amount(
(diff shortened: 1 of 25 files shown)


wjthieme (approved):
🚀🚀🚀

programs/whirlpool/src/instructions/v2/two_hop_swap.rs
```diff
@@ -381,6 +382,78 @@ pub fn handler<'info>(
     )
     */
 
+    let pre_sqrt_price_one = whirlpool_one.sqrt_price;
+    let (input_amount_one, input_transfer_fee_one, output_amount_one, output_transfer_fee_one) =
+        if a_to_b_one {
```
philcchen, line 387: nit: I think to reduce branching and enhance readability, I generally like to split the variables prior to function calls, so in this case it might be something like

```
let (input_amount_one, output_amount_one) = if a_to_b_one {
   (swap_update_one.amount_a, swap_update_one.amount_b)
} else {
   (swap_update_one.amount_b, swap_update_one.amount_a)
};
let input_transfer_fee_one = calculate_transfer_fee_excluded_amount(&ctx.accounts.token_mint_input, input_amount_one);
let output_transfer_fee_one = calculate_transfer_fee_excluded_amount(&ctx.accounts.token_mint_intermediate, output_amount_one);
```

philcchen: similar https://github.com/orca-so/whirlpools/pull/778/files#diff-6ca78fefc01c44e71a547895b03ec4c02e6cf89c8aa1ae1a7067cdb3eac7254eR149 and https://github.com/orca-so/whirlpools/pull/778/files#diff-425cb0a6e7a9a9c5ca156a41b0d6e9d0f8a7e75e204ef9cda0083fa035c3244eR423 as well

yugure-orca: Thanks! updated. It is much cleaner 😂

PR #795, opened 2025-03-03
sam/DX-251

parhim:
This serves as a wrapper over @orca/whirlpools sdk that enables building and sending transactions with the appropriate fee settings

every function, except **harvestAllPositionFees** will return an object containing a callback and either a quote or initialization cost (so that it can be checked before sending the transaction)

`await callback()` will build and send the transaction

https://linear.app/orca-so/issue/DX-251/orca-actions-ts-package


ts-sdk/actions/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@orca-so/actions",
```
wjthieme, line 2: Can we do @orca-so/whirlpools-actions?

ts-sdk/actions/src/helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+  currentInstructions: IInstruction[],
+  instructionsToAdd: IInstruction[]
+): boolean {
+  // Current Solana transaction size limit is 1232 bytes
+  const MAX_TRANSACTION_SIZE = 1232;
+  const INSTRUCTION_OVERHEAD = 40; // Each instruction has ~40 bytes overhead
+
+  // Calculate total size in one pass
+  const totalSize = [...currentInstructions, ...instructionsToAdd].reduce(
+    (sum, ix) => sum + INSTRUCTION_OVERHEAD + (ix.data?.length ?? 0),
```
wjthieme, line 19: Not sure this calculation is correct. Looks like this doesn't take instruction account addresses

ts-sdk/actions/src/positions.ts
```diff
@@ -0,0 +1,171 @@
...
+import { getPayer } from "./config";
+import {
+  CollectFeesQuote,
+  CollectRewardsQuote,
+  DecreaseLiquidityQuote,
+  IncreaseLiquidityQuote,
+} from "@orca-so/whirlpools-core";
+
+// Open a concentrated liquidity position
+export async function openConcentratedPosition(
```
wjthieme, line 25: Do you think it would be possible to create s builder function for this that takes as arguments and argument 'the function to be wrapped'. Would save some boilerplate

ts-sdk/actions/src/swap.ts
```diff
@@ -0,0 +1,73 @@
...
+): Promise<{
+  quote: ExactInSwapQuote | ExactOutSwapQuote;
+  callback: () => Promise<string>;
+}> {
+  const { rpcUrl } = getRpcConfig();
+  const rpc = rpcFromUrl(rpcUrl);
+  const owner = getPayer();
+
+  try {
+    const pools = await fetchWhirlpoolsByTokenPair(rpc, inputMint, outputMint);
```
wjthieme, line 30: Here I would just wrap the function from high-level sdk. The sdk is not meant to be a router so just supplying the pool you want to swap through is fine

ts-sdk/actions/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./pools";
```
wjthieme, line 1: Would it make sense to follow the same file structure as high-level sdk?

We might even want to reexport some of the 'fetch' functions here but I think we have a separate project for that

programs/whirlpool/src/math/u256_math.rs
```diff
@@ -92,8 +92,8 @@ impl U256Muldiv {
         }
 
         for i in (1..NUM_WORDS).rev() {
-            result.items[i] = result.items[i] << shift_amount
```
wjthieme, line 95: Can you undo changes in this file? If we change program code then our program will no longer be a verified build (unless we redeploy)

ts-sdk/actions/src/helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+  // Current Solana transaction size limit is 1232 bytes
+  const MAX_TRANSACTION_SIZE = 1232;
+  const INSTRUCTION_HEADER_SIZE = 3; // 1 byte for accounts length, 1 byte for data length, 1 byte for program id index
+  const ACCOUNT_SIZE = 33; // 32 bytes for public key, 1 byte for is_signer/is_writable flags
+
+  const totalSize = [...currentInstructions, ...instructionsToAdd].reduce(
+    (sum, ix) => {
+      let ixSize = INSTRUCTION_HEADER_SIZE;
+
+      const numAccounts = (ix.accounts?.length ?? 0) + 1; // +1 for programAddress
```
wjthieme, line 54: This does not take into account that versionedTransactions compile (dedup) the accounts. I don't know the exact logic so might be better to just try to compile it and see if it fails/succeeds? (or if it always succeeds you can check the size of the serialized message)

ts-sdk/actions/src/helpers.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @param params Parameters for the instruction function (excluding rpc and owner)
+ * @returns An object with callback and other relevant data
+ */
+export async function executeWhirlpoolInstruction<T extends unknown[], R>(
+  instructionFn: (
+    rpc: Rpc<SolanaRpcApi>,
+    ...params: [...T, TransactionSigner]
+  ) => Promise<R & { instructions: IInstruction[] }>,
+  ...params: T
+): Promise<R & { callback: () => Promise<string> }> {
```
wjthieme, line 22: Can this be `callback: () => Promise<Signature>` instead of `callback: () => Promise<string>`?

ts-sdk/actions/src/increaseLiquidity.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  openPositionInstructions,
+  openFullRangePositionInstructions,
+  increaseLiquidityInstructions,
+} from "@orca-so/whirlpools";
+import type { IncreaseLiquidityQuote } from "@orca-so/whirlpools-core";
+import { executeWhirlpoolInstruction } from "./helpers";
+
+// Open a concentrated liquidity position
+export async function openConcentratedPosition(
```
wjthieme, line 12: This you might even be able to make shorter so you don't have to make the parameters and return type explicit.

```ts
function wrapFunctionWithExecution<P, R extends { instructions: IInstruction[] }>(
  function: (...P) => R
): Promise<Omit<R, "instructions> & { callback: () => Promise<Signature> } {
  ...
}

export const openConcentratedPostion = wrapFunctionWithExecution(openPositionInstructions);
```

ts-sdk/actions/src/harvest.ts
```diff
@@ -0,0 +1,55 @@
...
+} from "@orca-so/whirlpools";
+import type { Address, IInstruction } from "@solana/kit";
+import { getPayer } from "./config";
+import {
+  executeWhirlpoolInstruction,
+  wouldExceedTransactionSize,
+} from "./helpers";
+
+// Harvest fees from all positions owned by an address
+export async function harvestAllPositionFees(): Promise<string[]> {
```
wjthieme, line 18: Should we add this to `@orca-so/whirlpools` instead and then here just have the wrapper again? We can just have the `harvestAllInstructions` function return all the instructions (regardless of whether they fit in a tx or not). Then when building the TX we can just split it up there and fire the in parallel (instead of sequentially)

ts-sdk/whirlpool/src/config.ts
```diff
@@ -170,3 +186,19 @@ export function resetConfiguration() {
   SLIPPAGE_TOLERANCE_BPS = DEFAULT_SLIPPAGE_TOLERANCE_BPS;
   NATIVE_MINT_WRAPPING_STRATEGY = DEFAULT_NATIVE_MINT_WRAPPING_STRATEGY;
 }
+
+let _payer: KeyPairSigner | undefined;
+
+export async function setPayerFromBytes(pkBytes: Uint8Array<ArrayBuffer>) {
```
wjthieme, line 192: Maybe we should add some typedoc here. This is only required for sdk-actions, right?

wjthieme (approved):
Looks super sleek!

PR #805, opened 2025-03-05
Fix fetch_positions_for_owner() to perform batch calls

pauldragonfly:


wjthieme (commented):
Can you also add this for the ts sdk?

.changeset/easy-maps-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@orca-so/whirlpools-rust": patch
```
wjthieme, line 2: Also need a changeset for the TS sdk

rust-sdk/whirlpool/src/utils.rs
```diff
@@ -0,0 +1,20 @@
...
+const DEFAULT_CHUNK_SIZE: usize = 100;
+
+pub(crate) async fn batch_get_multiple_accounts(
+    rpc_client: &RpcClient,
+    pubkeys: &[Pubkey],
+    chunk_size: Option<usize>,
+) -> Result<Vec<Option<Account>>, Box<dyn std::error::Error>> {
+    let mut results = vec![];
+
+    for chunk in pubkeys.chunks(chunk_size.unwrap_or(DEFAULT_CHUNK_SIZE)) {
```
wjthieme, line 14: nit: Clamp the chunk_size to (0, MAX_CHUNK_SIZE)?

ts-sdk/whirlpool/src/position.ts
```diff
@@ -130,22 +130,44 @@ export async function fetchPositionsForOwner(
...
 
-  // FIXME: need to batch if more than 100 position bundles?
-  const [positions, positionBundles] = await Promise.all([
-    fetchAllMaybePosition(rpc, positionAddresses),
-    fetchAllMaybePositionBundle(rpc, positionBundleAddresses),
-  ]);
+  const chunkSize = 100;
+
+  let promises = [];
+  for (let i = 0; i < positionAddresses.length; i += chunkSize) {
```
wjthieme, line 136: Would be nice if you can make this a separate function as well (like in rust). Something like:

```
export async function fetchMultipleAccountsBatched<T extends object>(
  rpc: Rpc<GetMultipleAccountsApi>,
  addresses: Address[],
  decoder: (account: MaybeEncodedAccount) => MaybeAccount<T>,
): Promise<MaybeAccount<T>[]> {
  const numChunks = Math.ceil(addresses.length / MAX_CHUNK_SIZE);
  const chunks = [...Array(numChunks).keys()].map((i) =>
    addresses.slice(i * MAX_CHUNK_SIZE, (i + 1) * MAX_CHUNK_SIZE),
  );

  const results: MaybeAccount<T>[] = [];
  for (const chunk of chunks) {
    const chunkResult = await fetchEncodedAccounts(rpc, chunk);
    chunkResult.forEach((account, i) => {
      const data = decoder(account);
      results.push(data);
    });
  }
  return results;
}
```

rust-sdk/whirlpool/src/utils.rs
pauldragonfly: > An iterator over a slice in (non-overlapping) chunks (chunk_size elements at a time), starting at the beginning of the slice.
> When the slice len is not evenly divided by the chunk size, the last slice of the iteration will be the remainder.

Clamp is necessary here?

.changeset/easy-maps-repair.md
pauldragonfly: Fixed.

ts-sdk/whirlpool/src/position.ts
pauldragonfly: Fixed.

rust-sdk/whirlpool/src/utils.rs
wjthieme: Anything less than 1 and more than 100 is invalid

pauldragonfly: Fixed.

PR #820, opened 2025-03-12
DX-272: Rust tx-sender crate

parhim:
https://linear.app/orca-so/issue/DX-272/build-and-test-crate-with-same-specs-as-ts-package

this still needs some thoughtful review but at least it works with and without fees
jito
exact + dynamic

not as clean as typescript but thats a rust limitation
tests are doing small transfers between orca wallets (can use whatever on devnet)
but you need a locally stored `mainnet-test-keypair.json` (ping me on slack if you need the lightly funded orca pk)


rust-sdk/tx-sender/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tokio::time::sleep;
+
+/// Default compute units for transactions
+pub const DEFAULT_COMPUTE_UNITS: u32 = 200_000;
+
+/// Transaction sender for building and sending Solana transactions
+#[derive(Clone)]
+pub struct TransactionSender {
```
wjthieme, line 36: What is the reason for not doing the same setup as on TS? (global default config and just only top-level functions)

rust-sdk/tx-sender/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                .await
+            {
+                Ok(signature) => return Ok(signature),
+                Err(err) => {
+                    // Check if we should retry
+                    if retries >= self.tx_config.max_retries {
+                        return Err(TransactionError::RpcError(err));
+                    }
+
+                    // Exponential backoff
```
wjthieme, line 155: Is this logic different then from the TS version? If so would it make sense to keep them the same as much as possible?

rust-sdk/tx-sender/src/rpc_config.rs
```diff
@@ -0,0 +1,130 @@
+use crate::error::{Result, TransactionError};
+use serde::{Deserialize, Serialize};
+use solana_client::nonblocking::rpc_client::RpcClient;
+use solana_sdk::hash::Hash;
+use std::str::FromStr;
+use std::time::Duration;
+
+/// Chain identifier based on genesis hash
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChainId(Hash);
```
wjthieme, line 10: Would it make sense to have a `enum` instead and implement `From<Hash>` for it?

rust-sdk/tx-sender/src/rpc_config.rs
```diff
@@ -0,0 +1,130 @@
...
+        } else if self.is_eclipse_testnet() {
+            "eclipse-testnet"
+        } else {
+            "unknown"
+        }
+    }
+}
+
+/// RPC configuration for connecting to Solana nodes
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
wjthieme, line 68: Does this need Serialize and Deserialize?

rust-sdk/tx-sender/src/rpc_config.rs
```diff
@@ -0,0 +1,130 @@
...
+        Self {
+            url: url.into(),
+            supports_priority_fee_percentile: false,
+            chain_id: None,
+            timeout: default_timeout(),
+        }
+    }
+
+    /// Async constructor with chain ID detection
+    pub async fn with_chain_detection(url: impl Into<String>) -> Result<Self> {
```
wjthieme, line 103: I would just only do the async constructor. Is there a scenario where we would want this without a chain_id?

rust-sdk/tx-sender/src/tests.rs
```diff
@@ -0,0 +1,70 @@
+#[cfg(test)]
+mod tests {
```
wjthieme, line 2: Why is this a separate file and not in-file?

rust-sdk/tx-sender/src/tx_config.rs
```diff
@@ -0,0 +1,56 @@
+use serde::{Deserialize, Serialize};
+use solana_client::rpc_config::RpcSendTransactionConfig;
+use solana_sdk::commitment_config::CommitmentLevel;
+use std::time::Duration;
+
+/// Transaction configuration for sending transactions
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
wjthieme, line 7: Does this need Serialize and Deserialize?

rust-sdk/tx-sender/src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+use solana_client::nonblocking::rpc_client::RpcClient;
+
+use crate::fee_config::{FeeConfig, PriorityFeeStrategy, JitoFeeStrategy};
+use crate::rpc_config::RpcConfig;
+
+/// Default transaction timeout in milliseconds
+pub const DEFAULT_TRANSACTION_TIMEOUT_MS: u64 = 30_000;
+
+/// Default number of retries
+pub const DEFAULT_RETRIES: usize = 3;
```
wjthieme, line 29: Are these still being used?

rust-sdk/tx-sender/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+    
+    let sim_result = rpc_client.simulate_transaction(&transaction)
+        .await
+        .map_err(|e| format!("Transaction simulation failed: {}", e))?;
+
+    if let Some(err) = sim_result.value.err {
+        return Err(format!("Transaction simulation failed: {}", err));
+    }
+
+    let expiry_time = Instant::now() + Duration::from_millis(options.timeout_ms);
```
wjthieme, line 143: Think the default for this should be 90 seconds instead of 30 (or just make it not configurable like on TS). The issue with lowering the timeout is that the transaction is valid for the full 90 seconds. This means that if it ends up landing on the final send this function never picks up on it. If you set the timeout to 30 seconds and it 'times-out' you can't definitively say that the transaction did not land

wjthieme (approved):
One comment about timeout and retries but other than that lgtm 🚀

PR #821, opened 2025-03-13
Fix position deviation calculation by using price instead of sqrt price

calintje:
### Fix calculation position deviation
Previously, the postion center price was calculated using `sqrt_price`, which resulted in an inaccurate deviation measurement. This PR updates the calculation to use actual prices instead of sqrt prices.

Before - incorrect calculation:
( sqrt(110) + sqrt(90) ) / 2 != sqrt(100)

After - correct calculation:
(110 + 90) / 2 == 100

### Update README
Add a note to the README to emphasize that his implementation works with SPL tokens only.


examples/rust-sdk/whirlpool_repositioning_bot/README.md
```diff
@@ -1,50 +1,68 @@
...
+
 1. Clone this repository:
-  ```bash
-  git clone https://github.com/orca-so/whirlpools.git
-  cd examples/rust-sdk/whirlpool_repositioning_bot
-  ```
+
+```bash
+git clone https://github.com/orca-so/whirlpools.git
+cd examples/rust-sdk/whirlpool_repositioning_bot
```
yugure-orca, line 40: We've not received any response from this dev, but I guess it is nice to run `yarn run build` at the root.

Our "working" directory has generated crate, so this procedure works.
But what happens if we clone the repository at a new place.

https://discord.com/channels/798712664590254081/1049776001367949413/1349564173893374014

yugure-orca (approved):
LGTM for code part.

Recently, one dev struggled at `cargo build --release`, so I think we can improve setup process.

examples/rust-sdk/whirlpool_repositioning_bot/README.md
calintje: Building the entire repo with `yarn build` from root can maybe be a bit overkill? For some users, that also causes issues if they're not using the correct version of Rust. Also, if devs want to continue building from this reference example, they will most likely use it in a new repo.

A few lines below this comment in the README there is a note to change the reference to the dependency from local to the versions.

```
2. Update `Cargo.toml`
   This project uses the local version of the dependencies. If you want to move this example project outside of this repo, update the `Cargo.toml`:

---
# other dependencies
orca_whirlpools = { version = "^2.0" }
orca_whirlpools_client = { version = "^2.0" }
orca_whirlpools_core = { version = "^2.0" }
---
```

Maybe it's more user friendly to have devs update the `cargo.toml`?

yugure-orca: > that also causes issues if they're not using the correct version of Rust. 

good point.

Before even considering running it outside of this repository, simply cloning the repository and running cargo build results in an error because generated does not exist.

Therefore, it would be helpful to either include instructions to modify the dependencies or provide an option to switch dependencies within Cargo.toml. (If the latter is a bit difficult, instructing users to modify Cargo.toml manually could be a viable approach.)

In any case, users who want to work with the SDK expect it to run smoothly, so reducing potential stumbling points would be beneficial.

calintje: The suggested option to switch dependencies within Cargo.toml is nice. However, I was not able to get that to work without errors.

For now, I will leave it like this:

- I improved the README, explaining that there are two ways to build and use this project:
  - build from root
  - build the project only

- I added comments in the `Cargo.toml`.

yugure-orca (approved):

wjthieme (approved):

PR #853, opened 2025-03-24
Remove specified rust version from anchor action

calintje:
The Publish workflow seems to fail on the build step. One identified difference compared to the Checks and PR workflows is that the Publish workflow uses a specified version of Rust, while the others do not.

### DIFF
--- a/.github/actions/anchor/action.yml
+++ b/.github/actions/anchor/action.yml
@@ -1,5 +1,5 @@
-name: 'Setup Anchor'
-description: 'Sets up an Anchor development environment.'
+name: "Setup Anchor"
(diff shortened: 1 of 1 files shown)


.github/actions/anchor/action.yml
```diff
@@ -1,59 +1,37 @@
...
   anchor-version:
-    description: 'The version of anchor to use.'
-    default: 'v0.29.0'
+    description: "The version of anchor to use."
+    default: "v0.29.0"
 
 runs:
   using: "composite"
   steps:
-    - name: Setup rustc
```
wjthieme, line 32: We still need to setup rust. We just need to set it up with the same version as the build workflow

PR #884, opened 2025-04-08
Prevent endless swap loop when token amount exceeds available liquidity

calintje:
This PR fixes a bug where the swap loop in the core SDK could run endlessly when the token amount exceeds available liquidity. The issue occurs when the `sqrt_price_limit` is set to 0 (which maps to `MIN_SQRT_PRICE` or `MAX_SQRT_PRICE` depending on `a_to_b`), causing the loop condition to never break.

**Changes**:
- Added INVALID_TICK_ARRAY_SEQUENCE error constant
- Modified the `next_initialized_tick` and `prev_initialized_tick` functions to throw an error when traversing beyond the tick array bounds, rather than returning None
- Added tests to verify this behavior

This PR aligns the SDK's behavior with both the program and Legacy SDK, which typically stop the swap loop when the next tick is not part of the tick array sequence.


rust-sdk/core/src/math/tick_array.rs
```diff
@@ -258,9 +257,8 @@ mod tests {
     #[test]
     fn test_get_next_initializable_tick_out_of_bounds() {
         let sequence = test_sequence(16);
-        let pair = sequence.next_initialized_tick(2817);
```
yugure-orca, line 261:
```
test sequence: [-1408      0) [0         1408) [1408          2816)
               start = -1408                                    end=2815
```
Then
- `sequence.next_initialized_tick(2813) -> Ok((None, 2815))`
- `sequence.next_initialized_tick(2814) -> Ok((None, 2815))`
- `sequence.next_initialized_tick(2815) -> Err(INVALID_TICK_ARRAY_SEQUENCE )` # next is NOT inclusive
- `sequence.next_initialized_tick(2816) -> Err(INVALID_TICK_ARRAY_SEQUENCE )`
- `sequence.next_initialized_tick(2817) -> Err(INVALID_TICK_ARRAY_SEQUENCE )`

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -298,8 +296,7 @@ mod tests {
     #[test]
     fn test_get_prev_initialized_tick_out_of_bounds() {
         let sequence = test_sequence(16);
-        let pair = sequence.prev_initialized_tick(-1409);
-        assert_eq!(pair.map(|x| x.1), Ok(-1408));
-        assert_eq!(pair.map(|x| x.0), Ok(None));
```
yugure-orca, line 303:
```
test sequence: [-1408      0) [0         1408) [1408          2816)
               start = -1408                                    end=2815
```
Then
- `sequence.prev_initialized_tick(-1407) -> Ok((None, -1408))`
- `sequence.prev_initialized_tick(-1408) -> Ok((None, -1408))` # prev is inclusive
- `sequence.prev_initialized_tick(-1409) -> Err(INVALID_TICK_ARRAY_SEQUENCE )`
- `sequence.prev_initialized_tick(-1410) -> Err(INVALID_TICK_ARRAY_SEQUENCE )`

yugure-orca (changes requested):
Thanks for the fix.
Could you check concentrated pool with full range position case ?

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -86,9 +87,8 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
         let mut next_index = tick_index;
         loop {
             next_index = get_next_initializable_tick_index(next_index, self.tick_spacing);
-            // If at the end of the sequence, we don't have tick info but can still return the next tick index
             if next_index > array_end_index {
-                return Ok((None, array_end_index));
+                return Err(INVALID_TICK_ARRAY_SEQUENCE);
             }
             let tick = self.tick(next_index)?;
             if tick.initialized {
```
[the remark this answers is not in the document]
yugure-orca: note: if `next_initialized_tick` returns `x`, it will receive `x` next time. (so `if tick_index >= array_end_index` should work).

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -105,9 +105,8 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
         let mut prev_index =
             get_initializable_tick_index(tick_index, self.tick_spacing, Some(false));
         loop {
-            // If at the start of the sequence, we don't have tick info but can still return the previous tick index
             if prev_index < array_start_index {
-                return Ok((None, array_start_index));
+                return Err(INVALID_TICK_ARRAY_SEQUENCE);
             }
             let tick = self.tick(prev_index)?;
             if tick.initialized {
```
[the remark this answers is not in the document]
yugure-orca: note: if `prev_initialized_tick` returns `x`, it will receive `x - 1` next time. (so `if tick_index < array_start_index` should work).

rust-sdk/core/src/math/tick_array.rs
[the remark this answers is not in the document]
calintje: I see. It took me a bit to notice the exact issue: not returning the index at the bound of the tick array sequence would result in unprocessed liquidity between the 'last initialized tick' and the bound. In the specific case you described above, the swap would just return an error even though there was sufficient liquidity between the `current_tick` and one of the bounds.

calintje:
Thank you for the review!

It took me a bit to notice the exact issue: not returning the index at the bound of the tick array sequence would result in unprocessed liquidity between the 'last initialized tick' and the bound. In the specific example you describe, the swap would just return an error even though there might have been sufficient liquidity between the current tick and one of the bounds.

.gitignore
```diff
@@ -29,3 +29,5 @@ generated
 .env
 wallet.json
 solana-release
+rust-sdk/integration/**/Cargo.lock
```
wjthieme, line 32: Maybe we should commit these ones too (since then we don't really require lockfile patching that the script currently doesn't) but I guess that is a question for another PR/issue

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -83,10 +84,12 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
         tick_index: i32,
     ) -> Result<(Option<&TickFacade>, i32), CoreError> {
```
wjthieme, line 85: Does it still make sense to return `Rusult<Option<...` here? I think the option part is always `Some`

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -83,10 +84,12 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
...
         let array_end_index = self.end_index();
+        if tick_index >= array_end_index {
+            return Err(INVALID_TICK_ARRAY_SEQUENCE);
+        }
         let mut next_index = tick_index;
         loop {
             next_index = get_next_initializable_tick_index(next_index, self.tick_spacing);
-            // If at the end of the sequence, we don't have tick info but can still return the next tick index
             if next_index > array_end_index {
```
wjthieme, line 93: I think this logic differs slightly from the contract. In the contract we do the following checks which I think we should try to match as closely as possible

* if the next tick index is outside of the squence return `INVALID_TICK_ARRAY_SEQUENCE`
  * if there is no next index, we are at the at MIN_TICK_INDEX, MAX_TICK_INDEX (<- think this one might still be missing)
  * if there is an index check if it is initialized or not

rust-sdk/core/src/math/tick_array.rs
calintje: I think the option part returns `None` if the tick at the bound of the tick array sequence is uninitialized. Then, it returns None for the tick, but it still returns the tick index at the bound, which the `while` loop in `compute_swap` uses to derive `next_tick_sqrt_price`, which it then uses for the calculation of the remaining liquidity.

rust-sdk/core/src/math/tick_array.rs
yugure-orca: hmm, it may be nice to test "ts=1" case.

if `MIN_TICK_INDEX` or `MAX_TICK_INDEX` is initializable (multiple of ts), these will be the return value of `get_next_initializable_tick_index` (because it is virtually a generator of numbers which are multiple of tick spacing).
And `next_index > array_end_index` doesn't match `MAX_TICK_INDEX`, so `MAX_TICK_INDEX` will be returned with `Some(tick)` if they are initialized.

But this is a simulation in my mind, so "test" is nice.

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -102,10 +105,12 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
         tick_index: i32,
     ) -> Result<(Option<&TickFacade>, i32), CoreError> {
         let array_start_index = self.start_index();
+        if tick_index < array_start_index {
+            return Err(INVALID_TICK_ARRAY_SEQUENCE);
+        }
         let mut prev_index =
             get_initializable_tick_index(tick_index, self.tick_spacing, Some(false));
         loop {
-            // If at the start of the sequence, we don't have tick info but can still return the previous tick index
```
yugure-orca, line 108: nit: I think no need to remove this comment.

rust-sdk/core/src/math/tick_array.rs
```diff
@@ -83,10 +84,12 @@ impl<const SIZE: usize> TickArraySequence<SIZE> {
         tick_index: i32,
     ) -> Result<(Option<&TickFacade>, i32), CoreError> {
         let array_end_index = self.end_index();
+        if tick_index >= array_end_index {
+            return Err(INVALID_TICK_ARRAY_SEQUENCE);
+        }
         let mut next_index = tick_index;
         loop {
             next_index = get_next_initializable_tick_index(next_index, self.tick_spacing);
-            // If at the end of the sequence, we don't have tick info but can still return the next tick index
```
yugure-orca, line 89: nit: I think no need to remove this comment.

yugure-orca (commented):
Thanks for the update!

The updated code looks good to me, but could you add a bit more test cases for `prev_initialized_tick` and `next_initialized_tick` ?

- prev_initialized_tick returns the first tick of the last tick array (-1408) and it is NOT initialized.
- prev_initialized_tick returns the first tick of the last tick array (-1408) and it is initialized.
- prev_initialized_tick returns MIN_TICK_INDEX and it is NOT initialized (ts=64, MIN is not multiple of ts).
- prev_initialized_tick returns MIN_TICK_INDEX and it is NOT initialized (ts=1, MIN is multiple of ts).
- prev_initialized_tick returns MIN_TICK_INDEX and it is initialized (ts=1, MIN is multiple of ts).

(similar to next_initialized_tick)

rust-sdk/core/src/math/tick_array.rs
calintje: I believe the second step is already covered in the implementation of `start_index()` and `end_index()`:
https://github.com/orca-so/whirlpools/blob/1fc3818ed2349f91bc456565100847e24b0756b2/rust-sdk/core/src/math/tick_array.rs#L46-L62

yugure-orca (approved):
LGTM!

PR #912, opened 2025-04-21
Expose useful utility functions in Rust and TypeScript SDKs

calintje:
This PR exposes useful utility functions in both Rust and TypeScript SDKs:
- Exported `try_get_token_estimates_from_liquidity`. This is more straightforward then pointing users to use 'decreaseLiquidityQuote` functions to estimate their token balances.
  - Add new interface `PositionTokenAmounts`
  - Add tests
- Exports the `orderMints` utility function from the TypeScript SDK, similar to how the Legacy SDK export `PoolUtil.ordermints`

### DIFF
--- a/rust-sdk/core/src/quote/liquidity.rs
+++ b/rust-sdk/core/src/quote/liquidity.rs
@@ -7,7 +7,8 @@ use crate::{
     order_tick_indexes, position_status, tick_index_to_sqrt_price, try_apply_transfer_fee,
     try_get_max_amount_with_slippage_tolerance, try_get_min_amount_with_slippage_tolerance,
     try_reverse_apply_transfer_fee, CoreError, DecreaseLiquidityQuote, IncreaseLiquidityQuote,
-    PositionStatus, TransferFee, AMOUNT_EXCEEDS_MAX_U64, ARITHMETIC_OVERFLOW, U128,
+    PositionStatus, PositionTokenAmounts, TransferFee, AMOUNT_EXCEEDS_MAX_U64, ARITHMETIC_OVERFLOW,
+    U128,
 };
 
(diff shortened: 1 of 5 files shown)


rust-sdk/core/src/quote/liquidity.rs
```diff
@@ -370,62 +371,41 @@ pub fn increase_liquidity_quote_b(
...
+///
+/// # Returns
+/// - A PositionTokenAmounts struct containing the token amounts in the position
+#[cfg_attr(feature = "wasm", wasm_expose)]
+pub fn get_position_token_amounts(
+    position_liquidity: U128,
+    current_sqrt_price: U128,
+    tick_index_1: i32,
+    tick_index_2: i32,
+    transfer_fee_a: Option<TransferFee>,
```
wjthieme, line 392: This is more of a product question but we might be able to do without the transfer fee in this case:

Is this:
* The amount of tokens in the position (= without transfer fee)
* The amount you can withdraw from this position (= with transfer fee)

calintje: And what about if it's a fictional position for which you want to know the token amounts?

calintje: In that case maybe exporting `try_get_token_estimates_from_liquidity` is exactly what we want. In a future pr we add a new function fetchPositionDetails, which fetches the position and gives you all the details: withdrawable, exact amounts, etc.

yugure-orca (approved):
Developers often just want to know the amount of tokens held inside a position — that’s a very common use case.
So I don’t think we need to take TransferFee into account in that context.

IMO, it is sufficient that TransferFee is considered only at the point when the tokens are actually withdrawn.

PR #915, opened 2025-04-23
Add Adaptive Fees math explainer + doc fixes

calintje:
This PR:
- Adds an Adaptive Fees Math explainer to the `Whirlpool Fees` doc
- Fixes basic usage examples related to `actions` in our Kit SDK

PR #918, opened 2025-04-25
Adaptive Fee Release

yugure-orca:
# Adaptive Fee
adaptive_fee_001

The details of Adaptive Fee will be published as [PR915](https://github.com/orca-so/whirlpools/pull/915)

# Commits
| commits | source |
| -- | -- |
| `98477d43` - `0fd8c5c` | staging repository specific (sync) |
| `1a80c7e` - `8565ac6` | approved & audited as [PR71](https://github.com/orca-so/whirlpools-staging/pull/71) (Adaptive fee base PR) |
| `134c8b2` - `a1772ad` | approved & audited as [PR76](https://github.com/orca-so/whirlpools-staging/pull/76) (skip feat) |
| `39b3c74` - `b786159` |  approved & audited as [PR82](https://github.com/orca-so/whirlpools-staging/pull/82) (major swap feat) |
| [`ab64b1a`](https://github.com/orca-so/whirlpools/pull/918/commits/ab64b1a2deac984ca4cbb5cf743c665db3f6a669), [`570bfd6`](https://github.com/orca-so/whirlpools/pull/918/commits/570bfd6625b973d9521202167eb432dfd385d43d) | approved & audited as [PR76](https://github.com/orca-so/whirlpools-staging/pull/76) (skip feat), last 2 commits ([`3554bd1`](https://github.com/orca-so/whirlpools-staging/pull/76/commits/3554bd16fdaa800cc6ebaebde90b81055f1c25a5), [`ad64da5`](https://github.com/orca-so/whirlpools-staging/pull/76/commits/ad64da538f1cbdf65adf654913898eb19ab30f1b)) |
|  | 👇️ The followings are for merge work 👇️ |
| [`f761e9e`](https://github.com/orca-so/whirlpools/pull/918/commits/f761e9e974f2e6c0ef5c05d036a0bbd90da70389), [`533ca94`](https://github.com/orca-so/whirlpools/pull/918/commits/533ca940339f5d25ee100e090e18c6a253c3d5ed) | merge & resolve obvious conflicts |
| [`685eb52`](https://github.com/orca-so/whirlpools/pull/918/commits/685eb524bbe6df2bad6df919927d514905a90bb9) | add `PoolInitialized` event on `initialize_pool_with_adaptive_fee` (AdaptiveFee + EventEmission) |
| [`de1217d`](https://github.com/orca-so/whirlpools/pull/918/commits/de1217dbb63fdb45a5adaf44a7e9d027ef91acfa) | cargo fmt & lint |
| [`592e477`](https://github.com/orca-so/whirlpools/pull/918/commits/592e477fd0f9d6570b7786ed7c806b086401505b) | fix error code conflict between new features (resolve semantic conflicts) |
| [`4ce4c31`](https://github.com/orca-so/whirlpools/pull/918/commits/4ce4c31b2bf76ff633ad388852bebf55c203a211) | add test case for `PoolInitialized` event on `initialize_pool_with_adaptive_fee` |
| [`8370214`](https://github.com/orca-so/whirlpools/pull/918/commits/8370214e9dc26cc1fb10708936457a7b921e084b) | fix unintentional changes due to public/staging repo difference |
| [`2c7b3b0`](https://github.com/orca-so/whirlpools/pull/918/commits/2c7b3b03645628ea3d87bde2e4feb39346f2f324) | fix ts-sdk/client test error (`tickSpacingSeed` -> `feeTierIndexSeed`) |
| [`3328328`](https://github.com/orca-so/whirlpools/pull/918/commits/33283282ee12ea58343e0f7cb1c59c98180f32e6) | changeset |
| [`ed3cebe`](https://github.com/orca-so/whirlpools/pull/918/commits/ed3cebe5e7e09f373dd80e93b9cbf25b4f3192a2), [`6d3d8b9`](https://github.com/orca-so/whirlpools/pull/918/commits/6d3d8b92100862b677950e73ae4fa238b06d4f09) | fix minor merge issues through final self review |
| [`f5b6df2`](https://github.com/orca-so/whirlpools/pull/918/commits/f5b6df2a537d749823949d3c99197d7417c0fb89) | test case improvement only (ported from SDK update PR to aggregate `/programs` updates) |
| [`8ff5401`](https://github.com/orca-so/whirlpools/pull/918/commits/8ff54016afe38a3486170fbe411e591841ef7c5a) | address review comment (code comment update only) |

# SDKs
Swap quote logic updates for SDKs are under the review.
This PR contains minimum updates for SDKs only, so these are not major update. (Major updates will come soon)


legacy-sdk/whirlpool/src/instructions/swap-ix.ts
```diff
@@ -134,6 +135,12 @@ export function swapIx(
     },
   );
 
+  // HACK: to make Oracle account mutable without breaking change
+  // The official way to assemble instructions for pools that use AdaptiveFee is to add remaining accounts with isWritable set to true,
+  // but this hack that overrides the IDL requirements works.
+  invariant(ix.keys[10].pubkey.equals(oracle));
```
wjthieme, line 141: What is the reason for not just adding it to the idl?

Behavior would be exactly the same but the idl would be correct

wjthieme (approved):

legacy-sdk/whirlpool/src/instructions/swap-ix.ts
yugure-orca: Thanks for the check.
This part was discussed during development.

There are several possible patterns regarding "adding to the IDL".

### (1) Adding as a remaining account

AFAIK, the IDL does not have a way to define what kind of remaining accounts are accepted.
Therefore, it is not possible to add accounts that are received as remaining accounts to the IDL.

### (2) Making the oracle mut and marking it writable in the IDL

This approach was discarded during the design phase because it would be a breaking change.

Although it would be possible by directly editing the IDL, maintaining the state where the IDL is generated from the code would require adding `mut` to `oracle` account.

As a result, Anchor would always verify whether the oracle account is writable before starting the instruction.
This would cause the transaction from existing clients using the current swap instruction to be rejected.
Of course, for pools that use AdaptiveFee, we must reject transactions without a writable oracle, but we concluded that it is better to avoid breaking the behavior of conventional pools that do not use AdaptiveFee.

### (3) Receiving something like mut_oracle as a regular account

This would also affect clients currently using the swap instruction.
Therefore, if we want to use AdaptiveFee while maintaining compatibility, the approach is to choose one of the following:

- Pass the oracle as writable (ignoring the IDL)
- Pass a writable oracle as a remaining account (the oracle effectively becomes writable)

rawfalafel (approved):
:shipit:

legacy-sdk/whirlpool/src/utils/public/pda-utils.ts
```diff
@@ -151,19 +151,19 @@ export class PDAUtil {
...
   ) {
     return AddressUtil.findProgramAddress(
       [
         Buffer.from(PDA_FEE_TIER_SEED),
         whirlpoolsConfigAddress.toBuffer(),
-        new BN(tickSpacing).toArrayLike(Buffer, "le", 2),
+        new BN(feeTierIndex).toArrayLike(Buffer, "le", 2),
       ],
       programId,
     );
```
calintje, line 169: On the `WhirlpoolAccountFetcher`, we have methods `getFeeTier` and `getAdaptiveFeeTier`, because they are obviously different accounts. Even though the PDAs for these accounts use the same seeds, I'm wondering if this distinction should be reflected in `PDAUtil` as well... probably not?

calintje (approved):
LGTM!

legacy-sdk/whirlpool/src/instructions/adaptive-fee/initialize-adaptive-fee-tier-ix.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Parameters to initialize an AdaptiveFeeTier account.
+ *
+ * @category Instruction Types
+ * @param whirlpoolsConfig - PublicKey for the whirlpool config space that the adaptive fee-tier will be initialized for.
+ * @param feeTierPda - PDA for the adaptive fee-tier account that will be initialized
+ * @param funder - The account that would fund the creation of this account
+ * @param feeAuthority - Authority authorized to initialize fee-tiers and set customs fees.
+ * @param feeTierIndex - The index of the fee-tier in the whirlpools config.
+ * @param tickSpacing - The tick spacing this fee tier.
```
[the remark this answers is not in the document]
yugure-orca: Fixed🫡

legacy-sdk/whirlpool/src/instructions/adaptive-fee/set-preset-adaptive-fee-constants-ix.ts
```diff
@@ -0,0 +1,71 @@
...
+  presetDecayPeriod: number;
+  presetReductionFactor: number;
+  presetAdaptiveFeeControlFactor: number;
+  presetMaxVolatilityAccumulator: number;
+  presetTickGroupSize: number;
+  presetMajorSwapThresholdTicks: number;
+};
+
+/**
+ * Updates a adaptive fee tier account with new preset adaptive fee constants.
```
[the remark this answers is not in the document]
yugure-orca: Thanks!
Fixed🫡 (including same fix on other files)

legacy-sdk/whirlpool/src/utils/public/pda-utils.ts
yugure-orca: As you know, `FeeTier` and `AdaptiveFeeTier` have different data layouts.
Therefore, the fetcher provides separate methods for each.

On the other hand, `FeeTier` and `AdaptiveFeeTier` share the same PDA space.
This is an intentional collision to prevent both a `FeeTier` and an `AdaptiveFeeTier` with the same `feeTierIndex` from being created.
`PDAUtil` reflects this concept and provides only a single method.

Since it's rare for devs to create `FeeTier` or `AdaptiveFeeTier` accounts, this likely won't be an issue.
If there are enough requests, we can always add separate methods — though the logic would be exactly the same. 😅

PR #921, opened 2025-04-26
Fix "Transaction failed to sanitize accounts offsets" by matching signature count in simulations

calintje:
### Background
The tx-sender SDK encountered "Transaction failed to sanitize accounts offsets correctly" errors in certain multi-signer scenarios. We identified that the `estimate_compute_units` function was simulating transactions with a fixed signature count that didn't match the actual number of signers used in the final transaction, causing a mismatch between simulation and execution.

### Changes
- Added signers parameter to `estimate_compute_units` function to ensure consistent signature counts
- Updated transaction simulation to use `vec![Signature::default(); signers.len()]` for creating properly sized signature arrays
#### After comments:
- removed `payer` from `build_transaction` and kept `signers`, bringing it in line with `build_and_send_transaction`.
- Updated Rust docs with example handling of `additional_signers`.
- Fixed `actions` examples

### Testing
Manually verified that transactions with multiple signers now compile, sign, and submit correctly.

### DIFF
--- a/rust-sdk/tx-sender/src/lib.rs
+++ b/rust-sdk/tx-sender/src/lib.rs
@@ -28,7 +28,7 @@ pub async fn build_and_send_transaction(
         .first()
         .ok_or_else(|| "At least one signer is required".to_string())?;
     // Build transaction with compute budget and priority fees
-    let mut tx = build_transaction(instructions, *payer, signers, address_lookup_tables).await?;
+    let mut tx = build_transaction(instructions, signers, address_lookup_tables).await?;
     // Serialize the message once instead of for each signer
     let serialized_message = tx.message.serialize();
     tx.signatures = signers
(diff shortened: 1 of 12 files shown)


parhim (approved):

rust-sdk/tx-sender/src/lib.rs
```diff
@@ -49,6 +49,7 @@ pub async fn build_and_send_transaction(
 pub async fn build_transaction(
     mut instructions: Vec<Instruction>,
     payer: &dyn Signer,
+    signers: &[&dyn Signer],
```
parhim, line 52: would it make sense to treat first signer as the payer ? and just have signers and no payer?
also docs probably need to be updated accordingly

calintje: removed `payer` from `build_transaction` and kept `signers`, bringing it in line with `build_and_send_transaction`.

Updated Rust tocs with example handling of `additional_signers`.
Fixed `actions` examples

wjthieme (approved):

PR #926, opened 2025-04-30
Adaptive Fee SDK update (base change)

yugure-orca:
This PR requires MAJOR version up of SDK because it has breaking changes.

# Legacy SDK `/legacy-sdk`
### Adaptive Fee supports
- added Oracle account
- added AdaptiveFeeTier account
- updated swap quote logic (`FeeRateManager` has been ported, but Rust -> TS conversion, so several test cases has been added)
- (WhirlpoolIx update has been included in contract release PR)
- renamed `tickSpacingSeed` -> `feeTierIndexSeed` (Whirlpool account)
- renamed `tickSpacing` -> `feeTierIndex` (PDA seeds of Whirlpool account and FeeTier account)

### Drop deprecated features
In v0.13.x, these features has been deprecated and clearly declared that these features will be dropped in the future release.
- PriceModule
- WhirlpoolRouter
- PoolGraph

# New SDK (Rust) `/rust-sdk`
### Adaptive Fee supports
- added Oracle account with gPA helper functions
- added AdaptiveFeeTier account with gPA helper functions
- updated swap quote logic (`FeeRateManager` is almost same to contract code (Rust))
- renamed `tick_spacing_seed` -> `fee_tier_index_seed` (Whirlpool account)
- renamed `tick_spacing` -> `fee_tier_index` (PDA seeds of Whirlpool account and FeeTier account)

# New SDK (TS) `/ts-sdk`
### Adaptive Fee supports
- added Oracle account with gPA helper functions
- added AdaptiveFeeTier account with gPA helper functions
- updated swap quote logic (WASM, no TS specific implementation)
- renamed `tickSpacingSeed` -> `feeTierIndexSeed` (Whirlpool account)
- renamed `tickSpacing` -> `feeTierIndex` (PDA seeds of Whirlpool account and FeeTier account)


legacy-sdk/whirlpool/src/quotes/public/swap-quote.ts
```diff
@@ -54,8 +71,10 @@ export type SwapQuoteParam = {
   aToB: boolean;
   amountSpecifiedIsInput: boolean;
   tickArrays: TickArray[];
+  oracleData: OracleData | null;
```
calintje, line 74: I noticed that `timestampInSeconds` was added as an optional field (using ?), which aligns with `timestampInSeconds`.

Given that, would it also make sense to make oracleData optional (`oracleData?: OracleData`) instead of required-but-nullable?

Right now, the difference between undefined and null adds some inconsistency, especially since users calling `swapQuoteWithParams` directly might not always have oracle data available.

I know this update will be a breaking change anyway, but making `oracleData` optional might reduce breakage and make the API a bit smoother for users who cache data or want to skip the oracle lookup.

Of course, if there's a specific reason to require the key explicitly, feel free to disregard — just wanted to raise the thought.

calintje (approved):
Thanks for the detailed implementation. It was interesting to follow how the math translated into code. I learned a lot going through it.

rust-sdk/client/src/gpa/adaptive_fee_tier.rs
```diff
@@ -0,0 +1,96 @@
...
+            ),
+            AdaptiveFeeTierFilter::MajorSwapThresholdTicks(major_swap_threshold_ticks) => {
+                RpcFilterType::Memcmp(Memcmp::new_raw_bytes(
+                    126,
+                    major_swap_threshold_ticks.to_le_bytes().to_vec(),
+                ))
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
yugure-orca: I excluded fields that clearly change over time, but included constant fields.
While considering which fields might actually be useful was one option, I noticed that other implementations covered most of the fields, so I decided to follow their style.

legacy-sdk/whirlpool/src/quotes/public/swap-quote.ts
yugure-orca: Good point!

I’d like to leave it as is. Here’s why:

As you know, timestampInSeconds is made optional because, in most cases, it's reasonable to rely on the system timestamp, and manual tuning is only needed in specific scenarios.

On the other hand, oracleData is not something that can be optionally specified—it’s required for pools that use AdaptiveFee, and must be null for pools that don’t use AdaptiveFee (since those pools currently have no associated oracle). If omitted, it will result in an error for AdaptiveFee-enabled pools.

I’ve intentionally made this a breaking change so that the "compiler" can catch this major change (i.e., the requirement of oracle data), and prompt users to update their code accordingly. 

The intention is to prevent potentially problematic code from compiling, rather than allowing oracleData to be optional and risk runtime errors caused by insufficient consideration of Oracle requirements.

In the legacy SDK, there are multiple swap quote functions—for example, swapQuoteByInputToken and swapQuoteByOutputToken handle fetching internally, abstracting away some of the complexity. These are easier to use but are less efficient in terms of performance. With this update, swapQuoteByInputToken still fetches the oracle internally, so it doesn’t introduce a breaking change 🙂

In other words, the update ensures that swapQuoteByInputToken, which prioritizes ease of use, continues to work without requiring users to be aware of the Oracle. Meanwhile, swapQuoteWithParams, intended for advanced users who fully understand the system, strictly requires Oracle data and does not allow it to be omitted.

rust-sdk/core/src/math/adaptive_fee.rs
```diff
@@ -0,0 +1,500 @@
...
+use crate::{
+    AdaptiveFeeConstantsFacade, AdaptiveFeeInfo, AdaptiveFeeVariablesFacade, CoreError,
+    ADAPTIVE_FEE_CONTROL_FACTOR_DENOMINATOR, FEE_RATE_HARD_LIMIT, INVALID_TIMESTAMP,
+    MAX_REFERENCE_AGE, MAX_TICK_INDEX, MIN_TICK_INDEX, REDUCTION_FACTOR_DENOMINATOR,
+    VOLATILITY_ACCUMULATOR_SCALE_FACTOR,
+};
+use ethnum::U256;
+
+#[derive(Debug)]
+pub enum FeeRateManager {
```
wjthieme, line 11: Do any of these functions need to be available in the js core library? If so they would probably need to be top-level functions and need the 'wasm_expose' macro. If not should we make them pub(crate) so they are also not exposed in rust-core?

wjthieme (approved):
Small nit but LGTM 🚀

rust-sdk/core/src/math/adaptive_fee.rs
yugure-orca: Good catch!
I think that users should not use logics in `FeeRateManager`, so I updated from `pub` to `pub(crate)`.

PR #929, opened 2025-05-02
Use num_required_signers from Message when building VersionedTransaction

jshiohaha:
**Details**

The current implementation relies on the caller wanting to build a transaction to know how many signers a transaction will need and have access to those signers. It's possible to generate a NoopSigner for each signer, but 
1. it seems like more work than should be required when building a transaction, and
2. If the caller gets the count wrong, the build transaction function will fail due to a simulation failure of the following nature: 

```bash
RPC response error -32602: invalid transaction: Transaction failed to sanitize accounts offsets correctly
```

This change addresses the same issue as [this previous PR](https://github.com/orca-so/whirlpools/pull/921), but uses the `num_required_signers` from the compiled Message.

📔 I added a major version bump here because the `build_transaction` and `build_transaction_with_config` function signatures changed. The `build_and_send_transaction` and `build_and_send_transaction_with_config` stayed the same.

**Testing**

I'm consuming this library in wavebreak, and verified that the update works as expected.


parhim (approved):
very nice

PR #938, opened 2025-05-08
Fix: check if price is zero in open_position_instructions for rust-sdk

boosik-sol:
This PR is fixing bug when user create `open_position_instructions` with `lower_price == 0` or `upper_price == 0`.
It is impossible to use zero price during calculating tick index, and now SDK just makes panic.
So I've implemented condition to check if price is zero or not, if price is zero it will return error and message, I expect sdk user will understand more about it.

- [ ] Specifies the conditions when this error occurs in docs
- [ ] Simple test


rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -593,8 +593,8 @@ pub async fn open_full_range_position_instructions(
 ///
 /// * `rpc` - A reference to the Solana RPC client.
 /// * `pool_address` - The public key of the liquidity pool.
-/// * `lower_price` - The lower bound of the price range for the position.
-/// * `upper_price` - The upper bound of the price range for the position.
+/// * `lower_price` - The lower bound of the price range for the position. It have to be greater than 0.0.
```
calintje, line 596: nit: have = has

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -658,6 +659,9 @@ pub async fn open_position_instructions(
     slippage_tolerance_bps: Option<u16>,
     funder: Option<Pubkey>,
 ) -> Result<OpenPositionInstruction, Box<dyn Error>> {
+    if lower_price == 0.0 || upper_price == 0.0 {
```
calintje, line 662: Just a thought:
If I were a developer using `orca_whirlpools` (which is our 'high-level' SDK) and specify a range for my position as:
```
lower_price: 0.0
upper_price: 0.1
```
my intention is to set the lower bound at the minimum possible value. So maybe instead of returning an error, use the MIN initializable tick index as the lower bound.
WDYT?

calintje: How about adding an assertion in the `core` library here:
https://github.com/orca-so/whirlpools/blob/5ba1a95a0669758488351e7269bc0dbea2e185ef/rust-sdk/core/src/math/tick.rs#L60-L64

As you stated in your analysis, if `sqrt_price: u128 == 0` -> `msb: u32 == 4294967231` (will underflow)

rust-sdk/whirlpool/src/increase_liquidity.rs
boosik-sol: oh, i will fix mistypo

rust-sdk/whirlpool/src/increase_liquidity.rs
boosik-sol: I think returning explicit error is one of the role of SDK. And there is possibility of occuring side effect with additional condition. (But this is just my personal opinion! :D)

How about using some flag for minimum & maximum price value? like
```
lower_price = MIN_PRICE;
upper_price = MAX_PRICE;
```

boosik-sol: so I think the important thing is that whether whirlpool accepts zero price or not. If it accepts, I think it is ok to return minimum tick index, if it doesn't accept, handling error is better @yugure-orca

calintje: I believe your reasoning is correct. Strictly speaking, the program instruction doesn't accept price (or sqrt price), but a tick index. The aim of the high-leve SDKs (`orca_whirlpools` and `@orca-so/whirlpools`) is to abstract the complexity around ticks away from the user as much as possible.

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -658,6 +659,9 @@ pub async fn open_position_instructions(
...
+        return Err("Price cannot be zero".into());
+    }
```
wjthieme, line 664: Is this also an issue in the ts-sdk? I would suggest checking against MIN_SQRT_PRICE and MAX_SQRT_PRICE instead because those are the limitations of the whirlpool program

.changeset/upset-dodos-take.md
```diff
@@ -0,0 +1,2 @@
+---
```
wjthieme, line 1: Could you add an actual changeset so we can publish the change and get a changelog?

rust-sdk/whirlpool/src/increase_liquidity.rs
boosik-sol: Ts-sdk only supports the number of token, or liquidity. I think this option is good.
```typescript
export type IncreaseLiquidityQuoteParam =
  | {
      /** The amount of liquidity to increase. */
      liquidity: bigint;
    }
  | {
      /** The amount of Token A to add. */
      tokenA: bigint;
    }
  | {
      /** The amount of Token B to add. */
      tokenB: bigint;
    };
 
```

boosik-sol: oh I found that openPositionInstructions in ts-sdk is also using lower-upper price. Checking if we need to write same logic with rust-sdk.

boosik-sol:
```typescript
export async function openPositionInstructions(
  rpc: Rpc<
    GetAccountInfoApi &
      GetMultipleAccountsApi &
      GetMinimumBalanceForRentExemptionApi &
      GetEpochInfoApi
  >,
  poolAddress: Address,
  param: IncreaseLiquidityQuoteParam,
  lowerPrice: number,
  upperPrice: number,
  slippageToleranceBps: number = SLIPPAGE_TOLERANCE_BPS,
  funder: TransactionSigner = FUNDER,
): Promise<OpenPositionInstructions> {
```

.changeset/upset-dodos-take.md
boosik-sol: Done!

rust-sdk/whirlpool/src/increase_liquidity.rs
boosik-sol: I checked with ts-sdk but it didn't occur panic or error. But i think it is good to add same logic both rust and ts for reliabilty.

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -678,6 +680,13 @@ pub async fn open_position_instructions(
...
 
+    let lower_sqrt_price = price_to_sqrt_price(lower_price, decimals_a, decimals_b);
+    let upper_sqrt_price = price_to_sqrt_price(upper_price, decimals_a, decimals_b);
+
+    if lower_sqrt_price < MIN_SQRT_PRICE || upper_sqrt_price > MAX_SQRT_PRICE {
+        return Err("Price out of range".into());
+    }
+
     let lower_tick_index = price_to_tick_index(lower_price, decimals_a, decimals_b);
     let upper_tick_index = price_to_tick_index(upper_price, decimals_a, decimals_b);
```
wjthieme, line 692: Actually now that I read it again. Better I guess to compare lower_tick_index and upper_tick_index against MIN_TICK_INDEX and MAX_TICK_INDEX. There is no guarantee that the user entered a lower number for lower_price than upper_price. I think in `internal_open_position` it orders the tick indexes correctly. Maybe we should add the check too in that function (after they are ordered correctly?)

Where exactly does the div/0 error occur?

boosik-sol: This is happend in `sqrt_price_to_tick_index` in `rust-sdk/core/src/math/tick.rs`.
Flows is like...
* `open_position_instructions with lower_price=0 or upper_price=0`
* `price_to_tick_index with price=0`
* `sqrt_price_to_tick_index with sqrt_price=0`
* `panic on let msb: u32 = 128 - sqrt_price_x64.leading_zeros() - 1;` Because result will be 1, but type of msb is unsigned integer.

so what I firstly thought that, it's better to let user know like this value may not be accepted in whirlpool.
And this bug is occured before getting tick index, so it is hard to directly compare with tick index, I think we can add some defense logic during calculating tick index. like... as @calintje said, we can add some checking logic in `sqrt_price_to_tick_index` or other calculation function.

odcheung: would it be simpler if price_to_tick_index & price_to_sqrt_price threw the error and these functions just bubbles up the error?

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,96 @@
...
+ * };
+ * const { instructions } = await transferLockedPositionInstructions(
+ *   devnetRpc,
+ *   param,
+ *   wallet
+ * );
+ *
+ * console.log(`Instructions: ${instructions}`);
+ */
+export async function transferLockedPositionInstructions(
```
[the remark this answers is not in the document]
boosik-sol: oh I checked out my branch but missed it. Thanks!

rust-sdk/whirlpool/src/increase_liquidity.rs
boosik-sol: how about just return error in `sqrt_price_to_tick_index` when price = 0?

boosik-sol: how about just throwing error in `sqrt_price_to_tick_index` when price = 0? one more depth compared to price_to_tick_index & price_to_sqrt_price. It's the point of panic.

calintje:
As I mentioned in a previous comment, I agree that `sqrt_price_to_tick_index` from our Core library should throw an error if `sqrt_price: u128 == 0`, as this helps prevent an underflow.

That said, I would suggest that the high-level `open_position_instructions` function **should not** throw in this case. Instead, it could assume the user intends to set the lower boundary of their position at the lowest possible tick.

If we require an error to be thrown here, the user would then need to:
- fetch the pool’s `tick_spacing`
- compute the lowest valid `tick_index` based on that spacing
- convert that `tick_index` back to a price

or alternatively:
- use `MIN_SQRT_PRICE`, which the function would then convert back to `sqrt_price` and `tick_index`, potentially introducing rounding errors or edge-case failures.

If that’s the expected flow, it raises the question: why offer price as an input at all? Since the purpose of high-level functions is to abstract away low-level tick logic, I believe it's more user-friendly to avoid throwing an error in this case.

boosik-sol:
In ts-sdk, there is no panic or error if `lower_price == 0`. But it returns tick index as `-1561061633`, it is invalid tick. So how about add same logic both rust-sdk and ts-sdk?

wjthieme:
> That said, I would suggest that the high-level open_position_instructions function should not throw in this case. Instead, it could assume the user intends to set the lower boundary of their position at the lowest possible tick.

Don't fully agree. Chaning inputs on the users behalf is usually confusing. If I input .0000001 and it gets adjusted to .000001 without me knowing that would be weird (rather have an error in that case)

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -678,6 +681,13 @@ pub async fn open_position_instructions(
     let decimals_a = mint_a.decimals;
     let decimals_b = mint_b.decimals;
 
+    let lower_sqrt_price = price_to_sqrt_price(lower_price, decimals_a, decimals_b);
+    let upper_sqrt_price = price_to_sqrt_price(upper_price, decimals_a, decimals_b);
+
+    if lower_sqrt_price < MIN_SQRT_PRICE || upper_sqrt_price > MAX_SQRT_PRICE {
```
wjthieme, line 687: Think instead we should checck the tick_indexes in `internal_open_position` after they have been ordered. There is no guarantee that lower_sqrt_price < upper_sqrt_price in this function. Also this requires and extra `price_to_sqrt_price` call

rust-sdk/whirlpool/src/increase_liquidity.rs
odcheung: i think that's fine. but we might want to reword the error message during the bubble up to make sure the error message relates to the user action. Ex. user entering an invalid price on the high level method but getting an error message on invalid sqrt-price/tick-index may be confusing.

rust-sdk/core/src/math/price.rs
```diff
@@ -21,9 +21,12 @@ const Q64_RESOLUTION: f64 = 18446744073709551616.0;
 /// # Returns
 /// * `u128` - The sqrt priceX64
 #[cfg_attr(feature = "wasm", wasm_expose)]
-pub fn price_to_sqrt_price(price: f64, decimals_a: u8, decimals_b: u8) -> U128 {
+pub fn price_to_sqrt_price(price: f64, decimals_a: u8, decimals_b: u8) -> Result<U128, CoreError> {
```
wjthieme, line 24: This is a breaking change (= the signature of the function changes from `U128` to `Result<U128, CoreError>`). Should we just return 0 if price is 0?

wjthieme: Only if there is no (clean) non-breaking way to do this should we resort to breaking changes

wjthieme: Agree that a Price struct would be more clear but I don't think we would be able to:
* Add this in a non-breaking way
* Make this work using wasm (so that we can use it in js)

Let's return 0 from `price_to_sqrt_price` or add in the rustdoc that a price of 0 would result in a panic. I'd say for a small change like this doing a breaking-change is not worth it (bump major version, all downstream consumers will need to manually update major version, different major versions are not backwards compatible, etc.).

I think to handle this non-breaking our only option would be to handle price == 0 case from `open_position_instructions`

boosik-sol: I understand it. let's do this way! Thank you :D

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -593,8 +593,8 @@ pub async fn open_full_range_position_instructions(
 ///
 /// * `rpc` - A reference to the Solana RPC client.
 /// * `pool_address` - The public key of the liquidity pool.
-/// * `lower_price` - The lower bound of the price range for the position.
-/// * `upper_price` - The upper bound of the price range for the position.
+/// * `lower_price` - The lower bound of the price range for the position. It returns error if the lower_price <= 0.0.
+/// * `upper_price` - The upper bound of the price range for the position. It returns error if the upper_price <= 0.0.
```
boosik-sol, line 597: Add comments in `open_position_instructions` that either `lower_price` or `upper_price` is less or equal than 0.0.

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -615,6 +615,8 @@ pub async fn open_full_range_position_instructions(
 /// - The pool or token mint accounts are not found or invalid.
 /// - Any RPC request fails.
 /// - The pool is a Splash Pool, as they only support full-range positions.
+/// - The lower price is less or equal to 0.0.
+/// - The upper price is less or equal to 0.0.
```
boosik-sol, line 619: Add comments in open_position_instructions that either lower_price or upper_price is less or equal than 0.0.

rust-sdk/core/src/constants/error.rs
```diff
@@ -52,3 +52,7 @@ pub const INVALID_TICK_ARRAY_SEQUENCE: CoreError = "Invalid tick array sequence"
 
 #[cfg_attr(feature = "wasm", wasm_expose)]
 pub const INVALID_ADAPTIVE_FEE_INFO: CoreError = "Invalid adaptive fee info";
+
+#[cfg_attr(feature = "wasm", wasm_expose)]
+pub const INVALID_FLOAT_PRICE: CoreError =
```
wjthieme, line 57: nit: is this being used?

wjthieme (approved):
Remind me again, do we need this for ts-sdk as well?

Otherwise LGTM

boosik-sol:
> Remind me again, do we need this for ts-sdk as well?
> 
> Otherwise LGTM

Typescript didn't return any panic, but it also returned invalid tick index. Meaning that if I use zero price, it will return index that is lower than MIN_TICK_INDEX. so I think it'd be better to return exception when user use zero price in ts-sdk, but how about doing it in other PR?

PR #940, opened 2025-05-09
Implements transfer_locked_position instruction in ts-sdk

boosik-sol:
This PR is implementing `transfer_locked_position` instruction in ts-sdk.

- [x] implementing instruction
- [x] unit test
- [x] testing in devnet, mainnet


ts-sdk/whirlpool/tests/transferLockedPosition.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    for (const [name, setup] of poolTypes) {
+      const [mintAKey, mintBKey] = name.split("-");
+      const mintA = mints.get(mintAKey)!;
+      const mintB = mints.get(mintBKey)!;
+      pools.set(name, await setup(mintA, mintB, tickSpacing));
+    }
+  });
+
+  const testTransferLockedPositionInstructions = async (
```
boosik-sol, line 78: I intended to test following
1. position owner opens new position
2. lock it
3. try to transfer locked position
4. after transfer, verify if position is exsit, and locked.

ts-sdk/whirlpool/tests/utils/tokenExtensions.ts
```diff
@@ -9,19 +9,20 @@ import {
...
+import type { Address, IInstruction, KeyPairSigner } from "@solana/kit";
 import { sendTransaction, signer } from "./mockRpc";
 import { getCreateAccountInstruction } from "@solana-program/system";
 import { DEFAULT_ADDRESS } from "../../src/config";
 import { getNextKeypair } from "./keypair";
 
 export async function setupAtaTE(
   mint: Address,
-  config: { amount?: number | bigint } = {},
+  config: { amount?: number | bigint; signer?: KeyPairSigner } = {},
```
boosik-sol, line 20: added signer(Optional) field because in this case I need to Ata without default signer.
But if there are some better ways, please let me know :D

ts-sdk/whirlpool/tests/utils/tokenExtensions.ts
```diff
@@ -30,7 +31,7 @@ export async function setupAtaTE(
   instructions.push(
     getCreateAssociatedTokenIdempotentInstruction({
       mint,
-      owner: signer.address,
+      owner: owner.address,
       ata: ata[0],
       payer: signer,
```
boohyunsik, line 36: If I set payer to owner, error is occurred because owner doesn't have enough balance.

boosik-sol:
No lockPosition test code, so need to test it first.

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,117 @@
...
+} from "@solana/kit";
+import { FUNDER } from "./config";
+import { wrapFunctionWithExecution } from "./actionHelpers";
+
+/**
+ * Parameters for transferring a locked position.
+ */
+export type TransferLockedPositionParam = {
+  /** The address of the position to transfer. */
+  position: Address;
```
wjthieme, line 21: Same comment as on #943. This now just wraps the `getTransferLockedPositionInstruction` function without adding anything.

ts-sdk/whirlpool/tests/transferLockedPosition.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const whirlpool = pools.get(poolName)!;
+    const param = { liquidity: 10_000n };
+
+    // Position owner has position, can lock it, and transfer it.
+    const positionOwner = signer;
+    const receiver = getNextKeypair();
+    setDefaultFunder(positionOwner);
+
+    // Open position by position owner
+    const { instructions, positionMint } = await openPositionInstructions(
```
wjthieme, line 92: Same thing here, you can use the `setupPosition` function here which was made especially for this

ts-sdk/whirlpool/tests/lockPosition.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    );
+    assert(
+      lockConfig.data.positionOwner === signer.address,
+      "Lock config position owner is not the same as the signer",
+    );
+  };
+
+  for (const poolName of poolTypes.keys()) {
+    for (const positionTypeName of positionTypes.keys()) {
+      const positionNameTE = `TE ${poolName} ${positionTypeName}`;
```
boosik-sol, line 148: Question) It seems if we open position with `setupPosition`(not supporting token extension), we couldn't lock position because it's token mint will be legacy SPL. Is there any effect in real service? Like, if users open their position with `open_position_instruction` (not `open_position_instruction_with_token_extension`, they might not be able to lock position.

ts-sdk/whirlpool/tests/utils/tokenExtensions.ts
```diff
@@ -42,7 +43,7 @@ export async function setupAtaTE(
       getMintToInstruction({
         mint,
         token: ata[0],
-        mintAuthority: signer,
+        mintAuthority: owner,
```
wjthieme, line 46: I think this still needs to be `signer` because owner is not the mint_authority

ts-sdk/whirlpool/tests/lockPosition.test.ts
wjthieme: Isn't there a `setupTEPosition` function?

boosik-sol: Yap, I tested only positions made with `setupTEPosition`. But what I am wondering is that don't we need to support `lockPosition` with position made with `setupPosition`. This case was failed. (So I didn't add this case.)

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Parameters for transferring a locked position.
+ */
+export type TransferLockedPositionParam = {
+  /** The address of the position mint. */
+  positionMintAddress: Address;
+
+  /** The address of the destination token account. */
+  detinationTokenAccount: Address;
```
wjthieme, line 33: This one can be derived from the `receiver`

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  /** The address of the lock config. */
+  lockConfig: Address;
```
wjthieme, line 36: This one can be derived from the `position` address

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,129 @@
...
+} from "@solana/kit";
+import { FUNDER } from "./config";
+import { wrapFunctionWithExecution } from "./actionHelpers";
+import assert from "assert";
+import { findAssociatedTokenPda } from "@solana-program/token";
+
+/**
+ * Parameters for transferring a locked position.
+ */
+export type TransferLockedPositionParam = {
```
wjthieme, line 28: I would just do direct params on the function call instead of a type (to stay in line with the other functions)

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,129 @@
...
+ *   wallet
+ * );
+ *
+ * console.log(`Instructions: ${instructions}`);
+ */
+export async function transferLockedPositionInstructions(
+  rpc: Rpc<
+    GetAccountInfoApi &
+      GetMultipleAccountsApi &
+      GetMinimumBalanceForRentExemptionApi &
```
wjthieme, line 87: Do we need `GetMultipleAccountsApi`, `GetMinimumBalanceForRentExemptionApi` and `GetEpochInfoApi` here?

ts-sdk/whirlpool/tests/lockPosition.test.ts
```diff
@@ -0,0 +1,154 @@
+import { describe, it, beforeAll } from "vitest";
```
wjthieme, line 1: Is this from the other PR?

wjthieme (commented):
Almost there! Couple of small comments left

ts-sdk/whirlpool/tests/lockPosition.test.ts
boosik-sol: No, I include `lockPosition` test case in this PR. Because `lockPosition` and `transferLockedPosition` share same logics, but there is no `lockPosition` test code.

ts-sdk/whirlpool/src/transferLockedPosition.ts
boosik-sol: as I commented other PR, I will look at them and let you know my idea.

ts-sdk/whirlpool/src/transferLockedPosition.ts
```diff
@@ -0,0 +1,119 @@
...
+    tokenProgram: positionMint.programAddress,
+  });
+
+  const destinationTokenAccount = await findAssociatedTokenPda({
+    owner: receiver,
+    mint: positionMintAddress,
+    tokenProgram: positionMint.programAddress,
+  });
+
+  const lockConfigPda = await getProgramDerivedAddress({
```
boosik-sol, line 86: @wjthieme Could we use this with @orca-so/whirlpool-client? I think it is not exported. Or, I will move this to `utils.ts` when I fix `lockPosition.ts`

ts-sdk/whirlpool/src/transferLockedPosition.ts
boosik-sol: it was very simple and regret not to look about this. yeah we don't need these and removed.

ts-sdk/whirlpool/tests/lockPosition.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      "Lock config position is not the same as the position",
+    );
+    assert(
+      lockConfig.data.positionOwner === signer.address,
+      "Lock config position owner is not the same as the signer",
+    );
+  };
+
+  for (const poolName of poolTypes.keys()) {
+    for (const positionTypeName of positionTypes.keys()) {
```
odcheung, line 147: can we try to use test.for?

ts-sdk/whirlpool/tests/lockPosition.test.ts
odcheung: I recommend that. we should have a test case to verify that non-TE tokens cannot use transfer lock just to be complete.

ts-sdk/whirlpool/tests/lockPosition.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  ["A-TEFee", setupWhirlpool],
+]);
+
+const positionTypes = new Map([
+  ["equally centered", { tickLower: -100, tickUpper: 100 }],
+  ["one sided A", { tickLower: -100, tickUpper: -1 }],
+  ["one sided B", { tickLower: 1, tickUpper: 100 }],
+]);
+
+describe("LockPosition instructions", () => {
```
odcheung, line 57: Suggest we add more test cases to cover all the parameter that this function takes in:
ex: 
1. receiver (self vs un-related wallet)
2. authority (funder vs another wallet that owns the position vs another wallet that doesn't own the position)
3. position mint types (TE vs non-TE) ✅ 
4. position lock state (locked vs unlocked) ✅

ts-sdk/whirlpool/tests/lockPosition.test.ts
boosik-sol: There are several codes using this pattern, so how about fix them together?

ts-sdk/whirlpool/tests/lockPosition.test.ts
odcheung: Sorry to clarify, this function takes in the receiver as where the locked position can be transferred to. Assuming walletA owns the locked position today, edge cases are:
1. What if walletA sets walletA as receiver? (tries to send to self)
2. walletA -> walletB (normal case)

yugure-orca (commented):
Could you apply the following two changes?
The PR has been open for a while, so I’d like to make only these changes and then merge it.

- <s>Export lockConfig.ts in ts-sdk/client/src/pda/index.ts and</s> ([PR1034](https://github.com/orca-so/whirlpools/pull/1034) did it) use `getLockConfigAddress` to derive lock config PDA.
- Add an instruction to idempotently initialize the destinationTokenAccount (ATA).

Since `@orca-so/whirlpools` provides high-level functions, it will create required ATAs. However, `transfer_locked_position` instruction does not initialize the destination in the instruction itself, so I think it’s better to provide an ATA initialization instruction on the client side.

Regarding test cases, scenarios blocked at the contract level (e.g., non-TE positions cannot be locked or transferred) are already covered in the legacy-sdk integration tests.
(There’s no need to remove them.) It’s not mandatory for ts-sdk tests to verify that the contract rejects these cases.

PR #943, opened 2025-05-09
Implements resetPositionRange instructions for ts-sdk

boosik-sol:
This PR implements `resetPositionRange` instructions in typescript-sdk

* [ ] : implement instruction
* [ ] : unit test
* [ ] : devnet/mainnet test

### DIFF
--- a/ts-sdk/whirlpool/src/resetPositionRange.ts
+++ b/ts-sdk/whirlpool/src/resetPositionRange.ts
@@ -1,0 +1,101 @@
+import { getResetPositionRangeInstruction } from "@orca-so/whirlpools-client";
+import type {
+  Address,
+  GetAccountInfoApi,
+  GetEpochInfoApi,
+  GetMinimumBalanceForRentExemptionApi,
+  GetMultipleAccountsApi,
+  Rpc,
+  IInstruction,
+  TransactionSigner,
+} from "@solana/kit";
+
+/**
+ * Parameters to reset position range.
+ *
+ * @param funder - The account that create original position.
+ * @param positionAuthority - Authority that owns the token corresponding to this desired position.
+ * @param position - PublicKey for the position which will be reset.
+ * @param positionTokenAccount - The associated token address for the	position token in the owners wallet.
+ * @param newTickLowerIndex - The tick specifying the lower end of the position range.
+ * @param newTickUpperIndex - The tick specifying the upper end of the position range.
+ * @param whirlpool - PublicKey for the whirlpool that the position belongs to.
+ */
+export type ResetPositionRangeParams = {
+  funder: TransactionSigner<Address>;
+  positionAuthority: TransactionSigner<Address>;
+  position: Address;
+  positionTokenAccount: Address;
+  newTickLowerIndex: number;
+  newTickUpperIndex: number;
+  whirlpool: Address;
+};
+
+/**
+ * Represents the instructions for resetting a position range.
+ */
+export type ResetPositionRageInstructions = {
+  instructions: IInstruction[];
+};
+
+/**
+ * Generates instructions to reset a position range.
+ *
+ * @param {SolanaRpc} rpc - The Solana RPC client.
+ * @param {ResetPositionRangeParams} params - The parameters for resetting a position range.
+ * @returns {Promise<ResetPositionRageInstructions>} A promise that resolves to an object containing instructions.
+ *
+ * @example
+ * import { resetPositionRangeInstructions } from "@orca-so/whirlpools";
+ * import { createSolanaRpc, devnet } from "@solana/kit";
+ *
+ * await setWhirlpoolsConfig("solanaDevnet");
+ * const devnetRpc = createSolanaRpc(devnet("https://api.devnet.solana.com"));
+ * const wallet = await loadWallet();
+ *
+ * const position = address("5uiTr6jPdCXNfBWyfhAS9HScpkhGpoPEsaKcYUDMB2Nw");
+ * const positionTokenAccount = address("2t3H9fSEJftE6TS7kgTYqRbnhdRUkCRfxdULybFTgWPu");
+ * const whirlpool = address("3KBZiL2g8C7tiJ32hTv5v3KM7aK9htpqTw4cTXz1HvPt");
+ * const newTickLowerIndex = 300_000;
+ * const newTickUpperIndex = 400_000;
+ *
+ * const instructions = await resetPositionRangeInstructions(
+ *   devnetRpc,
+ *   {
+ *     funder: wallet,
+ *     positionAuthority: wallet,
+ *     position,
+ *     positionTokenAccount,
+ *     newTickLowerIndex,
+ *     newTickUpperIndex,
+ *     whirlpool,
+ *   }
+ * );
+ */
+export async function resetPositionRangeInstructions(
+  rpc: Rpc<
+    GetAccountInfoApi &
+      GetMultipleAccountsApi &
+      GetMinimumBalanceForRentExemptionApi &
+      GetEpochInfoApi
+  >,
+  params: ResetPositionRangeParams,
+): Promise<ResetPositionRageInstructions> {
+  const instructions: IInstruction[] = [];
+
+  instructions.push(
+    getResetPositionRangeInstruction({
+      funder: params.funder,
+      positionAuthority: params.positionAuthority,
+      position: params.position,
+      positionTokenAccount: params.positionTokenAccount,
+      newTickLowerIndex: params.newTickLowerIndex,
+      newTickUpperIndex: params.newTickUpperIndex,
+      whirlpool: params.whirlpool,
+    }),
+  );
+
+  return {
+    instructions,
+  };
+}
--- a/ts-sdk/whirlpool/tests/resetPositionRange.test.ts
+++ b/ts-sdk/whirlpool/tests/resetPositionRange.test.ts
@@ -1,0 +1,147 @@
+import { describe, it, beforeAll } from "vitest";
+import type { Address } from "@solana/kit";
+import { setupAta, setupMint } from "./utils/token";
+import {
+  setupAtaTE,
+  setupMintTE,
+  setupMintTEFee,
+} from "./utils/tokenExtensions";
+import { setupWhirlpool } from "./utils/program";
+import { openPositionInstructions } from "../src/increaseLiquidity";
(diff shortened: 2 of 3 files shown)


ts-sdk/whirlpool/tests/resetPositionRange.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    for (const [name, setup] of poolTypes) {
+      const [mintAKey, mintBKey] = name.split("-");
+      const mintA = mints.get(mintAKey)!;
+      const mintB = mints.get(mintBKey)!;
+      pools.set(name, await setup(mintA, mintB, tickSpacing));
+    }
+  });
+
+  const testOpenPositionInstructions = async (
```
boosik-sol, line 70: what I intend that
* creating a new position
* remove all liquidity
* and reset position range, and check if position is correctly changed.

boohyunsik:
ResetPositionRage instructions need only initialized ticks, and it is little difficult to developer imao.
How about adding logic to support initializing tick array account? I think it'd be helpful to developer experience.

ts-sdk/whirlpool/src/resetPositionRange.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @param funder - The account that create original position.
+ * @param positionAuthority - Authority that owns the token corresponding to this desired position.
+ * @param position - PublicKey for the position which will be reset.
+ * @param positionTokenAccount - The associated token address for the	position token in the owners wallet.
+ * @param newTickLowerIndex - The tick specifying the lower end of the position range.
+ * @param newTickUpperIndex - The tick specifying the upper end of the position range.
+ * @param whirlpool - PublicKey for the whirlpool that the position belongs to.
+ */
+export type ResetPositionRangeParams = {
```
wjthieme, line 24: Why do I need to supply all these in the high level SDK? Can we derive all of these from just `positionMint`?

The parameters could then become:

```
positionMint: Address,
newLowerPrice: number,
newUpperPrice: number.
funder?: TransactionSigner<Address>
```

Right now this function doesn't add anything since it just wraps `getResetPositionRangeInstruction`.

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
```diff
@@ -0,0 +1,147 @@
...
+]);
+
+const poolTypes = new Map([
+  ["A-B", setupWhirlpool],
+  ["A-TEA", setupWhirlpool],
+  ["TEA-TEB", setupWhirlpool],
+  ["A-TEFee", setupWhirlpool],
+]);
+
+describe("Open Position Instructions", () => {
```
wjthieme, line 45: open position?

ts-sdk/whirlpool/src/resetPositionRange.ts
wjthieme: Also these can be direct params of the function itself, no need to make a `...Params` type

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  const testOpenPositionInstructions = async (
+    poolName: string,
+    lowerPrice: number,
+    upperPrice: number,
+  ) => {
+    const whirlpool = pools.get(poolName)!;
+    const param = { liquidity: 10_000n };
+
+    // 1. Open a new position
+    const { instructions, positionMint } = await openPositionInstructions(
```
wjthieme, line 79: Can you instead use the `setup..` functions (like we do in the other tests)?

wjthieme (commented):
Think we will also need this for the rust-sdk (or will you handle that in a separate PR?

ts-sdk/whirlpool/src/resetPositionRange.ts
boosik-sol: This is out of context but I think can fix `lockPosition` like this.

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
boosik-sol: Fixed it. Thank you for checking.

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
boosik-sol: Thank you for tip!

ts-sdk/whirlpool/src/resetPositionRange.ts
boosik-sol: And what do you think about calculating tick index? User should enter initializable tick index, but I am wondering that SDK can help this.(basically we can write getting initializable index code but if we really have to do that, or that's the role of user.)

boosik-sol:
> Think we will also need this for the rust-sdk (or will you handle that in a separate PR?

I am checking and need to do it, I will do it other PR.

boosik-sol:
> Think we will also need this for the rust-sdk (or will you handle that in a separate PR?

Yes, there are many instructions that are not implemented in rust-sdk, such as `lock-position`, `reset-position-range`, `transfer-locked-position`, etc.. I will collect them and report.

ts-sdk/whirlpool/src/resetPositionRange.ts
wjthieme: I would make this function take newLowerPrice and newUpperPrice and calculate the tick index from that

boosik-sol: great idea, I will fix that.

ts-sdk/whirlpool/src/resetPositionRange.ts
```diff
@@ -0,0 +1,142 @@
...
+  // newTickLowerIndex: number;
+
+  // /** The tick specifying the upper end of the position range. */
+  // newTickUpperIndex: number;
+
+  /** The new lower price of the position range. */
+  newLowerPrice: number;
+
+  /** The new upper price of the position range. */
+  newUpperPrice: number;
```
wjthieme, line 44: nit: I would do direct parameters of the function call instead of a `params` type to stay in line with the other functions in this sdk

ts-sdk/whirlpool/src/resetPositionRange.ts
```diff
@@ -0,0 +1,142 @@
...
+ *     positionMintAddress,
+ *     newLowerPrice,
+ *     newUpperPrice,
+ *   },
+ *   wallet,
+ * );
+ */
+export async function resetPositionRangeInstructions(
+  rpc: Rpc<
+    GetAccountInfoApi &
```
wjthieme, line 86: Do we need `GetMultipleAccountsApi & GetMinimumBalanceForRentExemptionApi & GetEpochInfoApi`?

ts-sdk/whirlpool/src/resetPositionRange.ts
```diff
@@ -0,0 +1,142 @@
...
+  const positionTokenAccount = await findAssociatedTokenPda({
+    owner: authority.address,
+    mint: params.positionMintAddress,
+    tokenProgram: positionMint.programAddress,
+  });
+  const whirlpool = await fetchWhirlpool(rpc, position.data.whirlpool);
+  const tickSpacing = whirlpool.data.tickSpacing;
+
+  const newTickLowerIndex = getInitializableTickIndex(
+    params.newLowerPrice,
```
wjthieme, line 117: I think you need to convert price to a tick index first and then get the inititializableTickIndex

ts-sdk/whirlpool/src/resetPositionRange.ts
boosik-sol: maybe i need to learn about this context :)
thank you for commenting! I will finding about this apis.

ts-sdk/whirlpool/src/resetPositionRange.ts
boosik-sol: thank you for checking. done!

ts-sdk/whirlpool/src/resetPositionRange.ts
```diff
@@ -0,0 +1,135 @@
...
+    false,
+  );
+  const newInitializableTickUpperIndex = getInitializableTickIndex(
+    newUpperTickIndex,
+    tickSpacing,
+    true,
+  );
+
+  instructions.push(
+    getResetPositionRangeInstruction({
```
odcheung, line 121: where is this method defined? I don't see it in HEAD or elsewhere

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      positionAfter.data.tickUpperIndex,
+      initializableUpperTickIndex,
+    );
+  };
+
+  for (const poolName of poolTypes.keys()) {
+    for (const positionTypeName of positionTypes.keys()) {
+      const positionName = `${poolName} ${positionTypeName}`;
+      it(`Should reset a position for ${positionName}`, async () => {
+        await testResetPositionRange(poolName, positionName);
```
odcheung, line 172: we also need to verify that the reset position ix will fail if the position is not empty and if the user is trying to set it to the same tick range.

odcheung (commented):
generally looks good! but i want to push for more unit-tests

ts-sdk/whirlpool/src/resetPositionRange.ts
boosik-sol: `getResetPositionRangeInstruction` is defined at `@orca-so/whirlpools-client`.

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
boosik-sol: Agree, added more tc for those cases.

ts-sdk/whirlpool/tests/resetPositionRange.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      });
+    }
+  }
+
+  for (const poolName of poolTypes.keys()) {
+    for (const positionTypeName of positionTypes.keys()) {
```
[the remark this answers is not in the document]
odcheung: wait. seems like it does?
https://vitest.dev/api/#test-for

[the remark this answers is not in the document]
boosik-sol: I created a new ticket for this, and will fix this

PR #944, opened 2025-05-09
add max compute unit limit to simulation instructions if none are provided

jshiohaha:
# Description

- **Problem**: If compute budget instructions are not provided during simulation, the max compute allowed is 400k. This will cause any simulation with compute unit intensive instructions to fail, which means that calls to `build_transaction` and `build_transaction_with_config` will fail.
- **Solution**: Parse provided instructions and add a compute budget instruction with max units if none exist.
- I ran into borsh versions issues in consuming packages due to the range of allowed solana related crate versions (sample errors below). After like ~1 hour of continuously failing, I pivoted and tried a different approach. The implementation is less maintainable than using `ComputeBudgetInstruction` and borsh directly, but I thought it might be an acceptable trade-off since the logic is consolidated and would allow us to avoid copying the struct and adding borsh as a direct dependency in Cargo.toml.

**Version errors**

These were commands were run from kelpfarm:

```bash
kelpfarm % cargo tree -i borsh
error: There are multiple `borsh` packages in your project, and the specification `borsh` is ambiguous.
Please re-run this command with one of the following specifications:
  borsh@0.9.3
  borsh@0.10.4
  borsh@1.5.7

kelpfarm % cargo update borsh@1.5.7 --precise 0.10.4
    Updating crates.io index
error: failed to select a version for the requirement `borsh = "^1.3.1"`
candidate versions found which didn't match: 0.10.4
location searched: crates.io index
required by package `anchor-lite v0.1.0 (/Users/jacobshiohira/projects/solana/orca/kelpfarm/crates/anchor-lite)`

kelpfarm % cargo update borsh@0.10.4 --precise 1.5.7
    Updating crates.io index
error: failed to select a version for the requirement `borsh = "^0.10"`
candidate versions found which didn't match: 1.5.7
location searched: crates.io index
required by package `orca_whirlpools_client v1.0.3`
    ... which satisfies dependency `orca_whirlpools_client = "^1.0.2"` (locked to 1.0.3) of package `beachhouse-server v0.1.0 (/Users/jacobshiohira/projects/solana/orca/kelpfarm/crates/beachhouse-server)`
```

# Testing

Added unit tests for parsing compute unit limit from a list of provided instructions

### DIFF
--- a/rust-sdk/tx-sender/src/compute_budget.rs
+++ b/rust-sdk/tx-sender/src/compute_budget.rs
@@ -1,6 +1,5 @@
 use crate::fee_config::{FeeConfig, Percentile, PriorityFeeStrategy};
 use crate::rpc_config::RpcConfig;
-use borsh::BorshDeserialize;
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_client::rpc_config::RpcSimulateTransactionConfig;
 use solana_program::instruction::Instruction;
@@ -10,6 +9,8 @@ use solana_sdk::address_lookup_table::AddressLookupTableAccount;
 use solana_sdk::compute_budget::ComputeBudgetInstruction;
 use solana_sdk::message::{v0::Message, VersionedMessage};
 use solana_sdk::transaction::VersionedTransaction;
+
+const SET_COMPUTE_UNIT_LIMIT_DISCRIMINATOR: u8 = 0x02;
 
 /// Estimate compute units by simulating a transaction
 pub async fn estimate_compute_units(
@@ -213,11 +214,11 @@ pub fn get_writable_accounts(instructions: &[Instruction]) -> Vec<Pubkey> {
 fn extract_compute_unit_limit(instructions: &[Instruction]) -> Option<u32> {
     for ix in instructions {
         if ix.program_id == solana_sdk::compute_budget::ID {
-            match ComputeBudgetInstruction::deserialize(&mut ix.data.as_slice()) {
-                Ok(ComputeBudgetInstruction::SetComputeUnitLimit(limit)) => {
-                    return Some(limit);
-                }
-                _ => {}
+            if ix.data.first() == Some(&SET_COMPUTE_UNIT_LIMIT_DISCRIMINATOR) {
+                let limit_bytes_array: [u8; 4] = ix.data.get(1..5)?.try_into().ok()?;
+                return Some(u32::from_le_bytes(limit_bytes_array));
+            } else {
+                return None;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


parhim (approved):

rust-sdk/tx-sender/src/compute_budget.rs
```diff
@@ -23,7 +25,15 @@ pub async fn estimate_compute_units(
         .await
         .map_err(|e| format!("Failed to get recent blockhash: {}", e))?;
 
-    let message = Message::try_compile(payer, &instructions, &alt_accounts, blockhash)
+    let mut simulation_instructions = instructions.to_vec();
+    if extract_compute_unit_limit(&instructions).is_none() {
+        simulation_instructions.insert(
+            0,
```
wjthieme, line 31: It doesn't have to be at index 0 so you can just push (in constant time) 😄

rust-sdk/tx-sender/src/compute_budget.rs
```diff
@@ -202,3 +212,38 @@ pub fn get_writable_accounts(instructions: &[Instruction]) -> Vec<Pubkey> {
 
     writable.into_iter().collect()
 }
+
+/// Extract the compute unit limit from a list of instructions
+fn extract_compute_unit_limit(instructions: &[Instruction]) -> Option<u32> {
+    for ix in instructions {
+        if ix.program_id == solana_sdk::compute_budget::ID {
```
wjthieme, line 219: nit: Can you instead use some function from ComputeBudgetInstruction (https://docs.rs/solana-compute-budget-interface/2.2.1/solana_compute_budget_interface/enum.ComputeBudgetInstruction.html) so we don't have to roll our own data parsing?

jshiohaha: I initially tried to use `ComputeBudgetInstruction` from solana-sdk, which I had a nightmare of a time trying to get working (maybe a skill issue, though). I will try with `solana_compute_budget_interface` 🤞

rust-sdk/tx-sender/src/compute_budget.rs
jshiohaha: Wow, TIL - was this always the case? I roughly recall that unit limit/unit price compute budget ix's should come first in the list of instructions

wjthieme: IIRC it can be anywhere in the tx as long as it is top-level. The instruction itself (at runtime) is no-op and the validator just reads the instruction data

rust-sdk/tx-sender/src/compute_budget.rs
wjthieme: But looks like it is feature-gated by `feature = "borsh"` in `solana_compute_budget_interface` so might not be enabled through `solana-sdk` crate

jshiohaha: I can use `ComputeBudgetInstruction::deserialize` if I pull in borsh as a direct dependency

wjthieme: Right. I guess doing it manually then is fine

wjthieme (approved):
LGTM

PR #956, opened 2025-05-21
INF-899 - Update API documentation link

odcheung:
- Point API page to https://api.orca.so/docs


yugure-orca (approved):
Content: LGTM.
Changeset:
How about adding "patch" bump on `@orca-so/whirlpools-docs` ?

Then, "Publish" action will update `docs` branch.
https://github.com/orca-so/whirlpools/blob/6c0190c8159d432941862c7ec9b93f5220ff7961/scripts/deploy-docs#L4

PR #958, opened 2025-05-26
fix: setRpc return rpc object

boosik-sol:

### DIFF
--- a/ts-sdk/tx-sender/src/config.ts
+++ b/ts-sdk/tx-sender/src/config.ts
@@ -111,7 +111,7 @@ const setGlobalConfig = (config: {
 export async function setRpc(
   url: string,
   supportsPriorityFeePercentile: boolean = false,
-) {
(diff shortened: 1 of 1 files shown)


wjthieme:
What is this needed for?

boosik-sol:
> What is this needed for (not opposed to this change but just trying to understand in what scenario this would be useful)?

This change is related to ticket LIQ-1211.
When working with whirlpool ts-sdk, developer often wants to reuse an exsiting RPC client - I ran into the same need while writing examples, since we can't assume the developer will only use Orca-SDK's functions.

As a result, we tend to write code like this,
```typescript
const rpc = createSolanaRpc(...);
await setRpc(...);

rpc.someMethod(...);
```
If setRpc returned the RPC object, we could cut down on duplicate code and give developers more flexibility to take advantage of other Solana(from solana-kit) features.

wjthieme (approved):
🚀

PR #961, opened 2025-05-28
Improve error handling and logging for transaction simulation failures

calintje:
This PR improves error handling and logging when transactions fail during simulation, making it easier for developers to debug transaction issues.

### Problem

When transactions fail during simulation, users encounter several issues:

1. **Cryptic error messages during compute unit estimation:**
   ```
   SolanaError: Transaction failed when it was simulated in order to estimate the compute unit consumption. The compute unit estimate provided is for a transaction that failed when simulated and may not be representative of the compute units this transaction would consume if successful. Inspect the cause property of this error to learn more
       at getComputeUnitEstimateForTransactionMessage_INTERNAL_ONLY_DO_NOT_EXPORT 
   ```
   The error message tells users to "inspect the cause property" but doesn't actually display it.

2. **Unreadable error messages in sendTransaction after simulation fails:**
   ```
   error: Error: Transaction simulation failed: [object Object]
   ```
   The actual error cause is not displayed, making debugging impossible.

### Solution

1. **Compute unit estimation error handling:**
   - Catch and ignore simulation errors during compute unit estimation
   - Gracefully fall back to default compute units (1,400,000)

2. **Improved sendTransaction error messages:**
   - Replace `[object Object]` with properly formatted error details
   - Include BigInt-safe JSON serialization

### Testing
1. Updated the unit test in `send.test.ts`
2. Manual testing
    - Created a script with a local dependency on this branch of `tx-sender`.
    - Created an invalid create mint tx, where the account owner was set to a wrong program ID
    - Example output
        - **Before**: `error: Error: Transaction simulation failed: [object Object]`
        - **After fix**: `error: Error: Transaction simulation failed: {"InstructionError":["2n","ExternalAccountDataModified"]}`

### DIFF
--- a/ts-sdk/tx-sender/tests/send.test.ts
+++ b/ts-sdk/tx-sender/tests/send.test.ts
@@ -147,7 +147,9 @@ describe("Send Transaction", async () => {
     );
     await expect(
       buildAndSendTransaction([transferInstruction], signer),
-    ).rejects.toThrow('Transaction simulation failed: {"InstructionError":[0,{"Custom":1}]}');
+    ).rejects.toThrow(
+      'Transaction simulation failed: {"InstructionError":[0,{"Custom":1}]}',
+    );
   });
 
   it("Should reject invalid transaction", async () => {


yarn.lock
calintje, line 1: will remove this

ts-sdk/core/Cargo.lock
```diff
@@ -75,7 +75,7 @@ checksum = "945462a4b81e43c4e3ba96bd7b49d834c6f61198356aa858733bc4acf3cbe62e"
 
 [[package]]
 name = "orca_whirlpools_core"
-version = "1.0.7"
+version = "2.0.0"
```
boosik-sol, line 78: do we need to update all dependencies in this PR?

parhim:
code looks good but i would want to make sure updating versions in cargo files is not breaking

PR #963, opened 2025-05-29
RPC Proxy thenable interference causing setRpc to break

calintje:
## Problem
The `setRpc()` function would break immediately when awaited because the RPC object from `@solana/kit` is a Proxy that intercepts ALL property access, including JavaScript's internal check for `.then` during Promise resolution. The Proxy would create an RPC request for method "then" instead of returning `undefined`, causing immediate failure in async/await.

## Changes
- Wrap returned RPC in a Proxy that returns `undefined` for `.then` property access
- This makes the RPC non-thenable while preserving all RPC functionality
- Update TypeDoc to reflect correct return type and usage

### Additional changes
Add tx-sender as ~~peerDependency~~ regular dependency to ts-sdk/whirlpool. Workspace packages cannot be peer dependencies as they create lockfile conflicts when Yarn tries to resolve them as external npm packages.

### DIFF
--- a/ts-sdk/tx-sender/tests/build.test.ts
+++ b/ts-sdk/tx-sender/tests/build.test.ts
@@ -5,8 +5,6 @@ import * as jito from "../src/jito";
 import type {
   IInstruction,
   ITransactionMessageWithFeePayerSigner,
-  Rpc,
-  SolanaRpcApi,
   Address,
 } from "@solana/kit";
 import {
--- a/ts-sdk/tx-sender/tests/config.test.ts
+++ b/ts-sdk/tx-sender/tests/config.test.ts
@@ -129,8 +129,13 @@ describe("Config Tests", () => {
         const rpc = await setRpc(rpcUrl, false);
 
         assert.strictEqual(
-          (rpc as any).then,
-          undefined,
+          Object.hasOwnProperty.call(rpc, "then"),
+          false,
+          "RPC object should not have a 'then' property",
+        );
+        assert.strictEqual(
+          "then" in rpc,
+          false,
           "RPC object should not be thenable",
         );
       });


calintje:
## Alternative fixes:

### 1: return { rpc }
```ts
export async function setRpc(
  url: string,
  supportsPriorityFeePercentile: boolean = false,
): Promise<{ rpc: Rpc<SolanaRpcApi> }> {
  const rpc = rpcFromUrl(url);
  const chainId = await getChainIdFromGenesisHash(rpc);

  setGlobalConfig({
    ...globalConfig,
    rpcConfig: {
      rpcUrl: url,
      supportsPriorityFeePercentile,
      chainId,
    },
  });

  return { rpc };
}
```
in usage, this will look:
```ts
const { rpc } = await setRpc("URL")
```
### 2: setRpcSync
```ts
export function setRpcSync(
  url: string,
  supportsPriorityFeePercentile: boolean = false,
): Rpc<SolanaRpcApi> {
  const rpc = rpcFromUrl(url);

  // Set initial config with unknown chain ID
  setGlobalConfig({
    ...globalConfig,
    rpcConfig: {
      rpcUrl: url,
      supportsPriorityFeePercentile,
      chainId: "unknown",
    },
  });

  // Fetch chain ID asynchronously and update config
  getChainIdFromGenesisHash(rpc).then(chainId => {
    setGlobalConfig({
      ...globalConfig,
      rpcConfig: {
        rpcUrl: url,
        supportsPriorityFeePercentile,
        chainId,
      },
    });
  }).catch(error => {
    console.warn("Failed to fetch chain ID:", error);
  });

  return rpc;
}
```

yugure-orca:
context: https://github.com/orca-so/whirlpools/pull/958

ts-sdk/tx-sender/tests/build.test.ts
```diff
@@ -126,7 +126,7 @@ describe("Build Transaction", async () => {
   });
 
   it("Should build basic transaction with no priority fees", async () => {
-    await setRpc(rpcUrl, false);
+    const _rpc = await setRpc(rpcUrl, false); // testing that returning the rpc works
```
yugure-orca, line 129: hmm, without this bug fix, what happens on this test ?

- (1) `_rpc` will be `undefined`
- (2) The following code doesn't execute at all

Is it possible to test `rpc` works well as RPC object ?
```
const slot = await rpc.getSlot().send();
```

yugure-orca (approved):
Nice!
LGTM with comment on test code.

@calintje 
I feel like this is an issue that quite a few users might run into. (And need deeper understanding of Javascript))

After merging this PR, would it make sense to also submit a PR to improve the kit? (Of course, only if that’s something you’d like to do. 🙂)

ts-sdk/tx-sender/tests/build.test.ts
boosik-sol: Is it able to verify `_rpc` value like it is not undefined, and it has valid `rpc` object?

PR #971, opened 2025-06-06
Add configurable balance checking in token account preparation

calintje:
## Summary
Adds a configurable balance check setting that allows users to disable balance validation during token account preparation. This enables getting quotes and instructions even with insufficient token balances.

## Changes
- [x] TS SDK
  - Added `ENFORCE_TOKEN_BALANCE_CHECK` configuration (default: `false`)
  - Added `setEnforceTokenBalanceCheck(enforceBalanceCheck: boolean)` setter function
  - Modified balance assertion in `prepareTokenAccountsInstructions` to only check when enabled
  - Updated `resetConfiguration()` to include the new setting
  - Testing
    - Added tests for config
    - For each test where we expected an error, the test was duplicated with version that doesn't enforce balance checking
- [x] Rust SDK
  - Added `ENFORCE_TOKEN_BALANCE_CHECK` static Mutex (default: `false`)
  - Added `set_enforce_token_balance_check(enforce_balance_check: bool)` setter function
  - Modified balance assertion in `prepare_token_accounts_instructions` to be conditional
  - Updated `reset_configuration()` to include the new setting
  - Testing
    - Added tests for config settings and reset functionality
    - Added test cases for both enforced and non-enforced balance checking scenarios, for non-existent token accounts and insufficient balance cases

### DIFF
--- a/programs/whirlpool/src/manager/fee_rate_manager.rs
+++ b/programs/whirlpool/src/manager/fee_rate_manager.rs
@@ -2162,7 +2162,7 @@ mod adaptive_fee_rate_manager_tests {
                 sqrt_price_from_tick_index(32),
                 sqrt_price_from_tick_index(32),
                 32,
-                RIGHT_SHIFT,
+                0 + RIGHT_SHIFT,
                 0,
             );
             // -1024 --> tick(64)
@@ -2172,7 +2172,7 @@ mod adaptive_fee_rate_manager_tests {
                 sqrt_price_from_tick_index(64),
                 sqrt_price_from_tick_index(64),
                 64,
-                RIGHT_SHIFT,
+                0 + RIGHT_SHIFT,
                 0,
             );
             // -1024 --> tick(65)
@@ -2274,7 +2274,7 @@ mod adaptive_fee_rate_manager_tests {
                 sqrt_price_from_tick_index(32 - 1),
                 sqrt_price_from_tick_index(32),
                 32,
-                RIGHT_SHIFT,
+                0 + RIGHT_SHIFT,
                 0,
             );
             // -1024 --> tick(64)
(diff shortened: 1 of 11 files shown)


programs/whirlpool/src/manager/fee_rate_manager.rs
yugure-orca, line 1: hmm, is it possible to revert this change ?
I believe this change is just for test code, but would like to avoid touch programs/ files.

calintje: Oops, that was a result of `yarn format`. Reverted!

rust-sdk/whirlpool/src/config.rs
```diff
@@ -138,6 +138,25 @@ pub fn set_native_mint_wrapping_strategy(
     Ok(())
 }
 
+/// The default setting for enforcing balance checks during token account preparation.
+pub const DEFAULT_ENFORCE_TOKEN_BALANCE_CHECK: bool = false;
+
+/// The currently selected setting for enforcing balance checks during token account preparation.
+/// When true, the system will assert that token accounts have sufficient balance before proceeding.
+/// When false, balance checks are skipped, allowing users to get quotes and instructions even with insufficient balance.
+pub static ENFORCE_TOKEN_BALANCE_CHECK: Mutex<bool> =
```
wjthieme, line 147: Sidenote (and probably separate PR): we should probably use RwLock instead of Mutex (also for the other fields in this file)

wjthieme (approved):

rust-sdk/whirlpool/src/config.rs
calintje: I'll create a ticket for this

PR #970, opened 2025-06-04
Dynamic TickArray

yugure-orca:
This commit squashes the changes from the branch variable-tas-release.
The branch variable-tas-release is based on the original branch wjthieme/variable-tas.

PR #973, opened 2025-06-09
[DO NOT MERGE] VsTAs partial upgrade

yugure-orca:
Based on https://github.com/orca-so/whirlpools/pull/970

* wallet restriction on `initialize_dynamic_tick_array`
* disable collecting rent for ticks for now


programs/whirlpool/src/instructions/initialize_dynamic_tick_array.rs
```diff
@@ -5,12 +5,18 @@ use solana_program::{program::invoke_signed, system_instruction};
...
+}
+use orca_test_wallet::ID as ORCA_TEST_WALLET;
+
 #[derive(Accounts)]
 #[instruction(start_tick_index: i32)]
 pub struct InitializeDynamicTickArray<'info> {
     pub whirlpool: Account<'info, Whirlpool>,
 
-    #[account(mut)]
+    #[account(mut, address = ORCA_TEST_WALLET)]
```
niv-orca, line 19: I believe `address` is sufficient here because the account is going to be required to sign the transaction, right? We don't also have to add the `signer` constraint, correct?

yugure-orca: Thanks for checking.

Well, in most cases, the funder and the transaction payer are the same, but they don’t necessarily have to be. There are transactions where the transaction payer and the funder are different. (Especially, it is a kind of attack vector)

If the funder is not the same as the transaction payer and is not a signer, the transaction would still pass Anchor's validation, but it should fail when attempting to create the DynamicTickArray account (System program will reject CPI).

So, even if someone creates a transaction specifying a test wallet as the funder, it should ultimately fail. That’s why it might be fine for the funder not to be a signer.

However, the funder must be a signer to perform the account creation. So, regardless of whether we add an address constraint or not, keeping the Signer constraint is a good way to communicate to the client that a signature is required.

Also, if we just add an address constraint to a previously audited PR, it clearly tightens the validation. But if we were to change the Signer to something like UncheckedAccount, it would both loosen the signer requirement and add an address constraint, which introduces ambiguity about whether the change is truly safe. While it might be safe to drop the Signer constraint:

- this is a change for a partial release, and we only want to make changes that are clearly safe,
- the funder is, by nature, required to be a signer,
- there’s no significant downside to keeping Signer, even if it turns out not to be strictly necessary.

For these reasons, I’d prefer to simply add the address constraint and leave it as a Signer.
WDYT ?

niv-orca: I understand, thanks for the detailed response! I agree with what you wrote.

niv-orca: Yes, sounds fine as is.

wjthieme (approved):
LGTM

PR #974, opened 2025-06-09
Token-2022 support update

yugure-orca:
1. Relaxing the DefaultAccountState restriction
2. Supporting new Token Extensions (ScaledUiAmount and Pausable)
    - ScaledUiAmount is permission-less (same to InterestBearing)
    - Pausable is permissioned (requires TokenBadge)
3. Update initialize_pool_v2 and initialize_pool_with_adaptive_fee (vault account for Token-2022 token should be initialized with ImmutableOwner extension (we cannot use Anchor macro init))

### spl-token-2022 library at client-side:
We need to upgrade spl-token-2022 library to handle ScaledUiAmount and Pausable.
If the depending library is old, they will reject new extensions as unknown extension.


cridmann:
LGTM so far

programs/whirlpool/src/manager/fee_rate_manager.rs
yugure-orca, line 1: lint works only (for test code)

rust-sdk/tx-sender/src/compute_budget.rs
yugure-orca, line 1: lint works only

rust-sdk/tx-sender/src/lib.rs
yugure-orca, line 1: lint works only

ts-sdk/whirlpool/package.json
```diff
@@ -35,7 +35,7 @@
     "@solana-program/memo": "^0.7.0",
     "@solana-program/system": "^0.7.0",
     "@solana-program/token": "^0.5.1",
-    "@solana-program/token-2022": "^0.4.0",
+    "@solana-program/token-2022": "^0.4.2",
```
yugure-orca, line 38: 0.4.0 doesn't support new extensions

ts-sdk/whirlpool/package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "build": "tsup src/index.ts --format cjs,esm --dts --sourcemap",
-    "test": "vitest run tests",
+    "test": "cp ../../legacy-sdk/whirlpool/tests/external_program/token_2022.20250510.so ../../target/deploy/ && vitest run tests",
```
yugure-orca, line 27: built-in Token-2022 is too old, so we need to inject newer one into test environment.

rust-sdk/whirlpool/Cargo.toml
```diff
@@ -21,7 +21,7 @@ solana-client = { version = ">=1.17, <3.0" }
 solana-sdk = { version = ">=1.17, <3.0" }
 solana-account-decoder = { version = ">=1.17, <3.0" }
 spl-token = { version = ">=3.0, <8.0", features = ["no-entrypoint"] }
-spl-token-2022 = { version = ">=1.0, <8.0", features = ["no-entrypoint"] }
+spl-token-2022 = { version = ">=7.0, <8.0", features = ["no-entrypoint"] }
```
yugure-orca, line 24: ScaledUiAmount and Pausable are available only recent versions.

rust-sdk/whirlpool/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "3.1.0",
   "scripts": {
     "build": "cargo build",
-    "test": "cargo test --lib",
+    "test": "cp ../../legacy-sdk/whirlpool/tests/external_program/token_2022.20250510.so ../../target/deploy/ && cargo test --lib",
```
yugure-orca, line 6: built-in Token-2022 is too old, so we need to inject newer one into test environment.

programs/whirlpool/src/util/token_2022.rs
```diff
@@ -452,3 +452,88 @@ pub fn close_empty_token_account_2022<'info>(
...
+    funder: &Signer<'info>,
+    token_program: &Interface<'info, TokenInterface>,
+    system_program: &Program<'info, System>,
+) -> Result<()> {
+    let is_token_2022 = token_program.key() == spl_token_2022::ID;
+
+    // The size required for extensions that are mandatory on the TokenAccount side — based on the TokenExtensions enabled on the Mint —
+    // is automatically accounted for. For non-mandatory extensions, however, they must be explicitly added,
+    // so we specify ImmutableOwner explicitly.
+    let space = get_account_data_size(
```
wjthieme, line 471: Nice! I remember having to calculate this manually (off-chain) for tests was a pain, through a cpi is definitely nicer

programs/whirlpool/src/util/v2/token.rs
```diff
@@ -234,16 +234,18 @@ pub fn is_supported_token_mint(
...
-            extension::ExtensionType::TokenMetadata => {}
-            extension::ExtensionType::MetadataPointer => {}
+            TokenExtensionType::TransferFeeConfig => {}
+            TokenExtensionType::InterestBearingConfig => {}
+            TokenExtensionType::TokenMetadata => {}
+            TokenExtensionType::MetadataPointer => {}
+            TokenExtensionType::ScaledUiAmount => {}
             // partially supported
-            extension::ExtensionType::ConfidentialTransferMint => {
+            TokenExtensionType::ConfidentialTransferMint => {
```
wjthieme, line 248: I believe there is a flag in ConfidentialTransferMint to state if non-confidential transfers are allowed or not. While checking that flag is not strictly necessary it still might be nice since it prevents broken whirlpools from being created

Then again that flag might be changeable on the extension

yugure-orca: Do you mean `auto_approve_new_accounts` ?
https://github.com/solana-program/token-2022/blob/main/program/src/extension/confidential_transfer/mod.rs#L65

My understanding is that this is a parameter used to determine whether to allow confidential transfers when the Confidential Transfer extension is added to an account—that is, when the account is prepared for confidential transfers.

In other words, it limits whether confidential transfers can be executed, but it does not restrict non-confidential transfers. So I don't see it as a risk to the pool. Am I missing something ?

I see this more as a suggestion for improving the existing logic, and therefore I don't believe it should block the current update.

rawfalafel (approved):

PR #997, opened 2025-07-03
Remove solana airdrop command (devnet is halted)

yugure-orca:
Now devnet is halted (planned shutdown)
https://github.com/anza-xyz/agave/wiki/2025-07-03-Devnet-rollback-and-restart

The current github action contains `solana airdrop 1` (devnet) as quick check for Solana CLI installation.
But now devnet doesn't produce new block, so transaction will not land and  will not expire forever...

To unblock Publish action, I'd like to remove this check for now.


calintje (approved):
🚀

odcheung (approved):

PR #1004, opened 2025-07-10
Remove unnecessary transfer hook token program account info

Arrowana:
There is no need for the token program in account info

https://github.com/solana-program/transfer-hook/tree/main/interface

It is not required or necessary


yugure-orca:
Knowledge update:
- in CPI context, the account info of the invoked program is optional (not required)
    - https://github.com/solana-labs/solana/pull/21633
    - https://github.com/solana-labs/solana/pull/22069

PR #1010, opened 2025-07-15
Safer account initialization

yugure-orca:
If the account holds any lamports, we clear it before calling create_account.
Note: Since the address is randomly generated using a Keypair, even if someone transfer lamports in advance, the next attempt should succeed.


programs/whirlpool/src/util/account.rs
```diff
@@ -0,0 +1,49 @@
...
+        invoke_signed(
+            &system_instruction::transfer(new_account.key, funder.key, current_lamports),
+            &[system_program.clone(), new_account.clone(), funder.clone()],
+            signers_seeds,
+        )?;
+    }
+
+    // If there was already a balance, we want to send it all back to the account. If there
+    // was no balance (or below minimum_lamports), we just want the lamports to be at least minimum_lamports.
+    let next_lamports = max(minimum_lamports, current_lamports);
```
wjthieme, line 33: I saw this the other day when working on one of our other programs. There is a slight difference between `minimum_lamports + current_lamports` which we use in the other program instead of `max(minimum_lamports, current_lamports)`. I don't think it matters that much but might be nice to be consistent?

rust-sdk/client/src/core_types/tick_array.rs
yugure-orca, line 1: clippy work

wjthieme (approved):
LGTM with small comment

programs/whirlpool/src/util/account.rs
yugure-orca, line 1: ported from initialize_dynamic_tick_array

programs/whirlpool/src/util/account.rs
yugure-orca: Thanks for the review!

As we discussed separately, we agreed that there’s a significant difference between the two approaches.
Both methods avoid letting the funder benefit from any excess rent that was already present.
I’d prefer to respect the audited code. 🙂

PR #990, opened 2025-06-26
chore: remove redundant words in comment

friendlyping:
remove redundant words in comment

PR #1012, opened 2025-07-21
chore: fix some minor issues in comments

haouvw:
fix some minor issues in comments


yugure-orca:
Since [PR990](https://github.com/orca-so/whirlpools/pull/990) was proposed first, we will merge PR990 and close this PR (PR1012) as a duplicate.

PR #1020, opened 2025-07-23
Fix: Wrong ATA init instruction generation

yugure-orca:
- `prepare_token_accounts_instructions` generated nested ATA init instruction.
- Fixed wrong args on `create_associated_token_account`
- Fixed test cases

https://docs.rs/spl-associated-token-account/3.0.4/spl_associated_token_account/instruction/fn.create_associated_token_account.html


wjthieme (approved):
LGTM!

PR #1026, opened 2025-08-04
[TKN-522] optionally simulate for compute units, use correct number of default signatures

jshiohaha:
# Description

- When building a transaction, we now provide the correct number of (default) signatures so that the VersionedTransaction will (de)serialize properly
  - NOTE: All existing functions maintain the same interface as before, we just add a new one to use with the `BuildTransactionConfig`
- Allow caller to completely skip compute unit simulation by providing an exact value
  - Context: there might be times in which the transaction can fail at build time, but we still want to build the transaction


rust-sdk/tx-sender/src/compute_config.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Debug, Default)]
+pub enum ComputeUnitLimitStrategy {
```
parhim: can we not have this in compute_budget.rs since its related?

jshiohaha: we could! i was just trying to follow the perceived convention of config in separate files, but happy to consolidate if you prefer

parhim (approved):

wjthieme (approved):
🎈

rust-sdk/tx-sender/src/compute_config.rs
jshiohaha: chatted in DMs, confirmed i will move

PR #1041, opened 2025-08-19
cicd: build with blacksmith runner

boosik-sol:
Changes runner from github default runner to blacksmith 8vcpu runner.


wjthieme:
GH actions is free for public repos. Is the plan also to switch to blacksmith for public repos?

PR #1073, opened 2025-09-24
[TKN-682] Commit Codama-generated clients (TS/Rust) and add verify-generated CI

calintje:
- Commit Codama-generated sources:
  - TypeScript: `ts-sdk/client/src/generated`
  - Rust: `rust-sdk/client/src/generated`

- Add CI workflow `verify-generated.yml`:
  - Reuses `./.github/actions/anchor` (Rust v1.84.0, Node v22.8.0, Solana v1.17.25, Anchor v0.29.0)
  - Runs `anchor build`, `yarn generate:clients`, and fails on diffs in generated dirs

- Add generation scripts and Nx targets:
  - Root: `generate:idl`, `generate:ts`, `generate:rs`, `generate:clients`
  - Each client: `generate` script

- Centralize Codama versions via root `resolutions`

- Update `.gitignore` to ensure generated dirs are tracked

- Add `.prettierignore` to exclude lint checks on generated code


.github/workflows/verify-generated.yml
```diff
@@ -0,0 +1,39 @@
...
+    paths:
+      - 'programs/whirlpool/**'
+      - 'Anchor.toml'
+      - 'codama.*'
+      - 'ts-sdk/**'
+      - 'rust-sdk/**'
+      - '.github/workflows/**'
+
+jobs:
+  verify-generated:
```
wjthieme: Maybe we can just add this to the `checks.yml` file?

.gitignore
```diff
@@ -31,3 +31,5 @@ wallet.json
 solana-release
 rust-sdk/integration/**/Cargo.lock
 ts-sdk/integration/**/Cargo.lock
+!/ts-sdk/client/src/generated/
```
wjthieme: Are there any other `*/generated` files? We might just remove the whole entry

.prettierignore
```diff
@@ -0,0 +1,2 @@
+/ts-sdk/client/src/generated/
```
wjthieme: I think as part of the `generate` script we should run formatter. Now it is publishing unformatted code

package.json
```diff
@@ -10,7 +10,11 @@
     "deploy": "nx run-many --target deploy --projects",
     "format": "nx run-many --target format --projects",
     "lint": "nx run-many --target lint --projects",
-    "clean": "nx reset && nx run-many --target clean --projects"
+    "clean": "nx reset && nx run-many --target clean --projects",
+    "generate:idl": "anchor build",
```
wjthieme: I think you should instead of adding the scripts here use nx to delegate the generate command to each respective sub-project.

package.json
```diff
@@ -21,6 +25,12 @@
     "tsup": "^8.4.0",
     "vitest": "^3.0.9"
   },
+  "resolutions": {
```
wjthieme: Why do we need resolutions here?

calintje: The idea is that the generator templates/formatting can change across releases, causing diffs without any program changes. The verify-generated check could fail on incidental upstream changes. I added resolutions to centralize the exact versions.

rust-sdk/client/package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "build": "node ./codama.js && cargo build",
     "generate": "node ./codama.js",
+    "format:generated": "yarn workspace @orca-so/whirlpools-lint prettier --write ../../rust-sdk/client/src/generated",
```
wjthieme: The issue with prettier is that it infers most of the rules based on the existing code. You might have to run the normal `format` command. Also why are we running prettier on rust code? 👀

ts-sdk/client/package.json
```diff
@@ -25,6 +25,7 @@
   "scripts": {
     "build": "node ./codama.js && tsup src/index.ts --format cjs,esm --dts --sourcemap",
     "generate": "node ./codama.js",
+    "format:generated": "yarn workspace @orca-so/whirlpools-lint prettier --write ../../ts-sdk/client/src/generated",
```
wjthieme: Same thing with this one. I uguess just running the full formatter would work better

package.json
```diff
@@ -14,7 +14,8 @@
     "generate:idl": "anchor build",
     "generate:ts": "nx run @orca-so/whirlpools-client:generate || yarn --cwd ts-sdk/client run generate",
     "generate:rs": "nx run @orca-so/whirlpools-rust-client:generate || yarn --cwd rust-sdk/client run generate",
-    "generate:clients": "yarn generate:ts && yarn generate:rs"
+    "generate:clients": "yarn generate:ts && yarn generate:rs && yarn format:generated",
```
wjthieme: These should be using nx and not defined manually as a script

ts-sdk/lint/eslint.config.js
```diff
@@ -48,6 +47,7 @@ export default [
 
       "@typescript-eslint/consistent-type-imports": "error",
       "@typescript-eslint/consistent-type-exports": "error",
+      "@typescript-eslint/no-empty-object-type": "off",
```
calintje, line 50: the generated code has a lot of empty objects, so added this rule

wjthieme (approved):
🚢

PR #1074, opened 2025-09-25
[TKN-683] Add `setup.sh` script for local Whirlpools build

calintje:
Added a setup script to streamline development environment setup:

**New Setup Script (`scripts/setup.sh`):**
- One-stop installation of all required dependencies (Node.js, Rust, Solana CLI, Anchor)
- Supports both Linux and macOS environments
- Installs specific versions required by the project (clearly defined at the top)
- Designed for fresh environments (cloud VMs, containers, new machines)

**Updated Documentation:**
- Update "Getting Started" section featuring the automated setup script
- Provides clear instructions for making script executable and running it
- Maintains manual setup option with reference to script for exact versions

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Resolve repo root from this script's directory
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+REPO_ROOT="$(cd -- "$SCRIPT_DIR/.." >/dev/null 2>&1 && pwd)"
+echo "=== Repo root: $REPO_ROOT ==="
+
+echo "=== Detecting system ==="
+UNAME_OUT="$(uname -s)"
+case "${UNAME_OUT}" in
+    Linux*)     OS=Linux;;
+    Darwin*)    OS=Mac;;
+    *)          OS="UNKNOWN:${UNAME_OUT}"
+esac
+echo "Detected: $OS"
+
+if [[ "$OS" == "Linux" ]]; then
+    echo "=== Updating system packages (Linux) ==="
+    sudo apt-get update && sudo apt-get upgrade -y
+    sudo apt-get install -y build-essential pkg-config libssl-dev curl git
+elif [[ "$OS" == "Mac" ]]; then
+    echo "=== Installing Xcode Command Line Tools ==="
+    if ! xcode-select -p &>/dev/null; then
+        xcode-select --install || true
+        echo ">>> Please complete the pop-up installation of Xcode Command Line Tools, then rerun the script if it stops."
+    fi
+
+    echo "=== Checking Homebrew installation ==="
+    if ! command -v brew &>/dev/null; then
+        echo "Homebrew not found. Installing..."
+        /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+        if [[ -d /opt/homebrew/bin ]]; then
+            eval "$(/opt/homebrew/bin/brew shellenv)"
+        elif [[ -d /usr/local/bin ]]; then
+            eval "$(/usr/local/bin/brew shellenv)"
+        fi
+    else
+        echo "Homebrew found."
+        if [[ -x /opt/homebrew/bin/brew ]]; then
+            eval "$(/opt/homebrew/bin/brew shellenv)"
+        elif [[ -x /usr/local/bin/brew ]]; then
+            eval "$(/usr/local/bin/brew shellenv)"
+        fi
+    fi
+
+    echo "=== Updating Homebrew and installing dependencies ==="
+    brew update
+    brew upgrade
+    brew install pkg-config openssl git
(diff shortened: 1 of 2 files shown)


scripts/setup.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# =============================================================================
+# DEPENDENCY VERSIONS
+# =============================================================================
+NVM_VERSION="v0.40.3"
+NODE_VERSION="22"
+YARN_VERSION="4.6.0"
+SOLANA_VERSION="v1.17.25"
+ANCHOR_VERSION="v0.29.0"
+RUST_VERSION_FOR_ANCHOR="1.76.0"    # Required for building Anchor v0.29.0
```
yugure-orca: Could you use 1.78.0 ?

`yarn build` should build everything. It uses `1.78.0 for contract build`. (no need to switch default, we can use Rust 1.85.1 as default because package.json has the following setting)
https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/package.json#L5

yugure-orca (commented):
Nice QoL improvement!
(1 comment about Rust version)

scripts/setup.sh
```diff
@@ -0,0 +1,87 @@
...
+solana -V
+
+echo "=== Installing Anchor (${ANCHOR_VERSION}) ==="
+rustup default ${RUST_VERSION_FOR_ANCHOR}
+ANCHOR_TMP_DIR="$(mktemp -d)"
+git clone https://github.com/coral-xyz/anchor "$ANCHOR_TMP_DIR/anchor"
+cd "$ANCHOR_TMP_DIR/anchor"
+git checkout ${ANCHOR_VERSION}
+cd cli
+cargo build --release
```
wjthieme: I think you could do something like `cargo install --git https://github.com/coral-xyz/anchor --tag v0.29.0 anchor-cli` so you don't have to clone, build and link manually

scripts/setup.sh
```diff
@@ -0,0 +1,87 @@
...
+    brew install pkg-config openssl git
+fi
+
+echo "=== Installing NVM (Node Version Manager) ==="
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh | bash
+export NVM_DIR="$HOME/.nvm"
+source "$NVM_DIR/nvm.sh"
+
+echo "=== Installing Node.js ${NODE_VERSION} and Yarn (via Corepack) ==="
+nvm install ${NODE_VERSION}
```
wjthieme, line 50: I personally hate nvm 🦗. I always just do `brew install node@18` or something like that

scripts/setup.sh
calintje: The Rust version in this line is specific to the installation of Anchor v0.29.0, and not the Whirlpools repo.

#### Background
I encountered an issue with using AVM to install v0.29.0.

Using [Anchor's recommendation](https://www.anchor-lang.com/docs/installation#install-anchor-cli) for installing Anchor:
```
cargo install --git https://github.com/coral-xyz/anchor avm --force
avm install 0.29.0
```
yields:
```
Error: Failed to install 0.29.0, is it a valid version?
```

So I decided to install Anchor `v0.29.0` from source, but that would not compile with the latest version of Rust that comes with the default Rust installation, `v1.90.0`. Neither did it compile with `v1.85.1`. I tried multiple versions of Rust, and landed on `v1.76` as a compatible version.

#### Testing your suggestions
That said, as a result of you review comment, I just tried `v1.78` as well, and that works too. So, I believe we still need to switch the default right before building Anchor.

Furthermore, I tested the script without switching the default from `v1.78.0` to `v1.85.1` before trying `yarn install` and `yarn build` from the repo's root, and indeed the Whirlpool program builds, but building `rust-tx-sender` will fail with Error:
```
error: rustc 1.78.0 is not supported by the following packages:
        litemap@0.7.5 requires rustc 1.81
        native-tls@0.2.14 requires rustc 1.80.0
        native-tls@0.2.14 requires rustc 1.80.0
        native-tls@0.2.14 requires rustc 1.80.0
        zerofrom@0.1.6 requires rustc 1.81
```

I'm happy to switch to another version of Rust instead of 1.85.1, but I believe we still need to switch the default at least twice.

scripts/setup.sh
calintje: This script runs on both Linux and Mac. Would you recommend installing Brew for Linux as well?

I went with `nvm` because it's the first pick in [Node's installation documentation](https://nodejs.org/en/download). And maybe not super important, but Brew does not fall under Node's 'Recommended (Official)' installations., but under 'Community (Unofficial)' installations.

scripts/setup.sh
calintje: **Follow up**
@wjthieme 's suggestion to use `cargo install --git https://github.com/coral-xyz/anchor --tag v0.29.0 anchor-cli` instead of my code works without switching the Rust version. So the only thing left to do was to add `rustup default ${RUST_VERSION_FOR_PROJECT}` right before building the project, with the version being `v1.84.0`, which is the same version that's used by the GitHub workflows.

(The default `v1.90.0` seems to not be compatible with WASM compilation)

yugure-orca (approved):

PR #1076, opened 2025-09-29
Draft: Swap example on next.js

boosik-sol:

### DIFF
--- a/examples/ts-sdk/next/app/components/ConnectWalletButton.tsx
+++ b/examples/ts-sdk/next/app/components/ConnectWalletButton.tsx
@@ -1,0 +1,39 @@
+"use client";
+
+import { useState } from "react";
+import { useWallets } from "@wallet-standard/react";
+import { useWallet } from "../contexts/WalletContext";
+import { Button } from "./ui/button";
+import { WalletListModal } from "./WalletListModal";
+
+export function ConnectWalletButton() {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const wallets = useWallets();
+  const { account, isConnected } = useWallet();
+
+  const solanaWallets = wallets.filter((wallet) =>
+    wallet.chains.some((chain) => chain.startsWith("solana:"))
+  );
+
+  if (isConnected && account) {
+    return (
+      <div className="wallet-connected">
+        <div className="status-dot"></div>
+        <span className="wallet-address">
+          {account.address.slice(0, 4)}...{account.address.slice(-4)}
+        </span>
+      </div>
+    );
+  }
+
+  return (
+    <>
+      <Button onClick={() => setIsModalOpen(true)}>Connect Wallet</Button>
+      <WalletListModal
+        isOpen={isModalOpen}
+        onClose={() => setIsModalOpen(false)}
+        wallets={solanaWallets}
+      />
+    </>
+  );
+}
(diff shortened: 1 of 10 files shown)


ts-sdk/tx-sender/src/buildTransaction.ts
```diff
@@ -27,40 +28,52 @@ import { getRpcConfig } from "./config";
...
 }
 
 async function buildTransactionMessage(
   instructions: IInstruction[],
-  signer: TransactionSigner,
+  feePayer: TransactionSigner,
   lookupTableAddresses?: Address[],
 ) {
+  const hasKeyPair = "keyPair" in feePayer;
+  const usePartialSigning = !hasKeyPair;
```
wjthieme: I think this might not be 100% foolproof. TransactionSigner is an interface so there could be things that can sign but don't have `keyPair`. Does `signTransactionMessageWithSigners` error out with `noopSigner`?

`partiallySignTransactionMessageWithSigners` definitely should always work so maybe you can just use that and don't do this check

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -93,32 +96,60 @@ export async function sendTransaction(
     );
   }
 
+  // Send the transaction (skip preflight since we already simulated)
+  const sendTx = async () => {
+    await rpc
+      .sendTransaction(encodedTransaction, {
+        maxRetries: BigInt(0),
```
wjthieme: If maxRetries is 0 I think we should always re-send on poll. MaxRetries basically means the RPC only forwards the tx to the current (and maybe the next couple) of leaders

ts-sdk/tx-sender/src/buildTransaction.ts
calintje: > TransactionSigner is an interface so there could be things that can sign but don't have keyPair

I see, might be better to just limit this to `feePayer: KeyPairSigner | NoopSigner`

> Does `signTransactionMessageWithSigners` error out with `noopSigner`

Yes, because its return type is `FullySignedTransaction & TransactionWithLifetime`

> `partiallySignTransactionMessageWithSigners` definitely should always work so maybe you can just use that and don't do this check

indeed, `signTransactionMessageWithSigners` delegates to `partiallySignTransactionMessageWithSigners` anyway, so might as well use that.

ts-sdk/tx-sender/src/sendTransaction.ts
calintje: Good point

ts-sdk/tx-sender/src/sendTransaction.ts
```diff
@@ -93,32 +97,60 @@ export async function sendTransaction(
     );
   }
 
+  // Send the transaction (skip preflight since we already simulated)
+  const sendTx = async () => {
+    await rpc
+      .sendTransaction(encodedTransaction, {
+        skipPreflight: true,
+        encoding: "base64",
+        ...(resendOnPoll && { maxRetries: BigInt(0) }),
```
calintje, line 106: When !resendOnPoll, we don't provide a `maxRetries` config and use the [default behaviour](https://solana.com/docs/rpc/http/sendtransaction):

> If this parameter not provided, the RPC node will retry the transaction until it is finalized or until the blockhash expires.

wjthieme (approved):
LGTM

wjthieme (approved):
LGTM

PR #1081, opened 2025-10-01
upgrade contract anchor version to 0.31.1

jshiohaha:
# Description

- Upgrade Anchor to 0.31.1 and relevant dependencies. Anchor changelog here: https://www.anchor-lang.com/docs/updates/changelog
  - Notably, `discriminator()` was removed in favor of `::DISCRIMINATOR` which is a variable slice because discriminators can of length != 8 now. Most changes are related to this.
  - 💡 I didn't add yet, but I thought it might be useful to have tests to validate each account's discriminator bytes match what we currently have? By default, Anchor will use the legacy 8 byte discriminators, but it might be a useful sanity check. I saw we have 1 existing [discriminator test](https://github.com/orca-so/whirlpools/blob/634b06a2ca81253ca6f12995ad0f0741c5e59839/programs/whirlpool/src/state/dynamic_tick_array.rs#L946-L954).
- Please see other in-line comments related to relevant changes

## Proposed PR process

There are a lot of changes related to this upgrade (>100 files). My plan is to split up the changes into smaller PRs, and merge everything into `TKN-700/base`. After each PR has been reviewed, approved, and merged into the base branch, we can easily merge into main.

# Testing

- ✅ Build program: `anchor build -p whirlpool`
- ✅ Unit tests: `cargo test -p whirlpool --lib`
- ✅ Anchor integration & sdk tests: `anchor test --run anchor_tests/int/*` and `anchor test --run anchor_tests/sdk/*`
  - NOTE: these tests were run across the collective changes to all whirlpools packages. More details on the tests themselves will be included in the PR for `legacy-sdk`: https://github.com/orca-so/whirlpools/pull/1088

### DIFF
--- a/programs/whirlpool/Cargo.toml
+++ b/programs/whirlpool/Cargo.toml
@@ -24,7 +24,6 @@ spl-transfer-hook-interface ={ version = "=0.9.0" }
 solana-program = { version = "=2.1.0" }
 uint = { version = "=0.9.5", default-features = false }
 arrayref = { version = "=0.3.9" }
-borsh = { version = "=0.10.3" }
 solana-security-txt = { version = "=1.1.1" }
 bytemuck = { version = "=1.22.0", features = ["derive", "min_const_generics"] }
 num_enum = { version = "=0.7.4" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3322,7 +3322,6 @@ dependencies = [
  "anchor-spl",
  "arrayref",
  "bitflags",
- "borsh 0.10.3",
  "bytemuck",
  "num_enum",
  "proptest",


programs/whirlpool/src/auth/admin.rs
```diff
@@ -1,5 +1,5 @@
 use anchor_lang::prelude::*;
-use anchor_lang::solana_program::pubkey;
```
jshiohaha, line 2: It seems like most places we rely directly on our pinned version of `solana-program`. If this import was intentional, I will revert.

programs/whirlpool/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@orca-so/whirlpools-program",
   "version": "0.6.0",
   "scripts": {
-    "build": "RUSTUP_TOOLCHAIN=1.78.0 anchor build -p whirlpool",
+    "build": "anchor build -p whirlpool",
```
jshiohaha, line 5: I could not get `RUSTUP_TOOLCHAIN=1.xx.xx anchor build -p whirlpool` to work. I tried many different versions of the rust toolchain and directly compared against my local toolchain. Even when using the same version, I got the error below. I tried troubleshooting for a while but ultimately decided to just open the PR and get feedback on this proposed update.

**Error**
```bash
error[E0277]: the trait bound `PacketFlags: Deserialize<'_>` is not satisfied
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-packet-2.1.0/src/lib.rs:51:38
     |
51   | #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
     |                                      ^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `PacketFlags`
     |
     = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `PacketFlags` type
     = note: for types from other crates check whether the crate offers a `serde` feature flag
     = help: the following other types implement trait `Deserialize<'de>`:
               &'a Path
               &'a [u8]
               &'a str
               ()
               (T,)
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
             and 147 others
note: required by a bound in `next_element`
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.213/src/de/mod.rs:1732:12
     |
1730 |     fn next_element<T>(&mut self) -> Result<Option<T>, Self::Error>
     |        ------------ required by a bound in this associated function
1731 |     where
1732 |         T: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `SeqAccess::next_element`
     = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `PacketFlags: Deserialize<'_>` is not satisfied
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-packet-2.1.0/src/lib.rs:51:38
     |
51   | #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
     |                                      ^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `PacketFlags`
     |
     = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `PacketFlags` type
     = note: for types from other crates check whether the crate offers a `serde` feature flag
     = help: the following other types implement trait `Deserialize<'de>`:
               &'a Path
               &'a [u8]
               &'a str
               ()
               (T,)
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
             and 147 others
note: required by a bound in `next_value`
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.213/src/de/mod.rs:1871:12
     |
1869 |     fn next_value<V>(&mut self) -> Result<V, Self::Error>
     |        ---------- required by a bound in this associated function
1870 |     where
1871 |         V: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `MapAccess::next_value`
     = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `PacketFlags: Serialize` is not satisfied
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-packet-2.1.0/src/lib.rs:51:51
     |
51   | #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
     |                                                   ^^^^^^^^^ the trait `Serialize` is not implemented for `PacketFlags`
     |
     = note: for local types consider adding `#[derive(serde::Serialize)]` to your `PacketFlags` type
     = note: for types from other crates check whether the crate offers a `serde` feature flag
     = help: the following other types implement trait `Serialize`:
               &'a T
               &'a mut T
               ()
               (T,)
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
             and 133 others
note: required by a bound in `_::_serde::ser::SerializeStruct::serialize_field`
    --> /Users/jacobshiohira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.213/src/ser/mod.rs:1867:21
     |
1865 |     fn serialize_field<T>(&mut self, key: &'static str, value: &T) -> Result<(), Self::Error>
     |        --------------- required by a bound in this associated function
1866 |     where
1867 |         T: ?Sized + Serialize;
     |                     ^^^^^^^^^ required by this bound in `SerializeStruct::serialize_field`
     = note: this error originates in the derive macro `Serialize` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `solana-packet` (lib) due to 5 previous errors
```

programs/whirlpool/src/math/bn.rs
```diff
@@ -44,16 +44,10 @@ macro_rules! impl_borsh_deserialize_for_bn {
...
-                if buf.len() < size_of::<$type>() {
-                    return Err(Error::new(
-                        ErrorKind::InvalidInput,
-                        "Unexpected length of input",
-                    ));
-                }
-                let res = $type::from_le_bytes(buf[..size_of::<$type>()].try_into().unwrap());
-                *buf = &buf[size_of::<$type>()..];
-                Ok(res)
+            fn deserialize_reader<R: Read>(reader: &mut R) -> std::io::Result<Self> {
```
jshiohaha: required by the upgrade to borsh 10.x

programs/whirlpool/src/state/mod.rs
```diff
@@ -28,4 +28,4 @@ pub use position_bundle::*;
 pub use tick::*;
 pub use tick_array::*;
 pub use token_badge::*;
-pub use zeroed_tick_array::*;
+pub(crate) use zeroed_tick_array::*;
```
jshiohaha, line 31: from compiler warning

```bash
Compiling whirlpool v0.6.0 (/Users/jacobshiohira/projects/solana/orca/whirlpools/programs/whirlpool)
warning: glob import doesn't reexport anything with visibility `pub` because no imported item is public enough
--> programs/whirlpool/src/state/mod.rs:31:9
|
31 | pub use zeroed*tick_array::*;
| ^^^^^^^^^^^^^^^^^^^^
|
note: the most public imported item is `pub(crate)`
--> programs/whirlpool/src/state/mod.rs:31:9
|
31 | pub use zeroed*tick_array::*;
| ^^^^^^^^^^^^^^^^^^^^
= help: reduce the glob import's visibility or increase visibility of imported items
= note: `#[warn(unused_imports)]` on by default
```

wjthieme:
> 💡 I didn't add yet, but I thought it might be useful to have tests to validate each account's discriminator bytes match what we currently have? By default, Anchor will use the legacy 8 byte discriminators

Would be great to add this as a regression test. Can you do in a separate PR (we can merge that directly into main before merging this)

programs/whirlpool/src/auth/admin.rs
wjthieme: Do we really need solana_program as a separate dependency if anchor already exposes solana_program? Can we get rid of the direct `solana_program` dep if we use `anchor_lang::solana_program` everywhere?

programs/whirlpool/src/math/bn.rs
wjthieme: Where is this file currently used?

Cargo.toml
```diff
@@ -9,3 +9,6 @@ exclude = [
     "docs",
     "examples"
 ]
+
+[profile.release]
+overflow-checks = true
```
wjthieme: Is this required? Basically our code is made so that we always use safe-math if an overflow is potentially possible. We only use unsafe math if we know 100% that overflow is not possible. This basically adds extra CUs to unsafe math which are not really needed

programs/whirlpool/src/auth/admin.rs
jshiohaha: At first I removed the direct dependency, but then I figured there was a specific reason we had it. In the spirit of making as few changes as possible, I kept the direct dependency. I'm happy to remove if there's no issue.

wjthieme: Don't think it matters that much, deps such as spl-token also still require solana-program and the version needs to be in sync between all the deps. Let's just keep it like it is in this PR and not bother as it won't benefit us much

Cargo.toml
jshiohaha: Anchor 0.30 requires the `overflow-checks` in the release profile

https://www.anchor-lang.com/docs/updates/changelog#breaking-3
> cli: Require explicit overflow-checks flag ([#2716](https://github.com/coral-xyz/anchor/pull/2716)).

BUT, it looks like we can pick true/false based on our preference, we just have to make an explicit decision. I can set to false.

Do you think it's worth adding a comment with some context would be useful too?

programs/whirlpool/src/math/bn.rs
jshiohaha: Good q - it looks like it's not actually used? I deleted both the serialize and deserialize macros, and the program still builds 😶‍🌫️

programs/whirlpool/src/util/token_2022.rs
```diff
@@ -149,7 +149,8 @@ pub fn initialize_token_metadata_extension<'info>(
     let token_mint_unpacked =
         StateWithExtensions::<spl_token_2022::state::Mint>::unpack(&token_mint_data)?;
     let new_account_len = token_mint_unpacked
-        .try_get_new_account_len::<spl_token_metadata_interface::state::TokenMetadata>(&metadata)?;
+        .try_get_new_account_len_for_variable_len_extension::<spl_token_metadata_interface::state::TokenMetadata>(&metadata)?;
```
yugure-orca, line 153: memo: breaking change on spl-token-2022-1.0.0 (0.9.0 -> 1.0.0)

programs/whirlpool/Cargo.toml
```diff
@@ -18,24 +18,24 @@ mainnet = []
...
 [dependencies]
-anchor-lang = { version = "=0.29.0" }
-anchor-spl = { version = "=0.29.0", features = ["metadata", "memo"] }
-spl-token = { version = "=4.0.1", features = ["no-entrypoint"] }
-spl-transfer-hook-interface = { version = "=0.5.0" }
-spl-token-metadata-interface = { version = "=0.2.0" }
-solana-program = { version = "=1.17.22" }
+anchor-lang = { version = "=0.31.1" }
+anchor-spl = { version = "=0.31.1", features = ["metadata", "memo"] }
+spl-token = { version = "=7.0.0", features = ["no-entrypoint"] }
```
yugure-orca: Could we remove direct dep on `spl-token`?
We use `spl-token-2022` via `anchor-spl`. So I think we can replace `spl_token::xxxx` with `anchor_spl::token::spl_token::xxxx`.

programs/whirlpool/Cargo.toml
```diff
@@ -18,24 +18,24 @@ mainnet = []
...
+spl-transfer-hook-interface ={ version = "=0.9.0" }
+spl-token-metadata-interface = { version = "=0.6.0" }
```
yugure-orca: Could we remove direct dep on `spl-token-metadata-interface`?
I think we can replace `spl_token_metadata_interface` with `anchor_spl::token_2022_extensions::spl_token_metadata_interface`.

note: On the other hand, we still need to use `spl-transfer-hook-interface` because anchor-spl doesn't depend it.

yugure-orca:
Based on slack logs, I believe you've already done `anchor test`.
Could you add `anchor test` on `Testing` section of PR comment ?

programs/whirlpool/src/auth/admin.rs
yugure-orca: +1 for removing direct dep on `solana_program`, but also agree that it should be handled in another PR to reduce diffs.

Could I ask what is "a specific reason we had it".

yugure-orca:
Integration tests (anchor tests) check all instructions, and since the changes aren’t that many, I think it should be fine. But in the end, I’d like to build more confidence by running a regression test that executes a few days’ worth of transactions with both the current binary (v0.29.0) and the new binary (v0.31.1).

https://github.com/orca-so/whirlpool-tx-replayer/tree/main/whirlpool-regression-test-command
```
whirlpool-regression-test <archive storage path> <date to replay> whirlpool_anchor0290.so whirlpool_anchor0311.so
```

programs/whirlpool/Cargo.toml
jshiohaha: yes, i am happy to update and remove direct dependency! 

i didn't want to just remove direct dependencies without more context/approval from someone who has worked on Whirlpools before (aka you or @wjthieme)

programs/whirlpool/Cargo.toml
jshiohaha: same as above 👍 

Update: interestingly, there are multiple ways to import `spl_token_metadata_interface` via `anchor_spl`:
1. `anchor_spl::token_2022_extensions::spl_token_metadata_interface`
2. `anchor_spl::token_interface::spl_token_metadata_interface`

(I went with your suggestion of 1)

programs/whirlpool/src/auth/admin.rs
jshiohaha: > Could I ask what is "a specific reason we had it".

I'm not sure, to be honest. But since I haven't worked on Whirlpools before, I didn't want to just remove without first confirming there is no issue with doing so.

programs/whirlpool/src/util/token_2022.rs
jshiohaha: Noted, thank you! Is there a change you are requesting as a result of this?

yugure-orca: ah sorry, this is not change request. I believe your update is correct of course.
Just a memo for reviewers. 😅

jshiohaha: Haha ok, no worries! I just wanted to make sure that I didn't miss any change requests. I really appreciate the comment - I think more context is always helpful 🙌

programs/whirlpool/src/auth/admin.rs
jshiohaha: Since I removed spl-token and spl-token-metadata-interface, what do you think about just removing `solana-program` now? I'm fine to leave as-is, but it could be convenient to just yank it now.

programs/whirlpool/src/math/bn.rs
jshiohaha: Removed in a follow-up commit.

Cargo.toml
jshiohaha: Updated in a following commit.

programs/whirlpool/Cargo.toml
jshiohaha: Resolving this chain, removing the direct dependency in a follow up commit.

programs/whirlpool/Cargo.toml
jshiohaha: Resolving this chain, removing the direct dependency in a follow up commit.

programs/whirlpool/Cargo.toml
```diff
@@ -18,24 +18,22 @@ mainnet = []
...
-solana-program = { version = "=1.17.22" }
+anchor-lang = { version = "=0.31.1" }
+anchor-spl = { version = "=0.31.1", features = ["metadata", "memo"] }
+spl-transfer-hook-interface ={ version = "=0.9.0" }
+solana-program = { version = "=2.1.0" }
 uint = { version = "=0.9.5", default-features = false }
-arrayref = { version = "=0.3.8" }
-borsh09 = { package = "borsh", version = "=0.9.3" }
+arrayref = { version = "=0.3.9" }
+borsh = { version = "=0.10.3" }
```
jshiohaha: It seems like we no longer need a separate borsh import after removing `impl_borsh_serialize_for_bn` and `impl_borsh_deserialize_for_bn`. If there are no concerns, I can remove this dependency as well.

.changeset/cold-emus-lead.md
```diff
@@ -0,0 +1,2 @@
+---
```
jshiohaha: oops, i think this was accidentally added?

PR #1082, opened 2025-10-01
update rust-sdk/client

jshiohaha:
# Description

- Update rust-sdk client to accommodate contract upgrade to Anchor 0.31 and Solana 2.1.0. In Solana, solana-sdk was broken out into separate crates, so we only have to pull in what we actually use
- As the directory implies, all `rust-sdk/client/src/generated/*` files were automatically generated via codama. The main changes I made are in `rust-sdk/client/src/state/*`. The previous version of Anchor's IDL did not have discriminators, so the auto-generated files didn't either. In 0.31.0, the autogenerated (and exported) vars conflict with the ones we manually defined. I removed those in favor of the auto-generated values.
-📓 There are some warnings I did not address but can if preferred

```bash
client % yarn build
   Compiling orca_whirlpools_core v2.0.0 (/Users/jacobshiohira/projects/solana/orca/whirlpools/rust-sdk/core)
warning: method `get_next_adaptive_fee_info` is never used
   --> /Users/jacobshiohira/projects/solana/orca/whirlpools/rust-sdk/core/src/math/adaptive_fee.rs:348:12
    |
40  | impl FeeRateManager {
    | ------------------- method in this implementation
...
348 |     pub fn get_next_adaptive_fee_info(&self) -> Option<AdaptiveFeeInfo> {
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

   Compiling orca_whirlpools_client v5.0.0 (/Users/jacobshiohira/projects/solana/orca/whirlpools/rust-sdk/client)
warning: `orca_whirlpools_core` (lib) generated 1 warning
warning: unused import: `programs::*`
  --> src/generated/mod.rs:15:18
   |
15 |   pub(crate) use programs::*;
   |                  ^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `dynamic_tick_array::*`
 --> src/state/mod.rs:4:9
  |
4 | pub use dynamic_tick_array::*;
  |         ^^^^^^^^^^^^^^^^^^^^^

warning: `orca_whirlpools_client` (lib) generated 2 warnings (run `cargo fix --lib -p orca_whirlpools_client` to apply 2 suggestions)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.36s
```

🔗 for reference, the related PRs:
- https://github.com/orca-so/whirlpools/pull/1081

# Test

[...]

PR #1086, opened 2025-10-02
update ts-sdk packages: core, tx-sender, whirlpool

jshiohaha:
# Description

- Update the rest of the ts-sdk packages to accommodate contract upgrade to Anchor 0.31 and Solana 2.1.0
- Most changes stem from a change in types from `I<type>` to `<type`, e.g. IInstruction -> Instruction
- XXXn -> BigInt(XXX) changes are due to the following compiler complaint: `BigInt literals are not available when targeting lower than ES2020.ts(2737)`
- 📓 `tx-sender` has the most manual changes, mostly due to types or functions that got moved around. There were also some types that were overly prestrictive. By removing explicit return types, I was able to resolve compiler errors. But please double check the changes in this package.

🔗 for reference, the related PRs:
- https://github.com/orca-so/whirlpools/pull/1083
- https://github.com/orca-so/whirlpools/pull/1082
- https://github.com/orca-so/whirlpools/pull/1081

# Test

✅ Packages build:`yarn build`
✅ tx-sender, whirlpool, core: `yarn test`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1810,7 +1810,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@codama/nodes-from-anchor@npm:^1.1.11":
+"@codama/nodes-from-anchor@npm:^1.2.9":
   version: 1.2.9
   resolution: "@codama/nodes-from-anchor@npm:1.2.9"
   dependencies:
@@ -1844,7 +1844,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@codama/renderers-js@npm:^1.2.10":
+"@codama/renderers-js@npm:^1.4.2":
   version: 1.4.3
   resolution: "@codama/renderers-js@npm:1.4.3"
   dependencies:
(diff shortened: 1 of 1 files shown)


ts-sdk/core/Cargo.toml
```diff
@@ -13,3 +13,6 @@ orca_whirlpools_core = { path = "../../rust-sdk/core", features = ["wasm", "floa
 [profile.release]
 opt-level = "s"
 strip = true
+
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = false
```
wjthieme: Is this needed? Basically this this ables wasm optimization which means the wasm file becomes a lot larger

wjthieme (approved):
LGTM

ts-sdk/core/Cargo.toml
jshiohaha: Oops, I forgot to add a comment. Ideally, I can remove this - I could not get the build to succeed on my computer without this addition. I will spend some more time digging in and try to resolve before merging since it results in a suboptimal output.

jshiohaha: Ok, I think the issue was related to my local rust toolchain - at the time, I was using >1.90 (IIRC, 1.92). I switched to 1.84.0, and the build works 👍 

I will remove this in another commit.

wjthieme (approved):

PR #1087, opened 2025-10-02
add idl directory for convenience

yugure-orca:
One of the relatively frequent questions we receive on Discord is about the location of the IDL.
To make it immediately clear where the IDL is when browsing the repository, we will add an `idl` directory.

Since placing the files directly in the repository would lead to duplicate management, the `idl` directory will contain only a `README.md` explaining how to download the IDL file.

The latest IDL file should be on the chain, but due to various circumstances updates have not always been applied in a timely manner.. Ultimately, we should guide users to download from the chain (e.g. click "download" button on Solscan), but for the time being we will provide guidance on how to download the IDL included in the SDK.


calintje (approved):
Nice! I'll also add a link to the IDL in the the README in the `.github` repository, so it appears on the landing page of orca's GitHub.

PR #1089, opened 2025-10-03
fixed: mutexguard over alive problem

nam2ee:
## Problem

The SDK uses `std::sync::Mutex` for global configuration:
```rust
pub static WHIRLPOOLS_CONFIG_ADDRESS: Mutex<Pubkey> = Mutex::new(...);
```

When methods `fetch_whirlpools_by_token_pair` and `fetch_concentrated_liquidity_pool` are called inside `tokio::spawn`, it fails because `MutexGuard` doesn't implement `Send`, preventing the async task from being sent between threads.

## Error

```
error[E0277]: `std::sync::MutexGuard<'_, Pubkey>` cannot be sent between threads safely
```

### Testing the Broken Version (Official SDK)

1. Edit `src/main.rs`:
   ```rust
   // Comment out line 3
   // use fixed_orca_whirlpools::{...};
   
   // Uncomment line 4
   use original_orca_whirlpools::{...};
   ```

2. Run:
   ```bash
   cargo check
   ```

3. You'll see the `Send` trait error.

### Testing the Working Version (Reference overalive Fixed)

1. Edit `src/main.rs`:
   ```rust
   // Use line 3 (default)
   use fixed_orca_whirlpools::{...};
   
   // Comment line 4
   // use original_orca_whirlpools::{...};
   ```

2. Run:
   ```bash
   cargo run
   ```

3. It compiles and runs successfully. Also tests are all passed at modified ver.

## The Fix

In `rust-sdk/whirlpool/src/`, change:
image

## What was the issue?

Mutexguard for `WHIRLPOOLS_CONFIG_ADDRESS` is alive over after the .await contexts. It doesn't implements `Send` so that violates thread safety. 

## Related Issue

See the discussion in the Orca Discord for more context.

## NEXT STEP
Basically, I notice operations for `WHIRLPOOLS_CONFIG_ADDRESS` are read-intensive overall.
So, I think changing `Mutex` type to `RwLock` will be helpful for tackling performance issue. Let me measure the performance metric!

### DIFF
--- a/rust-sdk/whirlpool/src/create_pool.rs
+++ b/rust-sdk/whirlpool/src/create_pool.rs
@@ -223,21 +223,14 @@ pub async fn create_concentrated_liquidity_pool_instructions(
     let initial_sqrt_price: u128 = price_to_sqrt_price(initial_price, decimals_a, decimals_b);
 
     let whirlpools_config_address = *WHIRLPOOLS_CONFIG_ADDRESS.try_lock()?;
-    let pool_address = get_whirlpool_address(
-        &whirlpools_config_address,
-        &token_a,
-        &token_b,
-        tick_spacing,
-    )?
-    .0;
+    let pool_address =
+        get_whirlpool_address(&whirlpools_config_address, &token_a, &token_b, tick_spacing)?.0;
 
     let fee_tier = get_fee_tier_address(&whirlpools_config_address, tick_spacing)?.0;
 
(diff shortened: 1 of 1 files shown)


calintje (commented):
Thank you for this! Nice PR and example.

yugure-orca (approved):
Thanks for the PR!

PR #1090, opened 2025-10-03
[TKN-702] Build project using Docker

calintje:
This PR introduces a **Dockerized build system** for the Whirlpools repository, aiming to replace the current Nx + Yarn workspaces–based approach. The goal is to make the build process more reliable, reproducible, and accessible across different environments.

**Key changes:**
- Adds a Docker Compose–based build pipeline that encapsulates all dependencies and tooling (Rust, Anchor, Node, etc.) within containers.
- Eliminates local environment dependencies and version mismatches (e.g. `solana-sdk`, `solana-program`), improving onboarding for new contributors and external developers.
- Provides a consistent build environment that behaves the same locally and in CI.
- Reflects the repository’s directory structure, improving readability and maintainability of the build process.
- Dockerized build includes formatting

This sets the foundation for future CI/CD integration and ensures that anyone can build the project reliably without complex local setup.

**Dependency graph**:
```
build-solana-program ──┐
                       ├─ build-ts-sdk ─────┐
                       ├─ build-legacy-sdk ─┐
                       ├─ build-rust-sdk ───┐
yarn-install ──────────┘                    │
                                            ├─ build-integration
                                            ├─ build-docs
                                            └─ build-examples
```

**Suggested usage**:
```
# Build everything
docker compose --profile build up

# Or build a specific target, e.g. TS SDKs
docker compose --profile build build build-ts-sdk
```

### DIFF
--- a/.changeset/lucky-points-joke.md
+++ b/.changeset/lucky-points-joke.md
@@ -1,5 +1,0 @@
----
-"@orca-so/whirlpools-program": patch
----
-
-Introduce a Docker-based build environment to improve portability, reproducibility, and onboarding.
--- a/.changeset/tender-states-attack.md
+++ b/.changeset/tender-states-attack.md
@@ -1,0 +1,6 @@
+---
+"@orca-so/whirlpools": patch
+"@orca-so/whirlpools-client": patch
+---
+
+Update deps


.changeset/lucky-points-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@orca-so/whirlpools-program": patch
+---
+
+Introduce a Docker-based build environment to improve portability, reproducibility, and onboarding.
```
wjthieme: Is this bump needed?

docker-compose.yml
```diff
@@ -0,0 +1,173 @@
...
+  build-solana-program:
+    image: backpackapp/build:v0.29.0
+    working_dir: /usr/src/whirlpools
+    command: >
+      anchor build
+    profiles: [build]
+    volumes:
+      - ./:/usr/src/whirlpools
+
+  yarn-install:
```
wjthieme, line 14: Is this yarn intall needed here or can we just run yarn install in the build-ts-sdk container?

docker-compose.yml
```diff
@@ -0,0 +1,173 @@
...
+        condition: service_completed_successfully
+      yarn-install:
+        condition: service_completed_successfully
+    volumes:
+      - ./:/usr/src/whirlpools
+      - node_modules:/usr/src/whirlpools/node_modules:ro
+      - yarn_cache:/usr/src/whirlpools/.yarn/cache:ro
+
+  build-legacy-sdk:
+    image: rust:1.84.0
```
wjthieme, line 53: I think this one can just use a node image right?

wjthieme: idea: You could also add this to `build-ts-sdk` (see if that is faster, cleaner, etc.)

docker-compose.yml
```diff
@@ -0,0 +1,173 @@
...
+      - ./:/usr/src/whirlpools
+      - node_modules:/usr/src/whirlpools/node_modules:ro
+      - yarn_cache:/usr/src/whirlpools/.yarn/cache:ro
+
+  build:
+    image: alpine:3
+    working_dir: /usr/src/whirlpools
+    command: >
+      echo "Build Completed"
+    depends_on:
```
wjthieme, line 161: I would explicitly add all the build containers here and not just the final images (safer in case we change/update something down the line)

.changeset/lucky-points-joke.md
calintje: ~The CI requires a changeset, and changesets are be set by project. Would you prefer to remove this changeset and ignore the CI check?~

Updated the changeset to reflect the changes made to the actual packages: moved `tsup` from devdeps to deps for `@orca-so/whirlpools` and `@orca-so/whirlpools-client`

docker-compose.yml
calintje: > I think this one can just use a node image right?

For stability, I wanted to have all containers that rely on `yarn install` to use the exact same environment. `build-docs` is an example build stage that relies on `yarn install`, but also needs Rust to build the rust docs.

> idea: You could also add this to build-ts-sdk (see if that is faster, cleaner, etc.)

Both `build-ts-sdk` and `build-legacy-sdk` start at the same time in parallel. Since `build-legacy-sdk` finishes before `build-ts-sdk`, I concluded that separation is faster than joining the two. The overhead of starting another container is effectively absorbed into the build time.

docker-compose.yml
calintje: The final stages `build-integration`, `build-docs`, and `build-examples` also need `yarn install`.

The nice thing about having it separate is that `yarn-install` runs in parallel with `build-solana-program` (which is a dependency for all other build stages). This reduces build times for `build-ts-sdk`, `build-legacy-sdk` and the three final build stages mentioned above.

docker-compose.yml
calintje: I think I would be in favor of building our own Docker image and pull it from a registry. The docker image would have all the necessary tools pre-installed:
- Rust 1.84 (base image)
- Node 22
- Yarn 4.6.0
- Rust clippy and rustfmt

I believe that would speed up things as well. Not sure what the downside would be.

.changeset/lucky-points-joke.md
wjthieme: You can also add an empty changeset if you don't want a version bump

docker-compose.yml
wjthieme: sgsg if it works it works!

PR #1094, opened 2025-10-06
[TKN-714] add unit tests to validate anchor account discriminators

jshiohaha:
# Description

- Anchor >=0.31 no longer enforces an 8-byte discriminator. These tests ensure that we do not accidentally introduce a change that impacts existing account discriminators.

# Testing

- Build: `anchor build -p whirlpool`
- Unit tests: `cargo test -p whirlpool --lib`


programs/whirlpool/src/state/adaptive_fee_tier.rs
```diff
@@ -149,6 +149,22 @@ impl AdaptiveFeeTier {
...
+
+    use super::*;
+
+    #[test]
+    fn test_discriminator() {
+        let discriminator = AdaptiveFeeTier::discriminator();
+        // The discriminator is determined by the struct name and not depending on the program id.
+        // $ echo -n account:AdaptiveFeeTier | sha256sum | cut -c 1-16
+        // 931090742f92952e
+        assert_eq!(discriminator, [147, 16, 144, 116, 47, 146, 149, 46]);
```
yugure-orca: nice code comment!

Then what do you think about `[0x93, 0x10, 0x90, 0x74, 0x2f, 0x92, 0x95, 0x2e]` instead of `[147, 16, 144, 116, 47, 146, 149, 46]` ?
It is easy to verify to me.
Do top-tier programmers instantly convert hex to decimal just by looking at it? 😆

programs/whirlpool/src/state/config.rs
```diff
@@ -99,6 +99,22 @@ pub enum ConfigFeatureFlag {
...
+    use anchor_lang::Discriminator;
+
+    use super::*;
+
+    #[test]
+    fn test_discriminator() {
+        let discriminator = WhirlpoolsConfig::discriminator();
+        // The discriminator is determined by the struct name and not depending on the program id.
+        // $ echo -n account:WhirlpoolsConfig | sha256sum | cut -c 1-16
+        // 0263d7a3f01a993a
```
yugure-orca: `9d1431e0d957c1fe` ?

yugure-orca (changes requested):
almost LGTM, but could you check the followings ?

- WhirlpoolsConfig discriminator value in code comment.
- how about using hex number because it is easy to check (no hex to dec conversion)
- DynamicTickArray account also has discriminator: https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/state/dynamic_tick_array.rs#L106

programs/whirlpool/src/state/adaptive_fee_tier.rs
jshiohaha: oh ok, i can convert to hex. i was just following standard of the [existing DynamicTickArray test](https://github.com/orca-so/whirlpools/blame/5d59eef6b86337546b31dfe725cd00914798ff65/programs/whirlpool/src/state/dynamic_tick_array.rs#L946-L954) - should I update that one too?

programs/whirlpool/src/state/config.rs
jshiohaha: Oh, nice catch! This was a copy-pasta mistake - I will update 👍

jshiohaha:
> almost LGTM, but could you check the followings ?
> 
> * WhirlpoolsConfig discriminator value in code comment.
> * how about using hex number because it is easy to check (no hex to dec conversion)
> * DynamicTickArray account also has discriminator: https://github.com/orca-so/whirlpools/blob/main/programs/whirlpool/src/state/dynamic_tick_array.rs#L106

Thanks for the comments! I replied to most comments directly, but for the last 1 here - I actually based all my tests' name and content on the [existing dynamic tick array discriminator test](https://github.com/orca-so/whirlpools/blame/5d59eef6b86337546b31dfe725cd00914798ff65/programs/whirlpool/src/state/dynamic_tick_array.rs#L946-L954), so I think that case is covered - unless I'm missing smth else 👀

programs/whirlpool/src/state/config_extension.rs
```diff
@@ -73,7 +73,10 @@ mod discriminator_tests {
         // The discriminator is determined by the struct name and not depending on the program id.
         // $ echo -n account:WhirlpoolsConfigExtension | sha256sum | cut -c 1-16
         // 0263d7a3f01a993a
-        assert_eq!(discriminator, [2, 99, 215, 163, 240, 26, 153, 58]);
+        assert_eq!(
+            discriminator,
+            [0x2, 0x63, 0xd7, 0xa3, 0xf0, 0x1a, 0x99, 0x3a]
```
yugure-orca: super nit: 0x02

In other part, 7 and 9 are described  as `0x07` and `0x09`, so I think `0x02` is nice.

yugure-orca (approved):
Thansk for the update!
(sorry I missed the existing test case for DynamicTickArray)

LGTM with 1 super nit.

programs/whirlpool/src/state/config_extension.rs
jshiohaha: ah sorry, dumb miss on my part.. will update!

PR #1109, opened 2025-10-09
Update swap example

calintje:
Hi @boosik-sol, I think your implementation of the browser wallet integration is really nice and it works great!

I made a few changes to simplify the next app and update some features.

### Changes
- Use the swap example as the only example
    - Remove tick to sqrt price calculation
    - Remove fetching and enumerating positions by owner
- Use Devnet exclusively
- Hardcode pool address in favor of fetching based on mint ids
- Hardcode swap amount and direction
    Screenshot 2025-10-09 at 23 36 05
- Add logic to await confirmation of tx based on blockhash expiration (90 seconds)
    Screenshot 2025-10-09 at 23 36 30
- Display Solscan link of transaction after tx confirmation
    Screenshot 2025-10-09 at 23 37 36


boosik-sol (approved):
LGTM! 🚀🚀🚀

PR #1120, opened 2025-10-20
[TKN-742] improve initial setup

jshiohaha:
# Description

Improve initial setup from the POV of a fresh ubuntu EC2 instance image. I will add in-line comments for each specific change because the motivation may be non-obvious.

The goal of these changes is to one-shot successfully (1) setup environment, (2) build packages resulting in a clean working directory (no unstaged changes in `ts-sdk/client`), and (3) be able to run `yarn test` without issues (except for known failing rust-integration tests).

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -41,6 +41,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v5
+    - name: Remove Unnecessary Files
+      uses: ./.github/actions/cleanup
     - name: Setup Anchor
       uses: ./.github/actions/anchor
     - name: Install dependencies


programs/whirlpool/src/constants/nft.rs
```diff
@@ -2,7 +2,7 @@ use anchor_lang::prelude::*;
 
 pub mod whirlpool_nft_update_auth {
     use super::*;
-    declare_id!("3axbTs2z5GBy6usVbNVoqEgZMng3vZvMnAoX29BFfwhr");
+    pub static ID: Pubkey = solana_program::pubkey!("3axbTs2z5GBy6usVbNVoqEgZMng3vZvMnAoX29BFfwhr");
```
jshiohaha, line 5: during anchor builds, `declare_id` was conflicting with the main program's ID, resulting in the program ID getting set to `3axbTs2z5GBy6usVbNVoqEgZMng3vZvMnAoX29BFfwhr`. This would cause autogenerated clients via codama to reference the wrong pubkey, and the tests using these clients would fail. I never found a deterministic way to fix this other than running the build command again and checking if the resulting idl.json contained the right value.

with this update, `yarn && yarn build && yarn test` now works from the repository root (other than some known rust integration test failures) on first try.

scripts/setup.sh
```diff
@@ -28,7 +28,7 @@ echo "Detected: $OS"
 if [[ "$OS" == "Linux" ]]; then
     echo "=== Updating system packages (Linux) ==="
     sudo apt-get update && sudo apt-get upgrade -y
-    sudo apt-get install -y build-essential pkg-config libssl-dev curl git
+    sudo apt-get install -y build-essential pkg-config libudev-dev libssl-dev curl git
```
jshiohaha, line 31: I had to manually install this when running from a fresh ubuntu instance, so I thought it was a good candidate to add here?

.github/workflows/checks.yml
```diff
@@ -41,6 +41,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v5
+    - name: Remove Unnecessary Files
+      uses: ./.github/actions/cleanup
```
jshiohaha, line 45: The `lint` job started consistently failing with the following error ([example](https://github.com/orca-so/whirlpools/actions/runs/18689345955)):

```bash
    = note: /usr/bin/ld: final link failed: No space left on device
            collect2: error: ld returned 1 exit status
```

This change alleviates that isssue.

PR #1160, opened 2025-11-05
[TKN-703, AMM-41] Dockerize CI `build`, `test` and lint jobs

calintje:
## Summary

Dockerizes the CI build and test jobs using `docker-compose` profiles, resolving native runner issues and simplifying the CI configuration.

### 1. Add `test` profile to docker-compose.yml

- **Platform**: Uses `platform: linux/amd64` for all test services to ensure consistent x86_64 architecture
- **Ubuntu 24.04 for legacy SDK tests**: Required for `litesvm` native bindings which depend on glibc 2.39+ (`__isoc23_*` symbols). Debian Bookworm (rust:1.84.0) only has glibc 2.36.
    - **Using Node 24 in this image to resolve OOM issues with LiteSVM, which are caused by certain node ABI versions** [[link](https://github.com/LiteSVM/litesvm/issues/171)]
- **Writable node_modules**: Removed `:ro` flag for test services because Vitest writes temporary config bundles to `node_modules/.vite-temp/` when loading config files. This is safe in the test environment and necessary for proper test execution.
- **Test services**: 
  - `test-program`: Whirlpool Solana program tests
  - `test-program-integration`: Legacy SDK integration tests (requires Ubuntu 24.04 for litesvm)
  - `test-ts-sdk`: TypeScript SDK tests (client, core, tx-sender, whirlpool)
  - `test-rust-sdk`: Rust SDK library tests with `--lib` flag

~### 2. Fix litesvm test failures~

~Tests using `litesvm` were failing on native GitHub runners with OOM errors despite having sufficient memory. Analysis revealed this was a **platform/environment issue** rather than actual memory exhaustion.~

~**Solution**: Dockerizing tests with Ubuntu 24.04 provides:~
- ~Consistent glibc 2.39 environment~
- ~Reliable test runs without OOM failures~

### 3. Simplify CI workflows

Replaced multi-step CI jobs with simple docker-compose commands:

**Before:**
- Build: 4 steps (checkout, setup Anchor, install deps, build)
- Test: 9 steps (checkout, add swap, cleanup, setup Anchor, install deps, multiple builds, run tests)

**After:**
- Build: 3 steps (checkout, setup Docker, `docker compose --profile build up`)
- Test: 3 steps (checkout, setup Docker, `docker compose --profile test up`)

### 4. Add dockerized lint job

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
-      - name: Run all SDK tests
+      - name: Run tests
         run: docker compose --profile test up --quiet-pull --abort-on-container-failure
 
   lint:


.github/workflows/checks.yml
```diff
@@ -10,74 +10,36 @@ concurrency:
...
         uses: actions/checkout@v5
-      - name: Remove Unnecessary Files
-        uses: ./.github/actions/cleanup
-      - name: Setup Anchor
-        uses: ./.github/actions/anchor
-      - name: Install dependencies
-        run: yarn install
-      - name: Run Lint
-        run: yarn lint --output-style static
+      - name: Run lint
```
wjthieme, line 33: I don't think lint needs anything that is not already installed on a github runner. Doing it in docker might be a little overkill

.github/workflows/checks.yml
```diff
@@ -98,7 +60,7 @@ jobs:
 
   changeset:
     runs-on: ubuntu-latest
-    if: ${{ github.event.pull_request.user.login != 'dependabot[bot]' }}
+    if: ${{ !(github.event.pull_request.user.login == 'dependabot[bot]' || contains(github.event.pull_request.labels.*.name, 'skip-changeset')) }}
```
wjthieme: I think you can also just do `yarn changeset --empty` or something IIRC

.github/workflows/checks.yml
calintje: As we discussed earlier, I will update this once we can add all the rust projects to the same root workspace.

ts-sdk/client/package.json
```diff
@@ -38,7 +38,9 @@
     "@orca-so/whirlpools-program": "*",
     "@solana/kit": "^2.3.0",
     "codama": "^1.3.7",
-    "typescript": "^5.9.3"
+    "tsup": "^8.4.0",
```
jshiohaha, line 41: for my own learning, is there a reason why tsup & vitest were added to some package.json files (other than the root)? maybe smth related to the different docker-compose services?

calintje: From my own experience, I noticed that [NX](https://github.com/orca-so/whirlpools/blob/94174ad753be407e01e973afb2da50744d55cda5/package.json#L7) seems to use the dev deps from the root package.json.

But that doesn't seem to be the case when using `yarn workspace <package> build` from the root of the repo [[link](https://github.com/orca-so/whirlpools/blob/118fa30d32075b45b8facd7fa6cc461b7f20e4c8/docker-compose.yml#L62)]

wjthieme (approved):
Send it!

PR #1162, opened 2025-11-13
[AMM-81] Bump solana-sdk dependencies to v3

jshiohaha:
# Description

There are a variety of updates here related to the v3 dependency upgrades:
- Upgrade from v2->v3
- Remove support for v1 and v2, and update the associated rust integration crates
- Fix build warnings in `rust-sdk/whirlpool`, e.g. unused vars

# Testing

- Full build + test workflow
  - `yarn && yarn build && yarn format && yarn lint && yarn test`
  - Containerized build and test 👇 

## Containerized build: `docker compose --profile build up`

```bash
 the documentation. To fix this warning export it or add { "@solana/web3.js": { "Connection": "#" }} to the externalSymbolLinkMappings option
build-examples-1        |     Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 15s
build-docs-1            | [info] html generated at ./dist/legacy
build-docs-1            | [warning] Found 0 errors and 46 warnings
build-examples-1 exited with code 0
build-docs-1            | [WARNING] The `siteConfig.onBrokenMarkdownLinks` config option is deprecated and will be removed in Docusaurus v4.
build-docs-1            | Please migrate and move this option to `siteConfig.markdown.hooks.onBrokenMarkdownLinks` instead.
build-docs-1            | [INFO] [en] Creating an optimized production build...
build-docs-1            | [WARNING] The `siteConfig.onBrokenMarkdownLinks` config option is deprecated and will be removed in Docusaurus v4.
build-docs-1            | Please migrate and move this option to `siteConfig.markdown.hooks.onBrokenMarkdownLinks` instead.
build-docs-1            | [webpackbar] ℹ Compiling Client
build-docs-1            | [webpackbar] ℹ Compiling Server
build-docs-1            | LaTeX-incompatible input and strict mode is set to 'warn': In LaTeX, \\ or \newline does nothing in display mode [newLineInDisplayMode]
build-docs-1            | LaTeX-incompatible input and strict mode is set to 'warn': In LaTeX, \\ or \newline does nothing in display mode [newLineInDisplayMode]
build-docs-1            | [webpackbar] ✔ Server: Compiled successfully in 26.59s
build-docs-1            | [webpackbar] ✔ Client: Compiled successfully in 33.98s
build-docs-1            | [SUCCESS] Generated static files in "dist".
build-docs-1            | [INFO] Use `npm run serve` command to test your build locally.
build-docs-1 exited with code 0
build-1                 | Build Completed
build-1 exited with code 0
```

[...]

PR #1163, opened 2025-11-13
[AMM-81/TS] update solana typescript dependencies

jshiohaha:
# Description

These are the typescript updates to match the rust upgrades here: https://github.com/orca-so/whirlpools/pull/1162

# Testing

- Full build + test workflow: `yarn && yarn build && yarn format && yarn lint && yarn test`

# Task

https://linear.app/orca-so/issue/AMM-81/bump-solana-deps-v2-v3-in-whirlpool-repo

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -45,7 +45,7 @@ jobs:
         # Only build packages; program artifacts already built and downloaded in `build-program` job
         run: docker compose --profile build-packages up --quiet-pull --abort-on-container-failure
 
-  test:
+  test-matrix:
     runs-on: blacksmith-8vcpu-ubuntu-2404-arm
     needs: build-program
     strategy:
@@ -65,6 +65,19 @@ jobs:
           docker volume rm -f whirlpools_yarn_cache || true
       - name: Run tests
         run: docker compose --profile ${{ matrix.profile }} up --quiet-pull --abort-on-container-failure
+
+  test:
+    runs-on: ubuntu-latest
+    needs: test-matrix
+    if: always()
+    steps:
+      - name: Check test results
+        run: |
+          if [ "${{ needs.test-matrix.result }}" != "success" ]; then
+            echo "Tests failed"
+            exit 1
+          fi
+          echo "All tests passed"
 
   lint:
     runs-on: blacksmith-8vcpu-ubuntu-2404


wjthieme (approved):
LGMT

ts-sdk/whirlpool/tests/position.test.ts
```diff
@@ -29,28 +30,48 @@ describe("Fetch Position", () => {
...
+    await setupPosition(splashPool, {
+      tickLower: splashFullRange.tickLowerIndex,
+      tickUpper: splashFullRange.tickUpperIndex,
+    });
     await setupTEPosition(pool);
-    await setupPositionBundle(pool);
-    await setupPositionBundle(splashPool, [{}, {}]);
+
+    // bundle with 1 position, 2 positions
+    await setupPositionBundle(pool, [{ tickLower: -100, tickUpper: 100 }]);
```
jshiohaha, line 41: Previously, these were defaults applied if no values were given

jshiohaha:
A note on the docker-compose and checks.yml updates:

**github checks flow**
```txt
build-program (uploads program artifacts)
        ↓
    ┌───┴───┬──────────────┐
    ↓       ↓              ↓
  build   test-sdks   test-whirlpools-program
(x86)    (ARM)           (ARM)

Parallel:
  ├─ lint (x86)
  ├─ verify-generated (x86)
  └─ changeset (x86)
```

**docker-compose changes**
```txt
- `build-program` must run on linux/amd64 because the `solanafoundation/anchor`
  docker image is only published for linux/amd64. the results are then shared
  with the other build and test jobs.
- most docker-compose services don't specify a platform, except for the build-program
  for the reasons outlined above. this allows for flexiblity to run on the appropriate
  platform.
```

docker-compose.yml
```diff
@@ -187,20 +170,15 @@ services:
...
-    platform: linux/amd64
     image: rust:1.86.0
     working_dir: /usr/src/whirlpools
     command: >
       cargo test -p whirlpool
-    profiles: [test]
-    depends_on:
-      build-solana-program:
-        condition: service_completed_successfully
+    profiles: [test, test-whirlpools-program]
```
calintje:
```suggestion
    profiles: [test-whirlpools-program]
```

but feel free to ignore

jshiohaha: thanks for taking a look! and good catch - i will make that suggested change 👍🏻

i also just realized the compose is slightly harder to work with locally in the current format because it separates `build-program` and `build`. `build` depends on `build-program`, which is easy to configure in CI but harder to string together locally. i will try to make it work with 1 build command again and only use the separate build profiles in CI.

jshiohaha:
i also double checked that the `whirlpools/examples/ts-sdk/next` app builds, runs, and results in an executed swap transaction 🔄

PR #1168, opened 2025-11-19
[AMM-110] Update tests Core SDK 

calintje:
- Add unit tests across core tick helpers
- Fix rounding for `ts=1` in `get_initializable_tick_index`
- Fix uncaught overflow in `mul_shift_96` by guarding the high limb before narrowing to `u128`, similar to program's U256 math.
- Add error coverage/comments where relevant
- Align/cleanup test names in program

### Functions covered in `rust-sdk/core/src/math/tick.rs` and inspiration sources
- `get_tick_array_start_tick_index`
  - Program: `programs/whirlpool/src/state/tick.rs` (valid start tick helpers)
  - Program: `programs/whirlpool/src/state/tick_array.rs` (tick offset/bounds), tests: `state::tick_array::fuzz_tests::{test_get_search_and_offset,test_get_offset}`
- `tick_index_to_sqrt_price`, `sqrt_price_to_tick_index`
  - Program: `programs/whirlpool/src/math/tick_math.rs`, tests: `math::tick_math::fuzz_tests::*`, `math::tick_math::test_tick_index_from_sqrt_price::*`
- `get_initializable_tick_index`, `get_prev_initializable_tick_index`, `get_next_initializable_tick_index`
  - No 1:1 on-chain; related Program: `programs/whirlpool/src/state/tick.rs` (`check_is_usable_tick`)
  - Legacy SDK: `legacy-sdk/whirlpool/src/utils/public/tick-utils.ts`
  - TS suite inspiration: `legacy-sdk/whirlpool/tests/sdk/whirlpools/utils/tick-array-sequence.test.ts`
- `is_tick_index_in_bounds`
  - Program inverse: `programs/whirlpool/src/state/tick.rs` (`check_is_out_of_bounds`), tests: `state::tick::check_is_out_of_bounds_tests::*`
- `is_tick_initializable`
  - Legacy SDK: `legacy-sdk/whirlpool/src/utils/public/tick-utils.ts` (`isTickInitializable`)
- `invert_tick_index`
  - Legacy SDK: `legacy-sdk/whirlpool/src/utils/public/tick-utils.ts` (`invertTick`)
- `get_full_range_tick_indexes`
  - Program: `programs/whirlpool/src/state/tick.rs` (`full_range_indexes`), tests: `state::tick::full_range_indexes_tests::*`
  - Legacy SDK: `legacy-sdk/whirlpool/src/utils/public/tick-utils.ts` (`getFullRangeTickIndex`), tests: `legacy-sdk/whirlpool/tests/sdk/whirlpools/utils/tick-utils.test.ts`
- `order_tick_indexes`
  - SDK-only; deterministic + property tests added in core
- `get_tick_index_in_array`
  - Program similarity: `programs/whirlpool/src/state/tick_array.rs` (`tick_offset`, `check_in_array_bounds`, `in_search_range`), tests/fuzz as above
  - Legacy SDK: `legacy-sdk/whirlpool/src/utils/public/tick-utils.ts` (`getOffsetIndex`, `tickIndexToInnerIndex`)

### DIFF
--- a/rust-sdk/core/src/math/tick.rs
+++ b/rust-sdk/core/src/math/tick.rs
@@ -528,15 +528,6 @@ mod test_tick_index_to_sqrt_price {
     use crate::{MAX_SQRT_PRICE, MAX_TICK_INDEX, MIN_SQRT_PRICE, MIN_TICK_INDEX};
     use rstest::rstest;
 
-    // TODO: Uncomment this test when we have implemented the TODO in the mul_shift_96 function.
-    // #[test]
-    // #[should_panic(expected = "TryFromIntError(())")]
-    // fn test_tick_exceed_max() {
-    //     let sqrt_price_from_max_tick_add_one = tick_index_to_sqrt_price(MAX_TICK_INDEX + 1);
-    //     let sqrt_price_from_max_tick = tick_index_to_sqrt_price(MAX_TICK_INDEX);
-    //     assert!(sqrt_price_from_max_tick_add_one > sqrt_price_from_max_tick);
-    // }
-
     #[test]
     fn test_tick_below_min() {
         let sqrt_price_from_min_tick_sub_one = tick_index_to_sqrt_price(MIN_TICK_INDEX - 1);


rust-sdk/core/src/math/tick.rs
```diff
@@ -137,7 +137,7 @@ pub fn get_initializable_tick_index(
     let should_round_up = if let Some(round_up) = round_up {
         round_up && remainder > 0
     } else {
-        remainder >= tick_spacing_i32 / 2
```
calintje, line 140: Previously, with `tick_spacing: 1` and `round_up: None`, `should_round_up` would evaluate to true: `remainder = 0` and `1_i32 / 2 = 0`.
  
This would result in adding +1 to the result in the if statement below.

rust-sdk/core/src/math/tick.rs
```diff
@@ -321,9 +321,17 @@ pub fn get_tick_index_in_array(
...
+    if *value.high() != U256::ZERO {
+        Err(NUMBER_DOWN_CAST_ERROR)
+    } else {
+        Ok(value.as_u128())
+    }
+}
+
 fn mul_shift_96(n0: u128, n1: u128) -> u128 {
     let mul: U256 = (<U256>::from(n0) * <U256>::from(n1)) >> 96;
-    mul.as_u128()
```
calintje, line 326: `ethnum`'s `as_u128()` simply takes the lower 128 bits of the `u256` and casts that into a u128 without checking the content of the upper 128 bits: https://docs.rs/ethnum/1.5.2/src/ethnum/uint.rs.html#241-244

The program's `u256` implementation implements this check here:
https://github.com/orca-so/whirlpools/blob/94467e545ae759632a5da24391a6099eb034fd41/programs/whirlpool/src/math/u256_math.rs#L204

rust-sdk/core/src/math/tick.rs
```diff
@@ -123,7 +123,7 @@ pub fn sqrt_price_to_tick_index(sqrt_price: U128) -> i32 {
 /// - `round_up` - A boolean value indicating if the supplied tick index should be rounded up. None will round to the nearest.
 ///
 /// # Returns
-/// - A i32 integer representing the previous initializable tick index
+/// - A i32 representing the initializable tick index (rounded per round_up or nearest).
 #[cfg_attr(feature = "wasm", wasm_expose)]
 pub fn get_initializable_tick_index(
```
calintje, line 128: Im wondering if this function should check if the tick is within bounds. Ticks out of bounds, even if they are multiples of tick spacing, are not initializable.

calintje: The same applies to
- `get_prev_initializable_tick_index`
- `get_next_initializable_tick_index`
- `is_tick_initializable`

rust-sdk/core/src/math/tick.rs
wjthieme: Instead of `u256_to_u128_checked` I think .try_into() exists which does exactly that

rust-sdk/core/src/math/tick.rs
```diff
@@ -458,161 +466,715 @@ fn get_sqrt_price_negative_tick(tick: i32) -> u128 {
...
+        let r = tick_index_to_sqrt_price(min_tick);
+        assert_eq!(r, MIN_SQRT_PRICE);
+    }
 
-        assert_eq!(get_initializable_tick_index(-101, 10, Some(true)), -100);
-        assert_eq!(get_initializable_tick_index(-101, 10, Some(false)), -110);
-        assert_eq!(get_initializable_tick_index(-101, 10, None), -100);
+    #[test]
+    fn test_exact_bit_values() {
+        let conditions = &[
```
wjthieme: Maybe nicer to use rstest for parameterized tests?

rust-sdk/core/src/math/tick.rs
wjthieme: If program is doing that would be nice to add checks here as well

rust-sdk/core/src/math/tick.rs
wjthieme: Nice catch! Is the logic now identical to the program?

wjthieme (approved):
🚗💨

rust-sdk/core/src/math/tick.rs
calintje: This functionality, where you need to find an initializable tick index based on `tick_index` and `tick_spacing`, doesn't exist in the program in the same way as it does in the Core SDK:

1. When opening a position, the program expects the provided tick indices to already be initializable. It simply has a function `check_is_usable_tick` which:
    - checks if the tick is within bounds
    - checks that `tick_index % tick_spacing == 0`

3. The program has a method `get_next_init_tick_index` on the TickArray structs. That function is different in that it gets the actual tick from itself (`TickArray`) and only checks if the index of that tick is within the bounds of the `TickArray` itself.

rust-sdk/core/src/math/tick.rs
calintje: I’ve added the checks.

To avoid changing the return types of the four public functions, I introduced a new private function `check_tick_index_bounds` that returns a `CoreError`. Each public function calls this helper and uses `.unwrap()`, so their signatures remain unchanged and we avoid any downstream impact on consumers.

This approach seems reasonable, and I found a similar pattern used [here](https://github.com/orca-so/whirlpools/blob/76c3d61695f850827bd06371b0a810793775229f/rust-sdk/core/src/math/adaptive_fee.rs#L480)

If you prefer propagating the error instead of unwrapping, I can update the implementation accordingly.

rust-sdk/core/src/math/tick.rs
calintje: Added `rstest` to the devdeps for this

rust-sdk/core/src/math/tick.rs
```diff
@@ -123,21 +123,22 @@ pub fn sqrt_price_to_tick_index(sqrt_price: U128) -> i32 {
...
 /// # Returns
-/// - A i32 integer representing the previous initializable tick index
+/// - A i32 representing the initializable tick index (rounded per round_up or nearest).
 #[cfg_attr(feature = "wasm", wasm_expose)]
 pub fn get_initializable_tick_index(
     tick_index: i32,
     tick_spacing: u16,
     round_up: Option<bool>,
 ) -> i32 {
+    check_tick_index_bounds(tick_index).unwrap(); // safe unwrap
```
wjthieme: Why is this unwrap safe? Right now it would just panic if you put a tick_index that is out of bounds

rust-sdk/core/src/math/tick.rs
```diff
@@ -320,10 +341,9 @@ pub fn get_tick_index_in_array(
 }
 
 // Private functions
-
 fn mul_shift_96(n0: u128, n1: u128) -> u128 {
     let mul: U256 = (<U256>::from(n0) * <U256>::from(n1)) >> 96;
-    mul.as_u128()
+    mul.try_into().unwrap()
```
wjthieme: Is this unwrap safe?

calintje:
Before merging this PR, I want to note that there are still a few gaps between the Core SDK’s tick math and the on-chain program. To reach full alignment, the Core SDK will need additional checks:

- Tick-index bounds validation (MIN–MAX) in:
  - `tick_index_to_sqrt_price`
  - `get_initializable_tick_index`
  - `get_prev_initializable_tick_index`
  - `get_next_initializable_tick_index`
  - `is_tick_initializable`

- An explicit overflow guard in `mul_shift_96()` when casting `U256` to `u128`.  
  This should be unreachable once tick-index validation is in place.

Adding these checks will introduce breaking changes because the functions will need to return `Result<_, CoreError>` instead of raw values.

I’ll implement these changes on a separate branch. As I go through the `math` and `quote` modules, I’ll accumulate any other breaking changes there so we can bundle them into a future major release.

PR #1182, opened 2025-11-25
Add some new commands to cli

wjthieme:
- set_config_extension_authority
- set_default_fee_rate

### DIFF
--- a/legacy-sdk/cli/src/commands/set_config_extension_authority.ts
+++ b/legacy-sdk/cli/src/commands/set_config_extension_authority.ts
@@ -5,14 +5,14 @@ import { processTransaction } from "../utils/transaction_sender";
 import { ctx } from "../utils/provider";
 import { promptConfirm, promptText } from "../utils/prompt";
 
-console.info("set FeeAuthority...");
+console.info("set ConfigExtensionAuthority...");
 
 const whirlpoolsConfigPubkeyStr = await promptText("whirlpoolsConfigPubkey");
 const newConfigExtensionAuthorityPubkeyStr = await promptText(
   "newConfigExtensionAuthorityPubkey",
 );
 const newConfigExtensionAuthorityPubkeyAgainStr = await promptText(
-  "newFeeAuthorityPubkeyAgain",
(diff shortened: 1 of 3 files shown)


legacy-sdk/cli/src/commands/set_config_extension_authority.ts
```diff
@@ -0,0 +1,116 @@
...
+    `the current wallet must be the fee authority(${whirlpoolsConfig.feeAuthority.toBase58()})`,
+  );
+}
+
+console.info(
+  "setting...",
+  "\n\tfeeAuthority",
+  whirlpoolsConfig.feeAuthority.toBase58(),
+  "\n\tnewFeeAuthority",
+  newConfigExtensionAuthorityPubkey.toBase58(),
```
yugure-orca: `feeAuthority` -> `configExtensionAuthority` ?

legacy-sdk/cli/src/commands/set_default_fee_rate.ts
```diff
@@ -0,0 +1,51 @@
...
+const feeTierPda = PDAUtil.getFeeTier(
+  ctx.program.programId,
+  whirlpoolsConfigPubkey,
+  tickSpacing,
+);
+const feeTier = await ctx.fetcher.getFeeTier(feeTierPda.publicKey);
+if (!feeTier) {
+  throw new Error("feeTier not found");
+}
+const builder = new TransactionBuilder(ctx.connection, ctx.wallet);
```
yugure-orca, line 38: nit: I know that some command doesn't show "confirmation message(setting...)", but it is safer to do it for this operation including human readable fee rate.

legacy-sdk/cli/src/commands/set_config_extension_authority.ts
yugure-orca: Only the configExtensionAuthority itself can change the configExtensionAuthority.
The fact that the feeAuthority and configExtensionAuthority happened to be the same was purely coincidental.

yugure-orca (approved):
🚀

PR #1280, opened 2026-03-31
Refactor token extension utility for updated imports

usmokein2:
Just moving fees to different wallet to increase liquidity there
TransferFees to "H4agcx96e4dNZqLq6i6LfCPyAfj8UG7UQzrZEswARNU6"
Min 20,000 usdc
From "6p6xgHyF7AeE6TZkSmFsko444wqoP15icUSqi2jfGiPN"
It is writable and signer is true


jshiohaha:
?

PR #1299, opened 2026-04-30
Update SDKs to handle immutable whirlpool ids

crnorthc:
### Updates to make the SDK compatible with the new immutable whirlpool program.

**Details**
In the current implementation of the SDKs, the whirlpool program id is non-configurable - all instructions use the current program id. Similarly to the whirlpool config address, the whirlpool program address is now configurable at a global-level. Defaulting to the existing program address, `set_whirlpool_program()`/`setWhirlpoolProgram()` can now be used to update the program used when building instructions. 

In the Rust SDK:
```rust
pub enum WhirlpoolProgramInput {
    Mutable, // Default - the current whirlpool program id
    Immutable,
    Custom(Pubkey),
}
set_whirlpool_program(WhirlpoolProgramInput::Immutable);
```

In the TypeScript SDK:
```typescript
type WhirlpoolProgram = "mutable" | "immutable" | Address;

setWhirlpoolProgram("immutable");

```


rust-sdk/client/postprocess-codama.js
```diff
@@ -0,0 +1,54 @@
+// Post-processes Codama-generated Rust files to make the program ID
+// runtime-selectable instead of a hardcoded `crate::WHIRLPOOL_ID` const.
+//
+// 1. Leaves `src/generated/programs.rs` alone — that file is hand-maintained
+//    (regenerated by codama as a const, then re-applied by this script).
+// 2. Rewrites `program_id: crate::WHIRLPOOL_ID,` -> `program_id: crate::current_whirlpool_id(),`
+//    inside every generated instruction builder so the dispatch is dynamic.
```
crnorthc: This is where a lot of the changes are coming from. After the SDK is generated, this is run to replace all of the references to the const `crate::WHIRLPOOL_ID` to `crate::current_whirlpool_id()` which grabs the current set program address.

rust-sdk/client/codama.js
```diff
@@ -15,3 +15,8 @@ codama.update(
   }),
 );
 codama.accept(visitor);
+
+// Post-process the generated files to make the program ID runtime-selectable.
+// `await` ensures the rewrites finish before this script exits, so a
+// chained `cargo build` doesn't race against an in-flight write.
+await import("./postprocess-codama.js");
```
jshiohaha: it might be more codama-coded to create a custom visitor to achieve this. maybe an example here: https://github.com/orca-so/whirlpools-automation/blob/5ee2d1539a296c77a9adcda02356921b8e98f54c/codama/codama-js-postprocess.cjs#L199-L206

using a new codama version, but the idea should be the same

rust-sdk/client/postprocess-codama.js
```diff
@@ -0,0 +1,54 @@
...
+  for (const name of readdirSync(dir)) {
+    if (!name.endsWith(".rs")) continue;
+    const path = join(dir, name);
+    const before = readFileSync(path, "utf8");
+    if (!before.includes(from)) continue;
+    writeFileSync(path, before.split(from).join(to));
+    rewritten += 1;
+  }
+  console.info(`postprocess-codama: rewrote ${rewritten} files in ${dir}`);
+}
```
jshiohaha: nit: we could probably simplify a bit

and if rewritten === 0, should we throw an error or something? 

```ts
writeFileSync(
  p,
  readFileSync(p, "utf8").replaceAll(
    "crate::WHIRLPOOL_ID",
    "crate::current_whirlpool_id()",
  ),
);
```

rust-sdk/client/src/program_id.rs
```diff
@@ -0,0 +1,160 @@
+//! Runtime-selectable Whirlpool program ID.
```
jshiohaha, line 1: should we overwrite the autogenerated file here? or should we add to the rust-client root, outside the generated dir? and i think we could then register the module in lib.rs and update imports 👀

rust-sdk/whirlpool/src/config.rs
```diff
@@ -71,12 +84,63 @@ impl From<WhirlpoolsConfigInput> for Pubkey {
...
+/// addresses to the canonical mainnet pair for that program. Use
+/// [`WhirlpoolProgramInput::Custom`] to set a program ID without touching the
+/// config addresses.
+pub enum WhirlpoolProgramInput {
+    /// The canonical (upgradable) Whirlpool program at `WHIRLPOOL_ID`.
+    Mutable,
+    /// The immutable Whirlpool program at `WHIRLPOOL_IMMUTABLE_ID`.
+    Immutable,
+    /// An arbitrary program address; leaves the config addresses unchanged.
+    Custom(Pubkey),
```
jshiohaha, line 106: curious, what's the use case here?

i think if a config account is initialized by 1 program, it couldn't be re-initialized by another program even if you save the keypair to re-use. theoretically you could re-use the account once it's closed, but i don't see an instruction to close existing configs

jshiohaha (commented):
still looking through, but checkpointing a few comments 📍

rust-sdk/whirlpool/src/config.rs
crnorthc: The main reason this variant was added was to copy the existing pattern of `WhirlpoolsConfigInput::Address(Pubkey)`. I think the use cases would be similar between the two - localnet or devnet deployments, forked whirlpool programs, maybe testing setups.

rust-sdk/client/src/program_id.rs
crnorthc: Good idea!

rust-sdk/client/codama.js
crnorthc: Updated to use a mapVisitor that lives inside`codama.js` now

rust-sdk/client/postprocess-codama.js
crnorthc: simplified when addressing the comment above!

rust-sdk/whirlpool/src/config.rs
jshiohaha: gotcha, i was mostly curious about `leaves the config addresses unchanged.`. like when would you ever change the program id without changing the config? and even the same program id can result in different config values depending on the cluster.

is there a more ergonomic way to expose this to consumers? it's not immediately obvious (or explicitly stated, that i know of) that someone must change config when changing program id.

jshiohaha:
1 thing i'm not sure i've fully thought through is the implication of adding the mutex to the rust client, as the lowest level crate. there are a lot of use cases, and it's difficult to know if we're going to cause issues for someone by doing this. going to think on this some more and may ask for some thoughts from others as well

some other approaches i am curious about wrt the rust client:

1. publish a separate immutable whirlpools client
    - reason:  the program id substitution works right now since the binaries are the same. i think this setup breaks as soon as the programs diverge. and unfortunately, i think that's a question of when and not if.
2. use a feature flag to choose what program id is selected
    - reason:  unlike the mutex approach, the consumer could choose which program to use simply by enabling a certain feature flag. i don't think this prevents consumers from working with both whirlpool instances either. they could simply import an aliased version of the crate with different feature flags enabled. however, this also breaks as soon as the programs diverge.

rust-sdk/whirlpool/src/config.rs
crnorthc: Hmm I see what you are saying. Maybe not set the config id at all when setting the program id and add a new variant `WhirlpoolsConfigInput::SolanaMainnetImmutable`?

rust-sdk/client/src/program_id.rs
```diff
@@ -0,0 +1,156 @@
...
+
+use solana_pubkey::{pubkey, Pubkey};
+use std::sync::Mutex;
+
+/// `whirlpool` program ID (canonical, upgradable deployment).
+pub const WHIRLPOOL_ID: Pubkey = pubkey!("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc");
+
+/// Immutable Whirlpool program ID. Bytecode-identical to [`WHIRLPOOL_ID`],
+/// deployed as a non-upgradable program.
+pub const WHIRLPOOL_IMMUTABLE_ID: Pubkey = pubkey!("iwhrLHdsgrvmnwU8GF2FSmyabSMjfHwFGJAX2ufJ3ZN");
```
yugure-orca, line 18: nit: `IMMUTABLE_WHIRLPOOL_ID` ?

In the comment, we call it "Immutable Whirlpool" program ID.

rust-sdk/client/src/program_id.rs
```diff
@@ -0,0 +1,156 @@
...
+            WhirlpoolProgram::Immutable => WHIRLPOOL_IMMUTABLE_ID,
+            WhirlpoolProgram::Address(addr) => addr,
+        }
+    }
+}
+
+static CURRENT_WHIRLPOOL_PROGRAM_ID: Mutex<Pubkey> = Mutex::new(WHIRLPOOL_ID);
+
+/// Returns the currently selected Whirlpool program address.
+pub fn current_whirlpool_id() -> Pubkey {
```
yugure-orca, line 44: nit: What came to my mind is `current_whirlpool_program_id`.

This is just my personal take, but “whirlpool” makes me think of a specific pool like SOL/USDC, whereas “whirlpool_program” clearly points to the program itself.

Also, since the global variable is CURRENT_WHIRLPOOL_PROGRAM_ID, I figured current_whirlpool_program_id() would be the most consistent choice.

rust-sdk/client/src/pda/tick_array.rs
yugure-orca, line 5: WDYT about either of these approaches?

(1) `get_<account>_address(x, x, program_id: Option<&Pubkey>)`
- None: will be populated by `current_whirlpool_id()`

(2) splitting it into two functions:
- `get_<account>_address(x, x)`
- `get_<account>_address_with_program_id(x, x, program_id: &Pubkey)`

In the second option, `get_<account>address(x, x)` would use the result of `current_whirlpool_id()` and call `get_<account>_address_with_program_id`.

In our backend routing program, we handle the Whirlpool program and Immutable Whirlpool program at the same time, because pools from both can be part of a route. That means PDA addresses for accounts like TickArray and Oracle need to be derived with the program_id in mind.

Also, it is multi-threaded program.

Changing the global variable state every time feels risky when considering other threads.

Given this, I think developers will eventually end up copying the PDA functions if it depends on the global config.
If we provide `_with_program_id` functions, they can use it without copying. And for cases where only one program is used, `get_<account>_address` remains convenient as-is.

Note: In TypeScript, the first approach can be expressed quite cleanly. But in Rust, since we can’t omit arguments, we’re effectively choosing between taking an Option or splitting it into two separate functions.

Both approaches have their pros and cons, but I think the first one is better in the long run because it avoids introducing a long _with_program_id suffix.

It also reinforces the idea that program_id is essential for PDA derivation and forces developers to be conscious of the Whirlpool program ID. From that perspective, introducing the first option as an intentional breaking change has some value.

Developers would be required to choose between None and Some, and if the global configuration is sufficient, they can simply pass None. (Code comment is required to emphasize that global config is used to resolve None)

rust-sdk/whirlpool/src/config.rs
```diff
@@ -33,6 +36,12 @@ const SOLANA_MAINNET_WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Pubkey = Pubkey::new_f
     152, 57, 209, 52, 207, 240, 174, 74, 201, 7, 87, 54,
 ]);
 
+/// The Whirlpools config account address for the immutable Whirlpool program.
+const SOLANA_MAINNET_WHIRLPOOLS_IMMUTABLE_CONFIG_ADDRESS: Pubkey = Pubkey::new_from_array([
```
yugure-orca, line 40: super nit: The constant name WHIRLPOOLS_CONFIG comes from the account name WhirlpoolsConfig, so splitting it into WHIRLPOOLS and CONFIG feels a bit unnatural. The WhirlpoolsConfig account represents configuration shared across multiple Whirlpool instances.

rust-sdk/whirlpool/src/config.rs
```diff
@@ -175,6 +240,16 @@ mod tests {
     use serial_test::serial;
     use std::str::FromStr;
 
+    /// The Whirlpools config extension account address for the immutable
+    /// Whirlpool program (`4Bsw8VVuegLmKQh2reevMBr2xw5R76WaJRKCvvxgcQrN`). PDA
+    /// derived from (`SOLANA_MAINNET_WHIRLPOOLS_IMMUTABLE_CONFIG_ADDRESS`,
+    /// immutable program).
+    const SOLANA_MAINNET_WHIRLPOOLS_IMMUTABLE_CONFIG_EXTENSION_ADDRESS: Pubkey =
```
yugure-orca, line 247: super nit: same to `WhirlpoolsConfig` (There is `WhirlpoolsConfigExtension` account type)

ts-sdk/client/src/pda/tickArray.ts
```diff
@@ -1,13 +1,13 @@
 import type { Address, ProgramDerivedAddress } from "@solana/kit";
 import { getAddressEncoder, getProgramDerivedAddress } from "@solana/kit";
-import { WHIRLPOOL_PROGRAM_ADDRESS } from "../generated/programs/whirlpool";
+import { getWhirlpoolProgramAddress } from "../program";
 
 export async function getTickArrayAddress(
```
yugure-orca, line 5: Same to the Rust version.
(It is helpful to accept programId arg, imo)
```
programId?: Address
```

yugure-orca (commented):
Thanks for the updates!

- I left a few nit comments on naming.
- It works fine if we’re only dealing with either Mutable or Immutable, but as soon as we run something like an arbitrage program that uses both under a multithreaded (Tokio) environment, even PDA helpers become timing-sensitive. That’s why I think it’s better to either provide `_with_program_id` variants or allow passing the desired `program_id: Option<&Pubkey>` explicitly as an argument.
- This isn’t limited to PDAs—higher-level helpers that build instructions like swap or increase_liquidity also don’t seem thread-safe. That said, those high-level functions prioritize convenience over performance. For example, increase_liquidity could fetch the Position account and use its owner as the program_id, which removes the need to rely on any global state. If you pass in a position from an Immutable Whirlpool, it’s clear that the instruction must target the Immutable Whirlpool program.

I can also imagine users asking something like, “Why doesn’t the SDK work with this pool?”—and it turns out the pool address belongs to the Immutable Whirlpool program.

jshiohaha:
whirlpool sdk updates made here

https://github.com/orca-so/whirlpools/pull/1301
https://github.com/orca-so/whirlpools/pull/1305
https://github.com/orca-so/whirlpools/pull/1316

PR #1300, opened 2026-05-01
Update TokenExtensions support doc

yugure-orca:
Added support for 'Pausable' extension and updated support dates for 'ScaledUiAmount'.
I added new V2 instructions as well.


jshiohaha (approved):
🚀

PR #1301, opened 2026-05-06
Add optional program_id to sdk functions

crnorthc:
## Add optional program id to SDK functions
  ### Updates to make the SDK compatible with the new immutable whirlpool program.

  **Details**

  Until now the Rust SDK could only target a single whirlpool program at a time, controlled by the
  process-global `WHIRLPOOLS_CONFIG_ADDRESS` mutex and the matching `set_whirlpools_config_address` /
  `WhirlpoolsConfigInput` API. That model can't represent two deployments living side-by-side — which is
  exactly what we need now that the immutable whirlpool program has its own program id and its own config
  account on mainnet.

  This PR replaces the global with per-call targeting and bundles the program id + config account into a
  single new `WhirlpoolDeployment` type so the two values can't drift apart.

  ### `TargetProgram`

  New type in `orca_whirlpools_client` (re-exported from `orca_whirlpools`) that pairs a program id with the
  `WhirlpoolsConfig` it operates against:

  ```rust
  WhirlpoolDeployment::mainnet()            // mutable mainnet program + mainnet config
  WhirlpoolDeployment::devnet()             // mutable mainnet program + devnet config
  WhirlpoolDeployment::mainnet_immutable()  // immutable mainnet program + immutable mainnet config
  WhirlpoolDeployment::custom(program_id, config)  // forks / local validators
  ```

  `WhirlpoolDeployment::default()` returns the mutable mainnet deployment, so existing call sites that pass None keep
   working unchanged.

  ### PDA helpers

  PDA helpers now accept an optional target. Helpers that need the config seed (whirlpool, fee_tier,
  token_badge, whirlpools_config_extension) take Option<WhirlpoolDeployment>; helpers that only need the program id
   (position, oracle, tick_array, lock_config, position_bundle, bundled_position) take Option<Pubkey>. None
  falls back to the mutable mainnet program in both cases.

  ### SDK function signatures

  Every SDK entry point now takes a single …Config struct in place of a long positional arg list. The new
  target_program / program_id field rides along with the existing parameters:

```rust
  // before
  swap_instructions(&rpc, pool, amount, mint, SwapType::ExactIn, Some(100), Some(signer)).await?;

  // after
  swap_instructions(
      &rpc, 
      pool,
      amount,
      mint,
      SwapType::ExactIn,
      SwapConfig {
          slippage_tolerance_bps: Some(100),
          signer: Some(signer),
          program_id: None,  // or Some(&target.id())
  }).await?;
```

  **Affected functions:** create_splash_pool_instructions, create_concentrated_liquidity_pool_instructions,
  swap_instructions, increase_liquidity_instructions, decrease_liquidity_instructions,
  close_position_instructions, harvest_position_instructions, open_full_range_position_instructions,
  open_position_instructions, open_position_instructions_with_tick_bounds, fetch_splash_pool,
  fetch_concentrated_liquidity_pool, fetch_whirlpools_by_token_pair.

  ### Codama generation

  codama.js switches from renderVisitor to getRenderMapVisitor so the generated output can be post-processed
  before it's written to disk. The Owner impl on each generated account is marked deprecated..

  ### Removed

  - WHIRLPOOLS_CONFIG_ADDRESS and WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS statics
  - set_whirlpools_config_address and WhirlpoolsConfigInput
  - The associated entries in reset_configuration

  ### Migration
```
  ┌──────────────────────────────────────────────────────────────────┬───────────────────────────────────┐
  │                               Old                                │                New                │
  ├──────────────────────────────────────────────────────────────────┼───────────────────────────────────┤
  │ set_whirlpools_config_address(WhirlpoolsConfigInput::SolanaDevne │ pass WhirlpoolDeployment::devnet() to   │
  │ t)                                                               │ the call you care about           │
  ├──────────────────────────────────────────────────────────────────┼───────────────────────────────────┤
  │ WhirlpoolsConfigInput::SolanaMainnetImmutable                    │ WhirlpoolDeployment::mainnet_immutable( │
  │                                                                  │ )                                 │
  ├──────────────────────────────────────────────────────────────────┼───────────────────────────────────┤
  │ get_whirlpool_address(&config, …)                                │ get_whirlpool_address(Some(target │
  │                                                                  │ ), …) or (None, …) for default    │
  ├──────────────────────────────────────────────────────────────────┼───────────────────────────────────┤
  │ swap_instructions(&rpc, pool, amt, mint, ty, slip, signer)       │ swap_instructions(&rpc,           │
  │                                                                  │ SwapConfig { … })                 │
  └──────────────────────────────────────────────────────────────────┴───────────────────────────────────┘
```
  Existing code that relied on the global default (mutable mainnet program + mainnet config) continues to
  work by passing None everywhere a target is now optional.


rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -124,6 +126,7 @@ pub async fn decrease_liquidity_instructions(
     param: DecreaseLiquidityParam,
     slippage_tolerance_bps: Option<u16>,
     authority: Option<Pubkey>,
+    program_id: Option<&Pubkey>,
```
jshiohaha: nice! i like this optional parameter approach. 1 thing that comes to mind is that this will break all consumers, and the list of params is getting a bit long. i wonder if we can (1) future-proof additional changes + (2) simplify call-sites by adding a config/options struct, e.g.

```rust
// or DecreaseLiquidityOpts
struct DecreaseLiquidityConfig {
    param: DecreaseLiquidityParam,
    slippage_tolerance_bps: Option<u16>,
    authority: Option<Pubkey>,
    program_id: Option<Pubkey>,
}
```

with Default trait, call sites could simply be smth like

```rust
decrease_liquidity_instructions(&rpc, mint, DecreaseLiquidityConfig {
    slippage_tolerance_bps: Some(50),
    ..Default::default()
}).await?

// or
decrease_liquidity_instructions(&rpc, mint, DecreaseLiquidityConfig::default()).await?

// instead of
decrease_liquidity_instructions(
    rpc,
    position_mint_address,
    DecreaseLiquidityParam::TokenA(100),
    None,
    None,
    None,
).await?
```

rust-sdk/whirlpool/src/decrease_liquidity.rs
jshiohaha: would you replace program_id here too if provided?

crnorthc: Yessir, good catch!

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: I like your idea of simplifying the call sites with struct based params. I added it in here, but if there are any strong opinions against, it would be easy to revert

rust-sdk/whirlpool/src/pool.rs
crnorthc, line 146: Do we want to add the same functionality here?

rust-sdk/whirlpool/src/pool.rs
```diff
@@ -206,10 +206,16 @@ pub async fn fetch_concentrated_liquidity_pool(
 ) -> Result<PoolInfo, Box<dyn Error>> {
```
crnorthc, line 220: Do we want to add the same functionality here?

rust-sdk/whirlpool/src/pool.rs
crnorthc, line 334: Do we want to add the same functionality here?

rust-sdk/client/src/pda/whirlpool.rs
```diff
@@ -2,11 +2,13 @@ use crate::generated::programs::WHIRLPOOL_ID;
...
 use solana_pubkey::Pubkey;
 
+/// A program_id of None resolves to the original whirlpool program id ("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc")
 pub fn get_whirlpool_address(
     whirlpools_config: &Pubkey,
     token_mint_a: &Pubkey,
     token_mint_b: &Pubkey,
     fee_tier_index: u16,
+    program_id: Option<&Pubkey>,
 ) -> Result<(Pubkey, u8), ProgramError> {
```
crnorthc: This could try to map the config addresses to the program ids instead of explicitly taking in the program id

rust-sdk/client/src/pda/whirlpools_config_extension.rs
```diff
@@ -2,12 +2,15 @@ use crate::generated::programs::WHIRLPOOL_ID;
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
+/// A program_id of None resolves to the original whirlpool program id ("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc")
 pub fn get_whirlpools_config_extension_address(
     whirlpools_config: &Pubkey,
+    program_id: Option<&Pubkey>,
```
crnorthc: Same here in regards to mapping config to program id

rust-sdk/client/src/pda/fee_tier.rs
```diff
@@ -2,17 +2,20 @@ use crate::generated::programs::WHIRLPOOL_ID;
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
+/// A program_id of None resolves to the original whirlpool program id ("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc")
 pub fn get_fee_tier_address(
     whirlpools_config: &Pubkey,
     fee_tier_index: u16,
+    program_id: Option<&Pubkey>,
```
crnorthc: Same here in regards to mapping config to program id

rust-sdk/client/src/pda/token_badge.rs
```diff
@@ -2,17 +2,20 @@ use crate::generated::programs::WHIRLPOOL_ID;
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
+/// A program_id of None resolves to the original whirlpool program id ("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc")
 pub fn get_token_badge_address(
     whirlpools_config: &Pubkey,
     token_mint: &Pubkey,
+    program_id: Option<&Pubkey>,
 ) -> Result<(Pubkey, u8), ProgramError> {
```
crnorthc, line 11: Same here in regards to mapping config to program id

rust-sdk/whirlpool/src/config.rs
```diff
@@ -33,6 +33,18 @@ const SOLANA_MAINNET_WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS: Pubkey = Pubkey::new_f
...
+    116, 62, 1, 180, 69, 120, 160, 190, 89, 235, 62, 144, 64, 210, 160, 63, 11, 157, 46, 125, 206,
+    46, 108, 53, 135, 184, 54, 34, 43, 41, 184, 124,
+]);
+
+/// The Immutable Whirlpools program's config account address for Solana Devnet.
+const SOLANA_DEVNET_IMMUTABLE_WHIRLPOOLS_CONFIG_ADDRESS: Pubkey = Pubkey::new_from_array([
+    217, 51, 106, 61, 244, 143, 54, 30, 87, 6, 230, 156, 60, 182, 182, 217, 23, 116, 228, 121, 53,
+    200, 82, 109, 229, 160, 245, 159, 33, 90, 35, 106,
+]);
+
```
crnorthc: I have the mainnet immuatble whirlpools program config address as `8pm8erUsaMpmZ47LttHAPgnDx7xGZUvxY4q47vTCs5Nj` and the devnet immuatble whirlpools program config address as `FcrweFY1G9HJAHG5inkGB6pKg1HZ6x9UC2WioAfWrGkR`

rust-sdk/client/src/generated/accounts/whirlpool.rs
```diff
@@ -5,145 +5,169 @@
...
-      fn owner() -> Pubkey {
+// TODO: How do we handle this?
+#[cfg(feature = "anchor")]
+impl anchor_lang::Owner for Whirlpool {
+    fn owner() -> Pubkey {
         crate::WHIRLPOOL_ID
-      }
-  }
+    }
+}
```
crnorthc: Apologies for the diff here. Only thing added in this file was the comment.

I am not too sure how to best update the `owner()` associated function. Do we remove it? Or perhaps make it a method and try to deduce it from `self.whirlpool_config`?

crnorthc: This also applies to the other generated accounts

rust-sdk/whirlpool/src/create_pool.rs
```diff
@@ -190,108 +217,137 @@ pub async fn create_splash_pool_instructions(
...
-
-    let token_badge_b = get_token_badge_address(&whirlpools_config_address, &token_b)?.0;
+    let pool_address = get_whirlpool_address(
+        &whirlpools_config_address,
+        &config.token_a,
+        &config.token_b,
+        config.tick_spacing,
+        config.program_id,
+    )?
+    .0;
```
crnorthc: The use of the global `WHIRLPOOLS_CONFIG_ADDRESS` does not seem ideal here and may lead to some confusion if not previously set when providing a program_id other than the default

rust-sdk/whirlpool/src/config.rs
```diff
@@ -73,7 +93,7 @@ pub fn set_whirlpools_config_address(input: WhirlpoolsConfigInput) -> Result<(),
     let address: Pubkey = input.into();
     *WHIRLPOOLS_CONFIG_ADDRESS.try_lock()? = address;
     *WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS.try_lock()? =
-        get_whirlpools_config_extension_address(&address)?.0;
+        get_whirlpools_config_extension_address(&address, None)?.0;
     Ok(())
 }
```
crnorthc: Wasn't too sure on how to best handle this. I thought of a few approaches:
1. Remove the setting of the config extension address here and have users set it similarly to how the config is set (enum with a catchall Address(Pubkey) variant)
2. change the signature to `set_whirlpools_config_address(input: WhirlpoolsConfigInput, program_id: Option<&Pubkey>)`

crnorthc: It seems like `WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS` is not being used within this repo. Would it be possible to remove all together? `get_whirlpools_config_extension_address()` would still provide a way for users to obtain this address.

crnorthc: Thinking about this more, are there any strong opinions against removing both `WHIRLPOOLS_CONFIG_ADDRESS` and `WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS`?

The extension address is not used within this repo, so that seems safe-ish to remove, given the functionality to retrieve it will remain with `get_whirlpools_config_extension_address()`.

As for the config address, it is currently only used in a few places:
- `create_concentrated_liquidity_pool_instructions()`
- `fetch_concentrated_liquidity_pool()`
- `fetch_whirlpools_by_token_pair()`

`create_concentrated_liquidity_pool_instructions()` could take an additional whirlpools_config param of either Option<&Pubkey> or a variant from the enum `WhirlpoolsConfigInput`

`fetch_concentrated_liquidity_pool()` and `fetch_whirlpools_by_token_pair()` would have to take both the config and program id

rust-sdk/whirlpool/src/config.rs
```diff
@@ -46,6 +58,8 @@ pub enum WhirlpoolsConfigInput {
     Address(Pubkey),
     SolanaMainnet,
     SolanaDevnet,
+    SolanaMainnetImmutable,
+    SolanaDevnetImmutable,
     EclipseMainnet,
     EclipseTestnet,
 }
```
crnorthc: Do you think it would be more intuitive to expose the whirlpool programs similar to how we are with the`WhirlpoolsConfigInput`? Then have call sites take an optional enum variant instead of an optional Pubkey for both the config and program?

```
/// Input type for a whirlpools program address/id
pub enum WhirlpoolProgramInput {
    Address(Pubkey),
    /// The default and something along the lines of "this is probably the one you want to use"
    Whirlpool,
    ImmutableWhirlpool,
}
```

with an additional 
```
impl From<WhirlpoolProgramInput> for Pubkey {
    fn from(val: WhirlpoolProgramInput) -> Self {
        match val {
            WhirlpoolProgramInput::Address(pubkey) => pubkey,
            WhirlpoolProgramInput::Whirlpool => WHIRLPOOL_PROGRAM_ADDRESS,
            WhirlpoolProgramInput::ImmutableWhirlpool => IMMUTABLE_WHIRLPOOL_PROGRAM_ADDRESS
        }
    }
}
```

rust-sdk/client/src/pda/whirlpool.rs
crnorthc: Just brainstorming, but for call sites requiring both the program and config addresses, we could replace with something like:

```
pub enum TargetWhirlpoolCluster {
    Addresses((Pubkey, Pubkey)),
    MainnetWhirlpool,
    DevnetWhirlpool,
    MainnetImmutableWhirlpool,
    DevnetImmutableWhirlpool,
}

impl TargetWhirlpoolCluster {
    pub fn program_id(&self) -> Pubkey {
        match self {
            TargetWhirlpoolCluster::Addresses((program_id, _)) => *program_id,
            TargetWhirlpoolCluster::MainnetWhirlpool | TargetWhirlpoolCluster::DevnetWhirlpool => {
                WHIRLPOOL_PROGRAM_ADDRESS
            }
            TargetWhirlpoolCluster::MainnetImmutableWhirlpool
            | TargetWhirlpoolCluster::DevnetImmutableWhirlpool => {
                IMMUTABLE_WHIRLPOOL_PROGRAM_ADDRESS
            }
        }
    }

    pub fn config_address(&self) -> Pubkey {
        match self {
            TargetWhirlpoolCluster::Addresses((_, config)) => *config,
            TargetWhirlpoolCluster::MainnetWhirlpool => SOLANA_MAINNET_WHIRLPOOLS_CONFIG_ADDRESS,
            TargetWhirlpoolCluster::DevnetWhirlpool => SOLANA_DEVNET_WHIRLPOOLS_CONFIG_ADDRESS,
            TargetWhirlpoolCluster::MainnetImmutableWhirlpool => {
                SOLANA_MAINNET_IMMUTABLE_WHIRLPOOLS_CONFIG_ADDRESS
            }
            TargetWhirlpoolCluster::DevnetImmutableWhirlpool => {
                SOLANA_DEVNET_IMMUTABLE_WHIRLPOOLS_CONFIG_ADDRESS
            }
        }
    }
}
```

rust-sdk/client/src/generated/accounts/whirlpool.rs
jshiohaha: might be able to avoid diff with `yarn format && yarn lint`

jshiohaha: as for how to handle, good q - i'm not sure how often it would actually get used since it's behind the anchor feature flag

removing is 1 option. another left curve take is that we could replace `crate::WHIRLPOOL_ID` with smth like `unimplemented!("fetch account via rpc to determine owner")`

rust-sdk/client/src/pda/whirlpool.rs
jshiohaha: > This could try to map the config addresses to the program ids instead of explicitly taking in the program id

i personally prefer exposing both params

however, i do like your idea of setting up an enum that would make simpler to work with and more readable? 

i know this was an example, but some ideas wrt the setup

- i think solana has an existing cluster enum, e.g. `ClusterType`
- in the naming, i think we can avoid mentioning whirlpool multiple times
- instead of 1 big enum, could we separate cluster and mutability concepts?, e.g.
  - just an example off the top of my head, maybe naming can still be improved

```rust
pub enum ProgramVariant {
    Mutable,
    Immutable,
}

pub struct TargetProgram {
    pub cluster: ClusterType,
    pub variant: ProgramVariant,
}

impl TargetProgram {
    pub fn id(&self) -> Pubkey { ... }
    pub fn config_address(&self) -> Pubkey { ... }

    // QoL helpers
    pub fn mainnet() -> Self { ... }
    pub fn mainnet_immutable() -> Self { ... }
}
```

rust-sdk/whirlpool/src/config.rs
jshiohaha: it looks like `FcrweFY1G9HJAHG5inkGB6pKg1HZ6x9UC2WioAfWrGkR ` is the devnet config for existing whirlpools. it doesn't look like there have been any transactions against immutable whirlpools on devnet https://solscan.io/account/iwhrLHdsgrvmnwU8GF2FSmyabSMjfHwFGJAX2ufJ3ZN?cluster=devnet

rust-sdk/whirlpool/src/config.rs
jshiohaha: i briefly looked at the underlying code using `WHIRLPOOLS_CONFIG_ADDRESS` and `WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS`. i wonder if we can get rid of these completely?

- usage of WHIRLPOOLS_CONFIG_ADDRESS isn't that high https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+%2F%28%3F-i%29WHIRLPOOLS_CONFIG_ADDRESS%2F&type=code
- it doesn't look like WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS is used at all https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS+language%3ARust&type=code

rust-sdk/whirlpool/src/config.rs
crnorthc: Ooo good catch, I copied the wrong one from a commit this [PR](https://github.com/orca-so/kelpfarm/pull/1703). 

It should be `GWzKtpy7tYrJnAwDSyJcrxejTNsdvAFVC9mhNUJy8Ty9`

crnorthc: Maybe, not 100% sure

jshiohaha: i think it's also ok to leave undefined for now and put behind a `todo!(...)` or comment

rust-sdk/client/src/pda/whirlpool.rs
crnorthc: Do we want to add/maintain support for the `Address(Pubkey)` variant? Seems like it could be helpful for a few use cases (development and forks)

jshiohaha: in these cases, dyt it makes more sense to add here or just have someone directly provide parameters to functions?

rust-sdk/client/src/generated/accounts/whirlpool.rs
crnorthc: I like the `unimplemented!(...)` approach, it gives good direction to those upgrading versions.

rust-sdk/client/src/pda/whirlpool.rs
crnorthc: If so it could be something like:

```
// already existing - maybe rename to ConfigVariant
pub enum WhirlpoolsConfigInput {
    Address(Pubkey),
    SolanaMainnet, // Mainnet?
    SolanaDevnet, // Devnet?
    SolanaMainnetImmutable, // MainnetImmutable?
    SolanaDevnetImmutable, // DevnetImmutable?
    EclipseMainnet, // remove?
    EclipseTestnet, // remove?
}

pub enum ProgramVariant {
    Address(Pubkey),
    Mutable,
    Immutable,
}

pub struct TargetProgram {
    pub config: ClusterType,
    pub variant: ProgramVariant,
}

impl TargetProgram {
    pub fn id(&self) -> Pubkey { ... }
    pub fn config_address(&self) -> Pubkey { ... }

    // QoL helpers
    pub fn mainnet() -> Self { ... }
    pub fn mainnet_immutable() -> Self { ... }
    ...
}
```

crnorthc: > in these cases, dyt it makes more sense to add here or just have someone directly provide parameters to functions?

Ohhh, I am on the same page now. Originally what I was thinking was the the struct/enum would replace both params in a callsite like:
`sdk_function(program_id: Option<&Pubkey>, whirlpool_config_address: Pubkey)` -> `sdk_function(target_program: TargetProgram)`

If I am indeed on the same page, would I be correct in saying you are thinking about posing `TargetProgram` as a utility to be used outside the sdk?

I am happy to go in either direction, but if the second is preferred (`TargetProgram` as a utility) I would suggest making callsites with both program and config require both (instead of an optional program) and add doc comments to point users in the right direction. In this case the example above would look something like:

```rust
/// Retrieves the whirlpool address by seeds
///
/// This function derives a whirlpool's address from the seeds (mints and fee tier index).
///
/// # Arguments
/// * `program_id` - The target whirlpool program address public key
/// * `whirlpools_config` - The target whirlpool config address public key
/// * `token_mint_a` - The whirlpool's token a address public key
/// * `token_mint_b` - The whirlpool's token b address public key
/// * `fee_tier_index` - The whirlpool's fee tier index
///
/// # Returns
///
/// A `Result` containing `(Pubkey, u8)` on success:
///
/// * `Pubkey` - The whirlpool's address public key
/// * `u8` - The whirlpool's bump seed
///
/// # Errors
///
/// This function will return an error if:
/// - Provided seeds do not result in a valid address
///
/// # Example
///
/// ```rust
/// use orca_whirlpools::{
///     get_whirlpool_address, TargetProgram
/// };
/// use solana_pubkey::Pubkey;
///
/// #[tokio::main]
/// async fn main() {
///     let mainnet_program = TargetProgram::mainnet();
///     let token_mint_a = Pubkey::from_str("So11111111111111111111111111111111111111112").unwrap();
///     let token_mint_b = Pubkey::from_str("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v").unwrap();
///     let fee_tier_index = 4u16;
///
///     let (whirlpool_address, bump_seed) = get_whirlpool_address(
///          &mainnet_program.id(),
///          &mainnet_program.config_address(),
///          &token_mint_a,
///          &token_mint_b,
///          fee_tier_index
///     )?;
///
///     println!("Whirlpool address: {}", whirlpool_address);
/// }
/// ```
pub fn get_whirlpool_address(
    program_id: &Pubkey,
    whirlpools_config: &Pubkey,
    token_mint_a: &Pubkey,
    token_mint_b: &Pubkey,
    fee_tier_index: u16,
) -> Result<(Pubkey, u8), ProgramError> {
    let fee_tier_index_bytes = fee_tier_index.to_le_bytes();
    let seeds = &[
        b"whirlpool",
        whirlpools_config.as_ref(),
        token_mint_a.as_ref(),
        token_mint_b.as_ref(),
        fee_tier_index_bytes.as_ref(),
    ];

    Pubkey::try_find_program_address(seeds, program_id).ok_or(ProgramError::InvalidSeeds)
}
```

jshiohaha: ahh i see! sorry, i wasn't specific. i think both are reasonable paths. i was originally thinking we would keep functions as

```rust
sdk_function(program_id: Option<&Pubkey>, whirlpool_config_address: Pubkey)
```

and then the `TargetProgram` struct would be a utility for users who want to make it easier to configure invocations instead of having to always directly inject pubkeys, e.g.

(**option 1**)

```rust
let target: TargetProgram = { ... };
sdk_function(target.program_id(), target.config());
```

(**option 2**)

instead of

```rust
sdk_function(
    Some("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc"),
    "2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ"
);
```

and if someone wanted a more flexible setup, they could just do option 2 🤷🏻‍♂️

crnorthc: Gotcha! Would you be alright with making the program field required instead of optional? It seems weird to have the config required, but the program optional.

Edit: Or have both optional

jshiohaha: > making the program field required instead of optional

i think that makes sense when both config and program id are required. but now i'm wondering if your approach is better (aka more elegant)

with the following, we could do smth like this

```rust
fn sdk_function(target: Option<TargetProgram>) {
    let target = target.unwrap_or_default();
    // ...
}
```

example setup

```rust
// note: idk what a good name is yet
pub struct TargetProgram {
    pub program_id: Pubkey,
    pub config: Pubkey,
}

impl TargetProgram {
    pub fn id(&self) -> Pubkey {
        self.program_id
    }

    pub fn config_address(&self) -> Pubkey {
        self.config
    }

    pub fn mainnet() -> Self {
        Self {
            program_id: WHIRLPOOL_ID,
            config: MAINNET_CONFIG,
        }
    }

    pub fn custom(program_id: Pubkey, config: Pubkey) -> Self {
        Self {
            program_id,
            config,
        }
    }

    // ... etc ...
}

impl Default for TargetProgram {
    fn default() -> Self {
        Self::mainnet()
    }
}
```

crnorthc: I like that a lot

rust-sdk/whirlpool/src/config.rs
crnorthc: Looks very straight forward to remove these. Will continue in this direction

.changeset/late-hornets-tell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@orca-so/whirlpools-rust": major
+"@orca-so/whirlpools-rust-client": major
```
jshiohaha, line 3: i think it's technically breaking changes for `@orca-so/whirlpools-example-rust-repositioning-bot` too

examples/rust-sdk/whirlpool_repositioning_bot/src/position_manager.rs
```diff
@@ -72,9 +73,12 @@ pub async fn run_position_manager(
...
         let close_position_instructions = close_position_instructions(
             rpc,
-            position.position_mint,
-            Some(args.slippage_tolerance_bps),
-            None,
+            ClosePositionConfig {
+                position_mint_address: position.position_mint,
+                slippage_tolerance_bps: Some(args.slippage_tolerance_bps),
+                authority: None,
+                program_id: None,
```
jshiohaha: maybe it's better to be explicit for the example, but you could also do

```rust
ClosePositionConfig {
    position_mint_address: position.position_mint,
    slippage_tolerance_bps: Some(args.slippage_tolerance_bps),
    ..Default::default()
}
```

rust-sdk/client/src/pda/fee_tier.rs
```diff
@@ -1,18 +1,24 @@
-use crate::generated::programs::WHIRLPOOL_ID;
+use crate::TargetProgram;
```
jshiohaha: i'm not actually sure about this name for the struct. it was just an example at the time. do you like it? or have any ideas for better names? 😅

small thing, but it's important to decide a good one since we'll be stuck with it

rust-sdk/client/src/pda/token_badge.rs
```diff
@@ -1,18 +1,24 @@
-use crate::generated::programs::WHIRLPOOL_ID;
+use crate::TargetProgram;
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
+/// Derives the token badge PDA for the given mint under the supplied target program.
```
jshiohaha, line 5: nit: curious if there's a particular reason why some functions have additional info and others don't?

rust-sdk/client/src/pda/whirlpools_config_extension.rs
```diff
@@ -1,13 +1,20 @@
-use crate::generated::programs::WHIRLPOOL_ID;
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
```
jshiohaha, line 4: nit: extra newline?

rust-sdk/whirlpool/src/tests/rpc.rs
```diff
@@ -87,7 +87,7 @@ impl RpcContext {
         )
         .unwrap();
 
-        let default_fee_tier = get_fee_tier_address(&config, 128).unwrap().0;
+        let default_fee_tier = get_fee_tier_address(None, 128).unwrap().0;
```
jshiohaha: purely for readability, wdyt about adding `Some(TargetProgram::default())` to the test instead of None?

rust-sdk/whirlpool/src/create_pool.rs
```diff
@@ -43,15 +41,27 @@ pub struct CreatePoolInstructions {
...
+    pub token_a: Pubkey,
+    /// The public key of the second token mint address to include in the pool.
+    pub token_b: Pubkey,
+    /// An optional initial price of token A in terms of token B. Defaults to 1.0 if not provided.
+    pub initial_price: Option<f64>,
+    /// An optional public key of the account funding the initialization process. Defaults to the global funder if not provided.
+    pub funder: Option<Pubkey>,
+    /// An optional public key of the whirlpool program to target.
+    /// Defaults to the mutable whirlpool program and mainnet config if not provided
+    pub target_program: Option<TargetProgram>,
```
jshiohaha: nit: might be more convenient to link to the doc comment on the referenced struct. this way, i think we only have to update 1 place instead many if behavior changes

```rust
  /// The Whirlpool program and config account to target.
  /// Uses [`TargetProgram::default`] when `None`.
```

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -60,6 +66,20 @@ pub struct DecreaseLiquidityInstruction {
...
+    /// The public key of the NFT mint address representing the pool position.
+    pub position_mint_address: Pubkey,
+    /// A variant of `DecreaseLiquidityParam` specifying the liquidity reduction method (by Token A, Token B, or liquidity amount).
+    pub param: DecreaseLiquidityParam,
+    /// An optional slippage tolerance in basis points. Defaults to the global slippage tolerance if not provided.
+    pub slippage_tolerance_bps: Option<u16>,
+    /// An optional public key of the account authorizing the liquidity removal. Defaults to the global funder if not provided.
+    pub authority: Option<Pubkey>,
+    /// An optional public key of the whirlpool program to target. Defaults to the original whirlpool program ("whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc")
+    pub program_id: Option<&'a Pubkey>,
```
jshiohaha: wdyt about removing the lifetime parameter from all the configs related to the optional pubkey type. it's kind of weird that other pubkeys don't have the same setup, and i'd prefer to be consistent when possible

and i think the value is small enough value that there's really not much downside to the extra copy/clone operation

kaizen-orca:
I think there’s still one real immutable-program hole in the Rust SDK surface.

This PR does a good job threading `program_id` / `TargetProgram` through the action-style entry points, but `fetch_positions_for_owner` in `rust-sdk/whirlpool/src/position.rs` still derives everything with the default mutable program:

- `get_position_address(&x.mint, None)`
- `get_position_bundle_address(&x.mint, None)`
- `get_bundled_position_address(..., None)`

So wallet position discovery is still pinned to the legacy program even after the rest of the SDK learns how to target immutable/custom deployments. In practice that means a caller can open/close/harvest against an immutable deployment when they already know the position mint, but `fetch_positions_for_owner` can still silently miss those same positions and bundles because it derives the wrong PDAs from the owned NFTs.

I’d recommend extending this API to accept the same target/program context and threading it through the position/bundle PDA derivations. Otherwise the migration is incomplete at a pretty important “list my positions” entry point.

kaizen-orca:
One structural nit after a second pass: the targeting abstraction landed a bit unevenly across the Rust surface.

Right now callers have to understand two different ways of expressing the same deployment choice:

- config-seeded paths use `Option<TargetProgram>`
- program-only paths use `Option<&Pubkey>`

That split is technically defensible, but in practice it makes the migration harder to reason about because the SDK layer now mixes `target_program` and `program_id` depending on which function you touched (`create_*_pool_*` / `fetch_*pool*` vs `swap` / `increase_liquidity` / `harvest` / etc.).

I think there’s a maintainability win in standardizing the SDK-facing config structs on a single deployment field (`Option<TargetProgram>` everywhere), and then projecting to `target.id()` internally for the PDA helpers / instruction builders that only need the program id. That would give callers one mental model, avoid the lifetime noise from `Option<&Pubkey>`, and make it much harder for the public API to drift into “sometimes this needs the full target, sometimes just the raw program id” over time.

kaizen-orca:
Another implementation-detail nit: the generated-account `Owner` change feels internally inconsistent right now.

`codama.js` rewrites every generated account type to:

```rust
impl anchor_lang::Owner for X {
    fn owner() -> Pubkey {
        unimplemented!("fetch account via rpc to determine owner")
    }
}
```

But `rust-sdk/client/src/state/tick_array.rs` still gives the wrapper enum a concrete static owner:

```rust
impl anchor_lang::Owner for TickArray {
    fn owner() -> Pubkey {
        Pubkey::from(crate::WHIRLPOOL_ID.to_bytes())
    }
}
```

So after this PR, some account types panic on `owner()` while the `TickArray` wrapper still advertises a fixed mutable-program owner. That’s a pretty awkward semantic split for consumers trying to understand what the `anchor` feature still promises.

If the new rule is “static `Owner` is no longer meaningful once multiple deployments exist”, I’d make that consistent across the wrapper layer too (or explicitly wall the `anchor` integration off in a different way). Keeping one surviving static owner path here feels like exactly the kind of exception that will confuse future maintenance.

rust-sdk/whirlpool/src/position.rs
```diff
@@ -162,12 +159,12 @@ pub async fn fetch_positions_for_owner(
 
     let position_addresses: Vec<Pubkey> = potiential_tokens
         .iter()
-        .map(|x| get_position_address(&x.mint).map(|x| x.0))
+        .map(|x| get_position_address(&x.mint, None).map(|x| x.0))
```
jshiohaha: agree with kaizen, it seems like we might want to propagate program id here?

sorry for repeating, just didn't want to miss

rust-sdk/client/src/pda/whirlpools_config_extension.rs
```diff
@@ -1,13 +1,20 @@
...
 
+use crate::TargetProgram;
+
+/// Derives the whirlpools config extension PDA under the supplied target program.
+///
+/// Passing `None` for `target_program` falls back to [`TargetProgram::default`] (the mutable
+/// mainnet deployment).
 pub fn get_whirlpools_config_extension_address(
-    whirlpools_config: &Pubkey,
+    target_program: Option<TargetProgram>,
```
jshiohaha: after seeing examples like this, dyt it makes sense for the parameter to be optional? unlike ts where you can simply omit, it looks _kind of_ weird for the only parameter to be None 🤔

rust-sdk/client/src/pda/fee_tier.rs
```diff
@@ -1,18 +1,24 @@
...
 use solana_program_error::ProgramError;
 use solana_pubkey::Pubkey;
 
+/// Derives the fee tier PDA for the given fee tier index under the supplied target program.
+///
+/// Passing `None` for `target_program` falls back to [`TargetProgram::default`] (the mutable
+/// mainnet deployment).
 pub fn get_fee_tier_address(
```
jshiohaha, line 8: nit: wdyt about reversing the inputs in these functions so that it's non-optional > optional?

rust-sdk/whirlpool/src/create_pool.rs
```diff
@@ -190,108 +214,118 @@ pub async fn create_splash_pool_instructions(
...
-    let funder = funder.unwrap_or(*FUNDER.try_lock()?);
+    let target_program = config.target_program.unwrap_or_default();
+
+    let initial_price = config.initial_price.unwrap_or(1.0);
+    let funder = config.funder.unwrap_or(*FUNDER.try_lock()?);
     if funder == Pubkey::default() {
         return Err("Funder must be provided".into());
     }
-    if order_mints(token_a, token_b)[0] != token_a {
+    if order_mints(config.token_a, config.token_b)[0] != config.token_a {
```
jshiohaha: extreme nit: we could avoid a bunch of diffs below by de-structuring the fields like

```rust
let CreateConcentratedLiquidityPoolConfig {
    target_program,
    token_a,
    token_b,
    tick_spacing,
    initial_price,
    funder,
} = config;
```

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -60,6 +66,20 @@ pub struct DecreaseLiquidityInstruction {
     pub additional_signers: Vec<Keypair>,
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct DecreaseLiquidityConfig<'a> {
+    /// The public key of the NFT mint address representing the pool position.
+    pub position_mint_address: Pubkey,
```
jshiohaha: i'd assume that `position_mint_address` is a required normal parameter since it's necessary for this operation. it doesn't really specify _how_ to perform the action is executed, which is kind of the case for other params. this would also apply to all other similar ix's.

wdyt?

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -881,29 +954,34 @@ mod tests {
...
-                    token_max_b: 100_000,
+                IncreaseLiquidityConfig {
+                    position_mint_address: position_mint,
+                    param: IncreaseLiquidityParam {
+                        token_max_a: 100_000,
+                        token_max_b: 100_000,
+                    },
+                    slippage_tolerance_bps: Some(100),
+                    authority: Some(ctx.signer.pubkey()),
+                    program_id: None,
```
jshiohaha: it looks like all tests use `program_id: None,`. is there any case in which you think it makes sense to provide a different value?

rust-sdk/whirlpool/src/position.rs
```diff
@@ -81,7 +81,7 @@ fn get_position_in_bundle_addresses(position_bundle: &PositionBundle) -> Vec<Pub
         let bit_index = i % 8;
         if position_bundle.position_bitmap[byte_index] & (1 << bit_index) != 0 {
             let result =
-                get_bundled_position_address(&position_bundle.position_bundle_mint, i as u8);
+                get_bundled_position_address(&position_bundle.position_bundle_mint, i as u8, None);
```
jshiohaha: agree with kaizen, it seems like we might want to propagate program id here?

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -1158,10 +1229,13 @@ mod tests {
         };
         let res = increase_liquidity_instructions(
             &ctx.rpc,
-            position_mint,
-            param,
-            Some(100), // slippage
```
jshiohaha, line 1163: nice thing about the config is that we don't need contextual comments like this anymore 🚀

.changeset/late-hornets-tell.md
crnorthc: Good catch!

rust-sdk/client/src/pda/fee_tier.rs
crnorthc: I mean it is not awful. wdyt about `WhirlpoolDeployment` or `WhirlpoolTarget`?

rust-sdk/client/src/pda/whirlpools_config_extension.rs
crnorthc: It makes sense to require!

rust-sdk/client/src/pda/fee_tier.rs
crnorthc: That makes sense

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Not quite sure what you mean here

rust-sdk/client/src/pda/fee_tier.rs
jshiohaha: > I mean it is not awful

i'd love to land on smth we feel good about.. not smth we don't actively hate 😂 

i also asked a few different LLMs and couldn't find anything better than `WhirlpoolDeployment`

so i'm down to move forward with that one if you're onboard

rust-sdk/whirlpool/src/decrease_liquidity.rs
jshiohaha: ah, sorry - i'm basically asking about the decision of what to include in the config vs normal parameters. like you added rpc as a normal parameter. curious about the thought process to add `position_mint_address` to the config

rust-sdk/client/src/pda/token_badge.rs
crnorthc: My guess is Claude did not want to overwrite my comments when updating the rest

rust-sdk/client/src/pda/fee_tier.rs
crnorthc: Yessir, that sounds good to me

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Gotcha! I just copied the existing pattern for `GetIncreaseLiquidityInstructionsParams` without giving it much extra thought. I am happy to go in whatever direction though

rust-sdk/client/src/pda/token_badge.rs
jshiohaha: gotcha! it would be nice to stay consistent where possible

rust-sdk/whirlpool/src/decrease_liquidity.rs
jshiohaha: ah ah, ok i see. i didn't realize we already had an existing paradigm - feel free to leave as is! resolving this comment

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Good callout - added test cases for immutable whirlpools across both crates

rust-sdk/client/src/config.rs
```diff
@@ -88,9 +88,95 @@ impl TargetProgram {
...
-impl Default for TargetProgram {
-    /// Defaults to the mutable mainnet deployment ([`TargetProgram::mainnet`]).
+impl Default for WhirlpoolDeployment {
+    /// Defaults to the mutable mainnet deployment ([`WhirlpoolDeployment::mainnet`]).
     fn default() -> Self {
         Self::mainnet()
     }
 }
+
+#[cfg(test)]
```
jshiohaha: nit: these seem like a little overkill, but feel free to leave them if you like

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -905,123 +908,150 @@ mod tests {
...
-    #[case("A-TEA",  "equally centered", -100, 100)]
-    #[case("A-TEA",  "one sided A",      -100, -1)]
-    #[case("A-TEA", "one sided B", 1, 100)]
-    #[case("TEA-TEB","equally centered", -100, 100)]
-    #[case("TEA-TEB","one sided A",      -100, -1)]
-    #[case("TEA-TEB", "one sided B", 1, 100)]
-    #[case("A-TEFee","equally centered", -100, 100)]
-    #[case("A-TEFee","one sided A",      -100, -1)]
-    #[case("A-TEFee", "one sided B", 1, 100)]
+    #[case("A-B",    "equally centered", -100, 100, WhirlpoolDeployment::mainnet())]
```
jshiohaha, line 901: small nit: i found it pretty hard to just spot check the same tests exist for various deployments. maybe it's easier to read if the same parameters with different deployments are paired together?

wdyt?

```rust
#[case("A-B",    "equally centered", -100, 100, WhirlpoolDeployment::mainnet())]
#[case("A-B",    "equally centered", -100, 100, WhirlpoolDeployment::mainnet_immutable())]
```

rust-sdk/whirlpool/src/tests/program.rs
```diff
@@ -51,21 +50,22 @@ pub async fn init_tick_arrays_for_range(
...
+            let mut instruction = InitializeTickArray {
+                whirlpool,
+                funder: ctx.signer.pubkey(),
+                tick_array: tick_array_addr,
+                system_program: solana_system_interface::program::id(),
+            }
+            .instruction(InitializeTickArrayInstructionArgs {
+                start_tick_index: current,
+            });
+            instruction.program_id = whirlpool_deployment.id();
```
jshiohaha, line 67: all these repeated `.program_id` setters (+ how ergonomic the ts interface is) had me wondering if we could make a helper. then, i started thinking we could possibly do by injecting an extra function via codama. maybe smth like this

```rust
.instruction_with_program_id(InitializeTickArrayInstructionArgs {
    start_tick_index: current,
}, whirlpool_deployment.id());
```

under the hood, it'd probably do smth similar

```rust
pub fn instruction_with_program_id(&self, args: CloseBundledPositionInstructionArgs, program_id: solana_pubkey::Pubkey) -> solana_instruction::Instruction {
  let mut ix = self.instruction_with_remaining_accounts(args, &[]);
  ix.program_id = program_id;
  ix
}
```

no change requested for now 👍 

i think we can do stuff like this in the future though since its an internal optimization and we just want to get the updates out.

rust-sdk/client/codama.js
```diff
@@ -1,17 +1,46 @@
...
+const ownerImplAnnotations = [
+  "/// Returns the mutable Whirlpool program ID.",
+  "///",
+  "/// Using this with an account owned by the immutable Whirlpool program will",
+  "/// cause anchor's owner check to reject a valid account. Prefer fetching",
+  "/// the account via RPC and reading its `owner` field directly.",
+  '#[deprecated(note = "returns mutable Whirlpool program ID only")]',
+];
+
+const patched = mapRenderMapContent(renderMap, (content) =>
```
jshiohaha, line 37: i think there's more codama native ways to handle these updates, but we can punt for now, as long as this generates the functionality we need 

example `renderVisitor` function https://github.com/amilz/kit-helpers/blob/main/renderers/js-docs/src/visitors/renderVisitor.ts

explicitly, no change requested for now 👍

rust-sdk/whirlpool/src/decrease_liquidity.rs
```diff
@@ -44,6 +45,12 @@ pub enum DecreaseLiquidityParam {
     Liquidity(u128),
 }
 
+impl Default for DecreaseLiquidityParam {
+    fn default() -> Self {
+        DecreaseLiquidityParam::Liquidity(0)
```
jshiohaha: so now that i'm looking at this a little more, we might want to pull DecreaseLiquidityParam out into a parameter. the default value of 0 will always cause an error. and i'm not sure there's a reasonable default we can set - Liquidity(1) might still be too small and setting token a/b is probably not right either

https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+%22data.liquidity_amount+%3D%3D+0%22&type=code

rust-sdk/whirlpool/src/decrease_liquidity.rs
jshiohaha: ok after this comment https://github.com/orca-so/whirlpools/pull/1301/changes#r3226863420., i'm reviving this. maybe we shouldn't put variables that have a default that don't make sense, i.e. `Pubkey::default()` will never be a valid parameter to this function

and allowing the caller to set invalid values with `Default::default()` feels like we're actively leading them in the wrong direction

sorry for flip flopping here, just trying to think from the pov of the sdk consumers 😅 

wdyt

rust-sdk/whirlpool/src/increase_liquidity.rs
```diff
@@ -1473,34 +1696,43 @@ mod tests {
...
-
-            Ok(())
         }
 
+        #[rstest]
+        #[case(WhirlpoolDeployment::mainnet())]
+        #[case(WhirlpoolDeployment::mainnet_immutable())]
         #[tokio::test]
         #[serial]
-        async fn fails_if_lower_tick_gte_upper_tick() -> Result<(), Box<dyn Error>> {
```
jshiohaha, line 1482: maybe dumb q, but what was the driver for this change?

rust-sdk/whirlpool/src/pool.rs
```diff
@@ -495,15 +538,24 @@ mod tests {
         }
     }
 
+    // Only parameterized for the mutable mainnet program. `fetch_whirlpools_by_token_pair` calls
+    // `fetch_all_fee_tier_with_filter`, whose underlying GPA helper in the client crate is hardcoded
```
jshiohaha: maybe dumb q, but why can't we thread program id here too?

jshiohaha (commented):
i think we're getting really close, a few more comments + some nits

rust-sdk/client/src/config.rs
crnorthc: I will remove, they were mostly for sanity checks to make sure I got the arrays all correct

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Makes sense to me!

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Yeah, that makes sense to me. I'll go through and clean these up

rust-sdk/whirlpool/src/increase_liquidity.rs
crnorthc: Ahh, I did a couple tests by hand (all of which had no return) and told Claude to copy the patterns across the rest and it seems like it included the no return in the pattern I wanted. Will revert.

rust-sdk/whirlpool/src/pool.rs
crnorthc: Ooo, that is one spot I left the hardcoded WHIRLPOOL_ID that Claude did not catch. Great callout!

rust-sdk/whirlpool/src/decrease_liquidity.rs
jshiohaha: nit: can remove now?

rust-sdk/whirlpool/src/pool.rs
jshiohaha: can add immutable now + delete comment? 👀

(same for other instances of this comment, not all are showing up in diff)

jshiohaha (approved):
ok, 2 small things left to consider, but i think changes look good now

great work on this 🏄

rust-sdk/whirlpool/src/decrease_liquidity.rs
crnorthc: Yes, good catch

rust-sdk/whirlpool/src/pool.rs
crnorthc: Yes, good catch

PR #1305, opened 2026-05-08
TypeScript SDK updates for immutable whirlpools

crnorthc:
## Add dynamic program ID support to TypeScript SDK

This is the TypeScript SDK counterpart to #1301, which added dynamic program_id support to the Rust
SDK. The mutable global config pattern (`WHIRLPOOLS_CONFIG_ADDRESS` / `setWhirlpoolsConfig`) is replaced
with a `WhirlpoolDeployment` value passed per-call, making it safe to use multiple deployments in the
same process and eliminating async initialization.

### Breaking changes

@orca-so/whirlpools-client
- New `WhirlpoolDeployment` type (and named constants .mainnet, .devnet, .mainnetImmutable,
.custom(programId, configAddress)) introduced in config.ts.
- All PDA helpers now accept a programId (or `WhirlpoolDeployment`) instead of deriving it from a
global. The positional `whirlpoolsConfig` argument is gone.
- All GPA helpers updated to accept a `programId` parameter.

@orca-so/whirlpools
- `setWhirlpoolsConfig` / `WHIRLPOOLS_CONFIG_ADDRESS` / `DEFAULT_WHIRLPOOLS_CONFIG_ADDRESSES` /
`WHIRLPOOLS_CONFIG_EXTENSION_ADDRESS` are removed.
- `WhirlpoolDeployment` and `DEFAULT_WHIRLPOOL_DEPLOYMENT` are re-exported from
@orca-so/whirlpools-client.
- All SDK functions (`createPool`, `increaseLiquidity`, `decreaseLiquidity`, `harvest`, `swap`, etc.) now accept
  a deployment field in their …Config option objects rather than reading from globals.
- `resetConfiguration` is now synchronous (no longer needs to reset the config address).

### Migration

```typescript
// Before
await setWhirlpoolsConfig("solanaMainnet");
const result = await createPool(rpc, ...args);

// After
const result = await createPool(rpc, ...args, {
  deployment: WhirlpoolDeployment.mainnet,
});
// or omit `deployment` to use DEFAULT_WHIRLPOOL_DEPLOYMENT (mainnet)
```

### What's included

- New `WhirlpoolDeployment` type, named constants, and `DEFAULT_WHIRLPOOL_DEPLOYMENT` in
@orca-so/whirlpools-client
- Updated PDA and GPA helpers across ts-sdk/client
- All @orca-so/whirlpools functions updated to thread deployment through option objects
- Full test suite updated (unit + integration + e2e)
- Documentation updated across SDK guides
- Changesets for major version bumps on @orca-so/whirlpools and @orca-so/whirlpools-client


kaizen-orca:
I think there’s one functional regression in the immutable-program migration.

Several position-by-mint flows now accept an optional `programId` and correctly use it when building the instruction / PDA side (`increaseLiquidityInstructions`, `decreaseLiquidityInstructions`, `harvestPositionInstructions`, `resetPositionRangeInstructions`, etc.). But the tests that verify the post-state still call `getPositionAddress(positionMintAddress)` *without* threading that same program ID through — for example in `ts-sdk/whirlpool/tests/decreaseLiquidity.test.ts`, `closePosition.test.ts`, and `resetPositionRange.test.ts`.

That means the current coverage is still validating the canonical mutable-program PDA even when the operation under test is meant to support alternate Whirlpool programs. For immutable/local/custom targets, `getPositionAddress(positionMintAddress, programId)` derives a different address, so these tests can still pass while the actual target-program path remains unverified.

I’d tighten those assertions to derive the expected position address with the same target program/program id used by the instruction path, otherwise this PR can claim immutable support without actually testing the core position-PDA split it introduces.

.changeset/swift-otters-target.md
```diff
@@ -0,0 +1,6 @@
+---
+"@orca-so/whirlpools": major
+"@orca-so/whirlpools-client": major
```
jshiohaha: i think we should technically include `@orca-so/whirlpools-example-ts-next` in the changeset as well

ts-sdk/whirlpool/src/config.ts
```diff
@@ -199,12 +150,12 @@ export function setEnforceTokenBalanceCheck(
 /**
  * Resets the configuration to its default state.
  *
- * @returns {Promise<void>} - Resolves when the configuration has been reset.
+ * Resets the funder, slippage tolerance, native mint wrapping strategy, and
+ * enforce-token-balance-check globals. The Whirlpools program / config
+ * targeting is no longer global — pass a `TargetProgram` to SDK functions
```
jshiohaha: nit nit: maybe it's best to only comment on the current state of functionality? right now, it's introducing a little bit of historical context as well

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -110,15 +109,15 @@ export function createSplashPoolInstructions(
  * @param {Address} tokenMintB - The second token mint address to include in the pool.
  * @param {number} tickSpacing - The spacing between price ticks for the pool.
  * @param {number} [initialPrice=1] - The initial price of token 1 in terms of token 2.
+ * @param {TargetProgram} [targetProgram=TargetProgram.default()] - The Whirlpool program/config to target. Defaults to the mutable mainnet deployment.
```
jshiohaha: nit: might leave this part out because this is another point we have to change if the functionality behind `default()` changes. if @param links to TargetProgram and its comment, then that's as useful (imo)

```ts
 Defaults to the mutable mainnet deployment.
```

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -179,39 +180,38 @@ export async function createConcentratedLiquidityPoolInstructions(
...
+        whirlpool: poolAddress,
+        tokenVaultA,
+        tokenVaultB,
+        tokenProgramA,
+        tokenProgramB,
+        feeTier,
+        tickSpacing,
+        initialSqrtPrice,
+      },
+      { programAddress: targetProgram.programId },
```
jshiohaha: 🔥

ts-sdk/whirlpool/src/harvest.ts
```diff
@@ -69,29 +70,13 @@ export type HarvestPositionInstructions = {
...
  * @param {Address} positionMintAddress
  *    The position mint address you want to harvest fees and rewards from.
+ * @param {Address} [programId=WHIRLPOOL_PROGRAM_ADDRESS]
+ *    The Whirlpool program ID to target. Defaults to the canonical mainnet program.
  * @param {TransactionSigner} [authority=FUNDER]
  *    The account that authorizes the transaction. Defaults to a predefined funder.
  *
  * @returns {Promise<HarvestPositionInstructions>}
  *    A promise that resolves to an object containing the instructions, fees, and rewards quotes.
- * @example
```
jshiohaha, line 77: was this rm on purpose?

(same comment for other instances)

ts-sdk/whirlpool/src/harvest.ts
```diff
@@ -299,6 +301,7 @@ export async function harvestAllPositionFees(): Promise<Signature[]> {
       const { instructions } = await harvestPositionInstructions(
         rpc,
         position.data.positionMint,
+        undefined,
```
jshiohaha: should this be undefined? or a value we thread?

ts-sdk/whirlpool/tests/utils/mockRpc.ts
```diff
@@ -32,6 +36,13 @@ import { getTokenSize } from "@solana-program/token-2022";
 export const signer = getNextKeypair();
 setDefaultFunder(signer);
 
+/**
+ * The TargetProgram used by every test in this suite, pointing at the
+ * locally-initialized WhirlpoolsConfig. Populated lazily by
+ * {@link getTestContext} on first use.
+ */
+export let currentTargetProgram: TargetProgram = TargetProgram.default();
```
jshiohaha: since the new TargetProgram is optional, it's pretty difficult to figure out if we updated all the required call sites. it would be awesome if we could thread TargetProgram more thoroughly in tests. it looks like it's broadly added where required `get_<account>_address` helpers, but not in instruction/tx builders

e.g. it looks like we missed `createConcentratedLiquidityPoolInstructions` in `ts-sdk/whirlpool/tests/createPool.test.ts`

ts-sdk/whirlpool/tests/createPool.test.ts
```diff
@@ -183,7 +183,7 @@ describe("Create Pool", () => {
         mintA,
         mintB,
         tickSpacing,
-        price,
+        price, currentTargetProgram,
```
jshiohaha: nit: maybe move currentTargetProgram to a newline? unless autoformatter did this

ts-sdk/whirlpool/src/createPool.ts
jshiohaha, line 154: i know typescript is a little bit different, but i this appending more parameters still feels suboptimal. now at call sites, i have to figure out what the order of parameters is and which map to undefined values. i think we could probably benefit from the config based parameter structure here as well 🤷 

curious how others feel?

examples/ts-sdk/next/app/page.tsx
```diff
@@ -64,6 +64,7 @@ function SwapPage({ account }: SwapPageProps) {
       },
       POOL_ADDRESS,
       100,
+      undefined,
```
jshiohaha: are we falling back to the mainnet instance here on purpose? i see devnet config mentioned below, but i'm not familiar with this example app

kaizen-orca:
Re-checked the latest version after `cc4ee6d0fb1252d0f8e80f24eeba1fd84999cf10`, and I still have the same coverage concern.

The instruction paths are threading `whirlpoolDeployment.programId` into `getPositionAddress(...)` / `getTickArrayAddress(...)` correctly in places like `ts-sdk/whirlpool/src/decreaseLiquidity.ts` and `resetPositionRange.ts`, but the tests I called out are still verifying post-state with the default-program PDA:

- `ts-sdk/whirlpool/tests/decreaseLiquidity.test.ts`
- `ts-sdk/whirlpool/tests/closePosition.test.ts`
- `ts-sdk/whirlpool/tests/resetPositionRange.test.ts`

Those files still do things like `await getPositionAddress(positionMintAddress)` without passing the deployment/program under test.

So for immutable/local/custom deployments, the operation code and the assertion code can still be talking about different position PDAs, which means this PR can still pass while the alternate-program position-by-mint path remains unverified. I think this still needs to be tightened before the immutable-program support is really covered end to end.

ts-sdk/whirlpool/tests/utils/mockRpc.ts
```diff
@@ -64,15 +91,24 @@ export async function getTestContext(): Promise<LiteSVM> {
...
     );
 
+    // Load immutable whirlpool program (used to parametrize tests against both deployments)
+    const whirlpoolImmutableProgram = fs.readFileSync(
+      path.join(process.cwd(), "../../rust-sdk/whirlpool/src/tests/fixtures/whirlpool_immutable.so"),
+    );
+    _testContext.addProgram(
+      toPublicKey(WhirlpoolDeployment.mainnetImmutable.programId),
+      whirlpoolImmutableProgram,
+    );
```
crnorthc, line 110: Not sure if this is the best approach. Could move the fixture to a higher level or create a copy in this dir

ts-sdk/client/src/pda/feeTier.ts
```diff
@@ -1,20 +1,26 @@
...
   getAddressEncoder,
   getProgramDerivedAddress,
   getU16Encoder,
 } from "@solana/kit";
-import { WHIRLPOOL_PROGRAM_ADDRESS } from "../generated/programs/whirlpool";
+import type { WhirlpoolDeployment } from "../config";
+import { DEFAULT_WHIRLPOOL_DEPLOYMENT } from "../config";
 
+/**
+ * Derives the fee tier PDA for the given fee tier index under the supplied target program.
```
jshiohaha: nit: supplied target program -> supplied whirlpools deployment? maybe left over from previous naming convention 😅

ts-sdk/client/tests/pda.test.ts
```diff
@@ -33,97 +31,214 @@ const TEST_NATIVE_MINT_ADDRESS = address(
   "So11111111111111111111111111111111111111112",
 );
 
-describe("derive program accounts", () => {
```
jshiohaha, line 36: nit: similar to rstest (and other test class in ts where you iterate through cases), i think we could probably run the same tests with different inputs instead of duplicating. not super pressing, moreso a nice to have

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -277,9 +310,32 @@ export async function createConcentratedLiquidityPoolInstructions(
...
-);
-export const createConcentratedLiquidityPool = wrapFunctionWithExecution(
-  createConcentratedLiquidityPoolInstructions,
-);
+export function createSplashPool(
+  tokenMintA: Address,
+  tokenMintB: Address,
+  config?: Omit<CreateSplashPoolConfig, "funder">,
+) {
+  return executeWithCallback((rpc, owner) =>
```
jshiohaha: nit: is funder more accurate than owner here?

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -277,9 +310,32 @@ export async function createConcentratedLiquidityPoolInstructions(
...
+    createSplashPoolInstructions(rpc, tokenMintA, tokenMintB, {
+      ...config,
+      funder: owner,
+    }),
+  );
+}
+
+export function createConcentratedLiquidityPool(
```
jshiohaha, line 309: it looks like there might be some other places to update, but i think we can punt non-code changes to a later PR https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+%22createConcentratedLiquidityPool%22+language%3AMDX&type=code&l=MDX

just need to track the work

ts-sdk/whirlpool/src/createPool.ts
```diff
@@ -277,9 +310,32 @@ export async function createConcentratedLiquidityPoolInstructions(
...
-export const createSplashPool = wrapFunctionWithExecution(
-  createSplashPoolInstructions,
-);
-export const createConcentratedLiquidityPool = wrapFunctionWithExecution(
-  createConcentratedLiquidityPoolInstructions,
-);
+export function createSplashPool(
+  tokenMintA: Address,
+  tokenMintB: Address,
+  config?: Omit<CreateSplashPoolConfig, "funder">,
```
jshiohaha: kind of interesting that we don't allow propagating the funder and must set payer through global setPayer function. it might not matter as much here, but if there's an existing authority we have to match, the fact that the user must call setPayer first seems a little unintuitive 🤔 

thoughts on updating?

ts-sdk/whirlpool/tests/utils/mockRpc.ts
jshiohaha: totally forgot when we were updating rust-sdk, but there's an existing process for loading external programs. external program binaries are all stored in a specific directory and copied into `target/deploy` at test time so that we don't link from random directories/have to maintain multiple copies

https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+external_program&type=code 

feel free to keep as is for this PR and maybe we can change in a small follow up PR

ts-sdk/whirlpool/tests/closePosition.test.ts
```diff
@@ -9,7 +9,12 @@ import assert from "assert";
 import { beforeAll, describe, it } from "vitest";
 import { closePositionInstructions } from "../src/decreaseLiquidity";
 import { swapInstructions } from "../src/swap";
-import { rpc, sendTransaction } from "./utils/mockRpc";
```
jshiohaha, line 12: these diffs are pretty hard to read, so i'm just going to assume that most logic is still the same. can you call smth out if it changed significantly?

ts-sdk/whirlpool/src/createPool.ts
jshiohaha: ah just saw this one, nice! https://github.com/orca-so/whirlpools/pull/1306

ts-sdk/whirlpool/src/createPool.ts
crnorthc: That makes sense, will switch

ts-sdk/whirlpool/src/createPool.ts
crnorthc: Yeah, that makes sense to me!

ts-sdk/whirlpool/tests/closePosition.test.ts
crnorthc: Yeah, not a lot of significant logic changes. Just a lot or re-arranging to best fit the mutable and immutable cases. The main logic changes are just to pipe through programId into the helpers that had the hardcoded mutable id and some changes in getTestContext for the immutable program deployment

ts-sdk/whirlpool/src/actionHelpers.ts
```diff
@@ -19,28 +19,30 @@ import {
...
+>(build: (rpc: Rpc<SolanaRpcApi>) => Promise<R>): Promise<ActionResult<R>> {
   const { rpcUrl } = getRpcConfig();
   const rpc = rpcFromUrl(rpcUrl);
-  const payer = getPayer();
-  const result = await build(rpc, payer);
+  const result = await build(rpc);
   return {
     ...result,
-    callback: () => buildAndSendTransaction(result.instructions, payer),
+    callback: (payer?: KeyPairSigner) => {
```
jshiohaha, line 42: oooh nice, i actually think this was a better update than i initially expected. since we don't actually need the payer to build the instructions, we only ask the caller to provide the keypair that will actually pay for the tx to the callback

🦖

ts-sdk/whirlpool/package.json
```diff
@@ -24,7 +24,7 @@
   ],
   "scripts": {
     "build": "tsup src/index.ts --format cjs,esm --dts --sourcemap",
-    "test": "cp ../../legacy-sdk/whirlpool/tests/external_program/token_2022.20250510.so ../../target/deploy/ && vitest run tests",
+    "test": "cp ../../legacy-sdk/whirlpool/tests/external_program/token_2022.20250510.so ../../target/deploy/ && cp ../../rust-sdk/whirlpool/src/tests/fixtures/whirlpool_immutable.so ../../target/deploy/ && vitest run tests",
```
jshiohaha: ah nice, thanks for updating here! but as long as you're making the change here, thoughts on moving the binary to the same external_program directory as the other programs?

jshiohaha (commented):
1 small last comment while we wait for CI to run (waiting to approve until after those finish). otherwise, i think changes look good!

ts-sdk/whirlpool/package.json
crnorthc: Makes sense to me!

jshiohaha (approved):
🚀

PR #1311, opened 2026-06-05
Remove migrate_repurpose_reward_authority_space ix

yugure-orca:
All whirlpools in Solana mainnet and Solana devnet have been migrated.

- This instruction was explicitly intended to be removed after all pools had been migrated.
- This instruction was designed to operate permissionlessly. It was safe for anyone to call, and any attempt to invoke it on an already migrated pool would terminate with a panic.
- Pools initialized after this instruction was deployed were marked as "migrated" from the outset, making migrate_repurpose_reward_authority_space inapplicable to them.


jshiohaha (approved):
changes look reasonable to me!

should we remove the ix from the idl and re-generate the codama clients before merging?

wjthieme (approved):

PR #1316, opened 2026-06-12
Add support for immutable whirlpools in legacy sdk

crnorthc:
**Title**

  Add immutable whirlpool support to the legacy SDK

  **Details**

  This PR adds support for targeting Orca's immutable (non-upgradable) Whirlpool deployment from the legacy `@orca-so/whirlpools-sdk`.

  Changes:

  - **New constants** in `types/public/constants.ts`:
    - `ORCA_WHIRLPOOL_PROGRAM_ID_IMMUTABLE` — the program ID for the immutable Whirlpool deployment.
    - `ORCA_WHIRLPOOLS_CONFIG_IMMUTABLE` — the `WhirlpoolsConfig` account paired with the immutable program.

  - **Optional `programId` argument** on `WhirlpoolContext.from` and `WhirlpoolContext.withProvider`. When provided, the context is built from the bundled IDL rebound to the given program ID via the new `getWhirlpoolIdl` helper. Because Anchor derives the target program ID from `idl.address`, cloning the IDL with a new address lets the SDK build instructions and PDAs against the immutable program (or any fork) without shipping a separate IDL artifact. When omitted, behavior is unchanged.

  To target the immutable deployment, pass `ORCA_WHIRLPOOL_PROGRAM_ID_IMMUTABLE` as the `programId` and use `ORCA_WHIRLPOOLS_CONFIG_IMMUTABLE` as the config.

  Testing:

  - `tests/sdk/constants.test.ts` — verifies the new constants.
  - `tests/integration/immutable/lifecycle.test.ts` — full lifecycle integration test against the immutable program.
  - `tests/utils/litesvm.ts` — test harness updates to load/run the immutable program.

  Includes a changeset (`minor` bump for `@orca-so/whirlpools-sdk`).


legacy-sdk/whirlpool/src/types/public/constants.ts
```diff
@@ -9,6 +9,20 @@ export const ORCA_WHIRLPOOL_PROGRAM_ID = new PublicKey(
   "whirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc",
 );
 
+/**
+ * Program ID hosting Orca's immutable (non-upgradable) Whirlpool program.
+ *
+ * This is a separate, non-upgradable deployment of the Whirlpool program built from the
+ * same source as {@link ORCA_WHIRLPOOL_PROGRAM_ID}. Because PDA derivation and instruction
```
jshiohaha: nit: this is not true in perpetuity. we probably should mention that it's deployed from the program at commit hash https://github.com/orca-so/whirlpools/commits/b5cb6b8aa69ddbc2c868c3389f2de4f1e2709984

this can be verified via the verifiable build api https://verify.osec.io/status/iwhrLHdsgrvmnwU8GF2FSmyabSMjfHwFGJAX2ufJ3ZN

legacy-sdk/whirlpool/src/types/public/constants.ts
```diff
@@ -17,6 +31,16 @@ export const ORCA_WHIRLPOOLS_CONFIG = new PublicKey(
   "2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ",
 );
 
+/**
+ * Orca's WhirlpoolsConfig PublicKey for the immutable Whirlpool program.
+ *
+ * This config account pairs with {@link ORCA_WHIRLPOOL_PROGRAM_ID_IMMUTABLE}.
+ * @category Constants
+ */
+export const ORCA_WHIRLPOOLS_CONFIG_IMMUTABLE = new PublicKey(
```
jshiohaha, line 41: interesting that we don't use this anywhere else other than tests.. i thought there was going to be more plumbing like the ts-sdk 😅

legacy-sdk/whirlpool/tests/sdk/constants.test.ts
```diff
@@ -0,0 +1,37 @@
+import * as assert from "assert";
```
jshiohaha: nit: i'm not sure how much value is added by testing consts

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          tickArrayLower: positions[0].tickArrayLower,
+          tickArrayUpper: positions[0].tickArrayUpper,
+        }),
+      ).buildAndExecute();
+
+      const positionAfterDecrease = (await fetcher.getPosition(
+        positions[0].publicKey,
+        IGNORE_CACHE,
+      )) as PositionData;
+      assert.ok(positionAfterDecrease.liquidity.lt(liquidityAmount));
```
jshiohaha: should this be 0?

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+      // open a fresh (empty) position and close it
```
jshiohaha: ooc, what does open + close accomplish in addition to the flow above?

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
```diff
@@ -0,0 +1,188 @@
+import * as anchor from "@coral-xyz/anchor";
```
jshiohaha, line 1: at first, i was curious about trying to run the entire test suite against either the mutable or immutable versions. but then i noticed a few hard coded instances of the [mutable program id](https://github.com/search?q=repo%3Aorca-so%2Fwhirlpools+ORCA_WHIRLPOOL_PROGRAM_ID+path%3A%2F%5Elegacy-sdk%5C%2Fwhirlpool%5C%2Ftests%5C%2Fsdk%5C%2Fwhirlpools%2F&type=code) and [preloaded accounts](https://github.com/orca-so/whirlpools/tree/b2fabd377c7f06f62fd546e49a0320b638e4d4fd/legacy-sdk/whirlpool/tests/preload_account). so it's probably more of a headache to get the full test suite working. and ofc, it's possible the tests would break when we inevitably update the mutable program 😬 

and ig the main point is probably to sanity check the sdk updates work for both programs, so i think this test accomplishes that 🤝

jshiohaha (commented):
nice work on this! you were right about this being much simpler than the other SDKs. i left a few comments, so lmk what you think. nothing too big though, so i suspect we can get this merged s00n 🌬️

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
crnorthc: It should be around 9_000_000, because it is only a partial decrease in liquidity. I can adjust to check to see if the new liquidity is within the expected +/- the slippage

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
crnorthc: It tests the close position, however I will refactor to not create a fresh position to close and just close the existing eposition

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
```diff
@@ -51,14 +48,17 @@ describe.each(DEPLOYMENTS)(
       assert.ok(ctx.program.programId.equals(programId));
     });
 
-    it("runs init pool -> open/increase -> swap -> decrease -> close", async () => {
-      // init config + fee tier + pool, open position, and fund liquidity
+    it("runs open position -> add liquidity -> decrease some -> decrease all -> close", async () => {
```
jshiohaha: ooc, what motivated the change here?

crnorthc: It better tests the full flow/tooling. Previously, the position was preloaded into litesvm. Now, it creates the position, adds liquidity, does a partial withdraw, a full withdraw, then closes the position.

jshiohaha: ok, fair enough. and did you get rid of the swap b/c it accrued fees to the position, meaning you had to call collect fees before you could close it?

crnorthc: ooo, that is a good point. That should be tested as well

legacy-sdk/whirlpool/tests/integration/immutable/lifecycle.test.ts
```diff
@@ -48,20 +61,27 @@ describe.each(DEPLOYMENTS)(
       assert.ok(ctx.program.programId.equals(programId));
     });
 
-    it("runs open position -> add liquidity -> decrease some -> decrease all -> close", async () => {
+    it("runs open position -> add liquidity -> swap -> collect fees -> decrease some -> decrease all -> close", async () => {
```
jshiohaha, line 64: no strong preference from me, but feel free to combine into 1 decrease if you think that'd be simpler

jshiohaha (approved):
nice work, changes lgtm 🚀

PR #1317, opened 2026-06-16
Prepare/Commit (aka on-chain quote interface)

yugure-orca:
This PR adds an on-chain quoting interface for Whirlpool using CPI and `get_return_data`.

To avoid simply doubling compute unit consumption, it takes a different approach: the intermediate results produced during quote generation are written to a `PreparedSwap` account, and the execution step simply applies those precomputed results. As a result, the overhead remains relatively low (around 15%).

Since Orca initializes the `PreparedSwap` account, integrators can use the following flow:

**If the quote is acceptable:**
* `prepare_swap_v2` CPI (get and evaluate a quote)
* `commit_swap_v2` CPI (execute the trade)

**If the quote is not acceptable:**
* `prepare_swap_v2` CPI (get and evaluate a quote)
* no further CPI calls

No cleanup or "rollback" step is required—the caller can simply discard the result.

### Expected use cases
* On-chain routing
* On-chain arb trade
* Building multi-hop ExactOut trades
* Obtaining quotes via RPC simulation (allowing developers to cover the most complex part of swap logic using only the Solana SDK, without needing a Whirlpool-specific SDK)

### Deployment note
We should merge https://github.com/orca-so/whirlpools/pull/1311 first, because we use "extension" field.
Note: we've already executed "migration" on all whirlpools, so we are safe even if we miss PR1311.


programs/whirlpool/src/instructions/adaptive_fee/set_adaptive_fee_constants.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
 
 #[derive(Accounts)]
 pub struct SetAdaptiveFeeConstants<'info> {
-    #[account(has_one = whirlpools_config)]
+    #[account(mut, has_one = whirlpools_config)]
```
yugure-orca, line 10: This is  a small breaking, but this instruction is only for the protocol, so there is no user impact.

programs/whirlpool/src/lib.rs
```diff
@@ -1328,10 +1328,100 @@ pub mod whirlpool {
...
+    /// Any user may initialize a PreparedSwap account. The resulting account state is identical
+    /// regardless of who performs the initialization; the caller only determines who pays the rent required to create the account.
+    ///
+    /// ### Parameters
+    /// - `nonce` - Nonce used to derive a unique PreparedSwap PDA.
+    ///             Must be less than or equal to MAX_PREPARED_SWAP_NONCE.
+    ///
+    /// #### Special Errors
+    /// - `PreparedSwapNonceMaxExceeded` - if the provided nonce exceeds MAX_PREPARED_SWAP_NONCE.
+    pub fn initialize_prepared_swap(
```
jshiohaha, line 1342: assuming that CU efficiency is important here, i'm curious why you chose to add via the anchor entrypoint over pinocchio?

programs/whirlpool/src/state/prepared_swap.rs
```diff
@@ -0,0 +1,1066 @@
...
+        amount_specified_is_input: bool,
+        a_to_b: bool,
+        slot: u64,
+    ) -> Result<()> {
+        // PreparedSwap is a zero-copy account.
+        // A version mismatch means the account data may be interpreted using the wrong
+        // layout, so the version must be checked before performing any other validation.
+        if self.version != PREPARED_SWAP_LAYOUT_VERSION {
+            return Err(ErrorCode::PreparedSwapVersionMismatch.into());
+        }
```
jshiohaha, line 294: it's not possible to have a version mismatch in a PreparedSwap account because the state should be reset and committed in the same slot? aka this flow is not possible:

- create PreparedSwap account
- update program version
- commit swap from previously set PreparedSwap account

programs/whirlpool/src/instructions/v2/commit_swap.rs
```diff
@@ -0,0 +1,238 @@
...
+        (&ctx.accounts.token_mint_b, &ctx.accounts.token_mint_a)
+    };
+    let input_transfer_fee =
+        calculate_transfer_fee_excluded_amount(token_mint_input, input_amount)?.transfer_fee;
+    let output_transfer_fee =
+        calculate_transfer_fee_excluded_amount(token_mint_output, output_amount)?.transfer_fee;
+    let (lp_fee, protocol_fee) = (swap_update.lp_fee, swap_update.next_protocol_fee);
+
+    // apply pending whirlpool update
+    update_and_swap_whirlpool_v2(
```
jshiohaha, line 245: very convenient that `swap_with_transfer_fee_extension` and this fn already exist and you don't have to do any major refactors 🙇

programs/whirlpool/src/state/prepared_swap.rs
```diff
@@ -0,0 +1,1066 @@
...
+        self.pending_updates.pending_tick_updates[self.pending_updates.pending_tick_updates_len as usize] = update;
+        self.pending_updates.pending_tick_updates_len += 1;
+    }
+
+    pub fn set_pending_post_swap_update(
+        &mut self,
+        update: &PostSwapUpdate,
+    ) {
+        self.pending_updates.pending_post_swap_update.amount_a = update.amount_a;
```
jshiohaha: nit: could reduce some verbosity with smth like

```rust
let dst = &mut self.pending_updates.pending_post_swap_update;
dst.amount_a = update.amount_a;
...
```

programs/whirlpool/src/instructions/v2/prepare_swap.rs
```diff
@@ -0,0 +1,206 @@
...
+    util::{
+        parse_remaining_accounts, to_timestamp_u64, AccountsType,
+        RemainingAccountsInfo, SparseSwapTickSequenceBuilder,
+    },
+};
+
+#[derive(Accounts)]
+pub struct PrepareSwapV2<'info> {
+    #[account(mut)]
+    pub prepared_swap: AccountLoader<'info, PreparedSwap>,
```
jshiohaha, line 19: am i correct in understanding that collisions in prepared_swap accounts in the same block are not an issue? aka it's possible to include multiple swaps with the same prepared_swap account, with the main downsides being lower chance of same block inclusion due to scheduling write-locked account?

but in either case, another prepare + commit swap in the same block could re-use the same PreparedSwap PDA

1. someone calls prepare_swap **without** commit_swap: we generate the quote, store in PreparedSwap PDA, but pending updates are never committed, no problem
2. someone calls prepare_swap **with** commit_swap: we generate the quote, store in PreparedSwap PDA, and pending updates are committed, expected full path

programs/whirlpool/src/instructions/v2/commit_swap.rs
```diff
@@ -0,0 +1,238 @@
...
+        ctx.remaining_accounts,
+        &remaining_accounts_info,
+        &[
+            AccountsType::TransferHookA,
+            AccountsType::TransferHookB,
+            AccountsType::SupplementalTickArrays,
+        ],
+    )?;
+
+    let swap_tick_sequence_builder = SparseSwapTickSequenceBuilder::new(
```
jshiohaha, line 98: the prepare (slot, whirlpool, state sequence) == commit (slot, whirlpool, state sequence) check is so much cleaner than other versions of the pre/post instruction pairs i've seen 🧼 

i was trying to think about what other non-whirlpool state could present an attack vector b/w prepare and commit, but the same-slot requirement seems to greatly reduce surface area, e.g. token 2022 transfer fee change since an update is only applied on epoch+2 ([source](https://github.com/solana-program/token-2022/blob/0f9c09f9d7bf12d4cedc536e1640f5e3156b3481/program/src/extension/transfer_fee/processor.rs#L105-L111)), etc.

i thought about transfer hooks, but presumably our token badge process helps verify possibly malicious hooks before allowing tokens in to a pool

programs/whirlpool/src/instructions/v2/prepare_swap.rs
jshiohaha: do you think it's worth acknowledging that someone could try to ddos by spamming prepare_swap transactions and locking all 16 prepared_swap accounts? it seems difficult to do with no/low economic incentive, but it might be something an (ai) vulnerability report would flag 😬

legacy-sdk/whirlpool/src/instructions/v2/prepare-swap-ix.ts
```diff
@@ -0,0 +1,88 @@
...
+} from "../../utils/remaining-accounts-util";
+import { CommitSwapV2Params } from "./commit-swap-ix";
+
+/**
+ * Raw parameters and accounts to prepare a swap on a Whirlpool.
+ *
+ * @category Instruction Types
+ * @param CommitSwapV2Params - Parameters in {@link CommitSwapV2Params}
+ */
+export type PrepareSwapV2Params = Omit<CommitSwapV2Params, "tokenProgramA" | "tokenProgramB" | "tokenOwnerAccountA" | "tokenVaultA" | "tokenOwnerAccountB" | "tokenVaultB" | "tokenTransferHookAccountsA" | "tokenTransferHookAccountsB">;
```
jshiohaha: nit: small downside of coupling is if CommitSwapV2Params diverges from PrepareSwapV2Params in the future, it would implicitly inherit requirements. ofc, we can just change types + i doubt that many people will consume these, so likely not a big deal 😅

legacy-sdk/whirlpool/tests/integration/multi-ix/prepare_commit_swap.test.ts
```diff
@@ -0,0 +1,668 @@
...
+      // tickarray
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, swapQuote.tickArray0, getAccountSize(AccountName.DynamicTickArray));
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, swapQuote.tickArray1, getAccountSize(AccountName.DynamicTickArray));
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, swapQuote.tickArray2, getAccountSize(AccountName.DynamicTickArray));
+
+      // token accounts
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, poolInfo.tokenAccountA, ACCOUNT_SIZE);
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, poolInfo.tokenAccountB, ACCOUNT_SIZE);
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, pool.getData().tokenVaultA, ACCOUNT_SIZE);
+      assertPostWritableAccountMatch(prepareAndCommitSimResult, swapV2SimResult, pool.getData().tokenVaultB, ACCOUNT_SIZE);
```
jshiohaha: ooh, these are nice checks 👌

programs/whirlpool/src/state/prepared_swap.rs
```diff
@@ -0,0 +1,1066 @@
...
+    pub lp_fee: u64,
+    pub next_liquidity: u128,
+    pub next_tick_index: i32,
+    pub next_sqrt_price: u128,
+    pub next_fee_growth_global: u128,
+    pub next_reward_growth_global: [u128; NUM_REWARDS],
+    pub next_protocol_fee: u64, // delta value (not next absolute value)
+
+    // Flattened fixed-size Option<T> for zero-copy serialization.
+    pub next_adaptive_fee_variables_is_some: bool,
```
jshiohaha, line 54: in the spirit of front-running some ai generated vulnerabilities - i asked an llm to come up with a vulnerability, and it called out the use of `#[zero_copy(unsafe)]` with `bool`: 

> Arbitrary byte values are not valid Rust bool representations. The layout-version check occurs only after the account has already been cast to this structure, so it cannot safely protect incompatible or corrupted layouts.

i don't think it becomes an issue unless we later update account layout and fail to parse correctly? 

that being said, i think we could migrate uses of `#[zero_copy(unsafe)]` to `#[zero_copy]` if we wanted. afaict, we can apply `#[zero_copy]`to existing structs like `AdaptiveFeeVariables` w/o issue since all fields implement both

```bash
bytemuck::Pod
bytemuck::Zeroable
```

programs/whirlpool/src/lib.rs
```diff
@@ -1328,10 +1328,100 @@ pub mod whirlpool {
...
+    /// ### Return Data
+    /// Serialized `PrepareSwapV2ReturnData`.
+    ///
+    /// #### Special Errors
+    /// Most swap-related errors are returned as PrepareSwapV2ReturnData::QuoteError
+    /// rather than causing the transaction to fail.
+    pub fn prepare_swap_v2<'info>(
+        ctx: Context<'_, '_, '_, 'info, PrepareSwapV2<'info>>,
+        amount: u64,
+        sqrt_price_limit: u128,
```
jshiohaha, line 1368: similar to other swap ix's, should we add `other_amount_threshold` to check the amount generated by swap update?

jshiohaha (commented):
nice work! the prototype is quite cool. i failed to find any DART related documentation for this, it seems like this is totally new proposed flow? is the intended flow smth like the following

1. the aggregator determines candidate routes off-chain,
2. the aggregator constructs a transaction that invokes DART,
3. the on-chain router requests quotes from the available venues and dynamically adjusts the route based on execution-time quotes and realized fills?

i tried executing a few DART swaps to see what the transactions looked like, and i am shocked by how many accounts are being piped in.. 133 accounts passed into swap_route_v3 in transaction [23EQ...twdz](https://solscan.io/tx/23EQw9mwgqaH8CVhyubpDdjkad88zHmVSVoWfJ8353AAyFGTjwhAH9sNiEfvaEZ8cmhrC7dJwZCantij2Mbvtwdz) 🤯

programs/whirlpool/src/lib.rs
yugure-orca: initialize_prepared_swap is initialized by Orca at its own expense (covering the rent) after a program upgrade, and users simply share and use it. It is called in advance and is not included in the swap transaction itself.

As a result, initialize_prepared_swap is not something we need to be concerned about from a CU (compute unit) perspective. Therefore, we're using Anchor.

For prepare_swap_v2 and commit_swap_v2, it's absolutely true that CU efficiency is important, and there's no doubt that a Pinocchio-based implementation would be ideal.

However, since swap_v2 itself is still using Anchor, those instructions are also implemented in Anchor to make releases and audit easier.

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: Because commits are required to occur within the same slot, the only problematic flow is the following. All of these transactions would need to be executed in the same slot:

```
tx1: prepare_swap_v2
tx2: program upgrade including a layout change
tx3: commit_swap_v2
```

As I understand it, once a program upgrade occurs, no further transactions using that program can be executed in the same slot. Any attempt to do so will fail.

Therefore, unless this behavior changes in the future, I don't think the above sequence of transactions can all be executed within the same block. However, if Solana were ever improved such that a program becomes immediately usable after an upgrade, this could become problematic, so we're taking a defensive approach.

Also, whether execution occurs within the same slot is determined by the slot stored in the Precondition. Since we may change the placement or contents of the Precondition in the future, I believe the safest approach is to reject any account state that follows a previous layout altogether.

The fact that we're being defensive here is briefly reflected in the field comment below.
```
    // PreparedSwap account layout version
    // Guard against stale PreparedSwap accounts being used after a program upgrade that modifies the PreparedSwap layout or size.
    // Although instructions are not expected to execute in the same slot after a program upgrade,
    // this version check is included as a safeguard against future changes.
    pub version: u16,
```

programs/whirlpool/src/instructions/v2/prepare_swap.rs
yugure-orca: Integration tests should cover both usage across multiple transactions and multiple uses within a single transaction.

Your understanding is correct: a PreparedSwap account is designed to be reusable multiple times within the same block. As you pointed out, the practical upper bound is likely the shared write-lock CU limit (12M CU).

One reason I haven't been too concerned about this model is that even Jito Tip accounts currently define only eight accounts in total. That said, while writing this, it occurred to me that it wouldn't be surprising if Jito validators had some special protection against transactions that intentionally mention and write-lock multiple Jito Tip accounts. So I'm somewhat less optimistic about the comparison than I was initially. 😅

From an attacker's perspective, however, continuously write-locking popular pools would likely have a very similar effect. The PreparedSwap seeds currently allow up to 256 PDA definitions, and it might make sense to leave room for increasing that number in the future.

**As you suggested, I think it's worth documenting that we considered this attack vector and concluded that it is not materially different from locking the pools themselves. Leaving a comment to that effect would be helpful.**

programs/whirlpool/src/instructions/v2/commit_swap.rs
yugure-orca: Thanks for checking this.

We want to enforce that prepare and commit are executed atomically, but since these instructions are intended to be callable via CPI, checking SysvarInstructions is not an appropriate solution. As a result, the invariant we enforce is that the relevant state must not change within the same slot.

This naturally raises the following attack vectors:

- Modify the state of a Whirlpool, TickArray, or Oracle before calling commit.
- Modify the state of a token account before calling commit.
- Modify the state of a token extension before calling commit.

For Whirlpools, TickArrays, and Oracles, the state sequence check provides protection.

Token account balances do not affect the swap math (this is not a CPMM), so changes there are not a concern.

Token Extensions are somewhat more interesting, but only TransferFee extension affects swap math.

- TransferFee: The fee can only change on an epoch boundary, so it cannot change midway through a slot.
- InterestBearing and ScaledUiAmount: Whirlpool only operates on raw token amounts, so changes in the conversion rate are irrelevant.
- TransferHook: A flow such as prepare -> transfer hook configuration change -> commit is theoretically possible. However, a transfer hook cannot alter the amount of tokens being transferred. Even if a transfer hook change causes the trade to fail, that does not affect the interpretation that swapV2 was executed at commit time.

Since the only token extension currently incorporated into the swap math is TransferFee, and TransferFee cannot change during a slot, this appears to be safe.

programs/whirlpool/src/lib.rs
yugure-orca: There are three reasons why `other_amount_threshold` was removed.
WDYT ?

- `other_amount_threshold` is necessary to protect users from slippage. However, prepare/commit is designed to be executed atomically via CPI, so slippage errors cannot occur by construction.
- If the caller is an on-chain program that can evaluate the result itself (such as an aggregator), `other_amount_threshold` is typically set to `0` (ExactIn case) and the validation is performed by the caller. Since prepare returns the quote as return data, the caller can evaluate it and decide whether to invoke commit.
- I want prepare and commit to take as similar a set of parameters as possible, but requiring `other_amount_threshold` in commit does not make much sense. Whether the threshold is satisfied is already known before commit is invoked. If the threshold is not met, the caller can simply choose not to call commit at all.

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: ## Context
IIRC, there was a breaking change around zero_copy when upgrading to Anchor 0.29. If we wanted to preserve the behavior of `repr(packed)`, we had to switch from `zero_copy` to `zero_copy(unsafe)`. In Whirlpool, TickArray used `repr(packed)`, so we kept compatibility by using `zero_copy(unsafe)`.

https://github.com/orca-so/whirlpools/pull/134
https://github.com/otter-sec/anchor/pull/2330

## Why we use `packed` for `PreparedSwap`
Most of the space in `PreparedSwap` is occupied by `PendingTickUpdate`.

With `repr(packed)`, a `PendingTickUpdate` is 37 bytes. However, if we satisfy alignment requirements and allow padding, it grows to 48 bytes (+11 bytes per entry).

Since there are 264 `PendingTickUpdate` entries, that means:

11 x 264 = 2,904 bytes

of additional space would be consumed purely by padding.

Given that Orca pays the rent and initializes the account, wasting ~3 KB is tolerable. The bigger issue is that accounts larger than 10 KB become more cumbersome to initialize. We would either need to make it a non-PDA account and use a separate create_account + initialization flow, or split initialization into initialize and extend instructions.

Since `TickArray` already uses `zero_copy(unsafe)` (what we actually need is `repr(packed)`), there was little motivation to avoid `zero_copy(unsafe)` here. As a result, I chose to keep the account as a ~10 KB PDA that can be created with a single initialize instruction.

If we ever need to grow beyond 10 KB in the future, we can introduce an extend instruction. Preserving that upgrade path is one of the reasons the `version` field exists.

WDYT ?

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: I agree with the verbosity you pointed out in the review. I was honestly tired of writing it that way too. 😂

That said, the reason I avoided the more convenient form is that these structs use `repr(packed)`.

The same constraint is also why I couldn’t use `assert_eq!` in the unit test cases and instead used local variables like `read_x`.

In this particular case, the compiler did not emit any warnings 🤔, but I’m still concerned about the risk of taking references to `repr(packed)` structs or fields.

Given that these are `repr(packed)` structs, do you still think the reduced verbosity is worth it?

programs/whirlpool/src/instructions/v2/commit_swap.rs
yugure-orca: I completely agree.

Fortunately, the original implementation—where the swap loop returns updates instead of directly mutating the Whirlpool or Oracle state—ended up fitting the prepare/commit model quite well.

TickArray was the exception since it was being updated directly. Initially, I considered buffering those changes in a Vec<PendingTickUpdate>, but I felt it was better to write them directly into the PreparedSwap account rather than introducing heap allocations and additional copies.

As a result, PendingTickUpdate ended up being handled a bit differently from everything else.

legacy-sdk/whirlpool/src/instructions/v2/prepare-swap-ix.ts
yugure-orca: One of the reasons I removed `other_amount_threshold` (as mentioned in the other comment) is that I believe prepare and commit should receive as much of the same data and accounts as possible.

There are no state changes in between, so passing the same inputs twice to two separate instructions feels like the simplest model to me. Beyond avoiding duplicated definitions, this implementation is also based on the idea that prepare and commit should fundamentally operate on the same inputs.

I also think this is easier to understand than having a long list of very similar parameter structs. Instead, it becomes obvious that "prepare can be called with a subset of the accounts for commit."

That said, there is one area where I intentionally deviate from the "pass the same things" principle.

The original name for this feature was actually "on-chain quote interface". One of the benefits of introducing the on-chain quote interface is that, as long as a client can perform RPC simulations, it can obtain swap quotes even in environments where no SDK exists. For example, languages like Python or Go may not have an official SDK, but if they have a Solana SDK capable of making RPC calls, they can still gain access to the swap math.

In that context, I felt it was better not to require accounts that are not fundamentally used by prepare. That's why prepare intentionally reduces the set of required accounts rather than requiring the exact same account list as commit.

I wouldn't say I have particularly strong conviction that this approach is the right one. There's also a part of me that feels simply passing the exact same accounts to both prepare and commit may ultimately be the simpler design.

WDYT?

programs/whirlpool/src/lib.rs
jshiohaha: ah sorry, i meant for the comment to apply to all the new handlers. `initialize_prepared_swap` makes sense as an anchor instruction since it can be called whenever 👍 

and i suppose keeping `prepare_swap_v2` and `commit_swap_v2` the same as swap makes sense too, unless the on-chain routers explicitly require a more CU efficient implementation 😅

programs/whirlpool/src/state/prepared_swap.rs
jshiohaha: ok, great — thanks for walking through that. i also didn't realize this, but makes a lot of sense

> Therefore, unless this behavior changes in the future, I don't think the above sequence of transactions can all be executed within the same block.

for context, i was mostly asking in case it was possible to have 2 versions that exist simultaneously such that the error would not allow us to process an older version. in this case, smth like the following might be useful. but as you explained, this case is not possible today 👍 

```rust
match self.version {
    1 => handle_v1(...),
    // panic arm just for example
    _ => unimplemented!()
}
```

programs/whirlpool/src/state/prepared_swap.rs
jshiohaha: > I was honestly tired of writing it that way too😅

😅 

> That said, the reason I avoided the more convenient form is that these structs use repr(packed).

ah ok, my understanding was that it's ok since we are taking a reference to a packed struct where the alignment = 1. however, a problem would arise if we took reference to a field where alignment > 1 inside this. for example, if you take reference to a u128 after taking the mutable reference to PendingPostSwapUpdate, you _should_ see a compiler warning

```rust
let dst = &mut self.pending_updates.pending_post_swap_update;
// [pasted warning] 👇 
// reference to packed field is unaligned
// packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
// creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
// copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)
let compiler_should_complain = &dst.next_liquidity;
```

that being said, it's probably better to be safe than sorry, so i am onboard if you prefer to leave as is 🙂

programs/whirlpool/src/instructions/v2/prepare_swap.rs
jshiohaha: ok, thanks for confirming same-block reusability! the jito tip example is an interesting point, i hadn't thought much about it previously. i am now really curious to know if they had considered more and if they ever have issues 😅 

regarding nonce as u8 vs u16 — i do think 256 is a pretty reasonable upper bound. but out of curiosity, is there any downside to going with u16 and artificially limiting the usable range, similar to how we're doing with u8 now? i saw that nonce isn't written into any account state, so we don't consume any more bytes. i'm comparing to if we go with u8, it might be more difficult to expand later.

> As you suggested, I think it's worth documenting that we considered this attack vector and concluded that it is not materially different from locking the pools themselves. Leaving a comment to that effect would be helpful.

unfortunately, i don't think a documented will stop any ai vulnerabilities from trying to report this issue 😭  but at least we would have documentation to point them to 👍

programs/whirlpool/src/state/prepared_swap.rs
jshiohaha: ah, thank you for the historical context. that is very helpful to know. for starters, i am aligned that we should minimize risk where possible, so i'm onboard with the previous approach if that's what is best 🤝 

but since i spent some time looking into this, i thought i'd share the findings 💡 also, i'd like to know if i'm misinterpreting/misunderstanding anything 🙌 (apologies if this is what you were saying and i read incorrectly)

### Cases

(1) it seems that using unsafe is only giving us the unsafe impl Pod/unsafe impl Zeroable versus the derived implementations without unsafe. including `packed` forces a packed layout, whereas #[repr(C)] alone would force the natural alignment

(2) removing unsafe with `packed` preserves the expected alignment=1 layout but introduces the derived implementations of Pod and Zeroable. This works for  `AdaptiveFeeConstants` because all fields are compatible. This breaks with `Tick` because the `initialized: bool` field.

(3) removing **both** unsafe and packed forces the natural alignment and breaks previous layout assumptions

```rust
// 1. today's code
#[zero_copy(unsafe)]
#[repr(C, packed)]
#[derive(Default, Debug, PartialEq, Eq)]
pub struct AdaptiveFeeConstants { ... }

// 2. changing zero_copy(unsafe) -> zero_copy
#[zero_copy]
#[repr(C, packed)]
#[derive(Default, Debug, PartialEq, Eq)]
pub struct AdaptiveFeeConstants { ... }

// 3. removing `packed` from representation, using zero_copy without unsafe
#[zero_copy]
#[repr(C)]
#[derive(Default, Debug, PartialEq, Eq)]
pub struct AdaptiveFeeConstants { ... }
```

i added a unit test for this to sanity check this was actually the case. it passes for (1) and (2) and doesn't even compile on 3 with the compiler error that follows

### Test

```rust
#[cfg(test)]
mod layout_tests {
    use super::*;

    #[test]
    fn is_expected_packing() {
        assert_eq!(std::mem::align_of::<AdaptiveFeeConstants>(), 1);
        assert_eq!(
            std::mem::size_of::<AdaptiveFeeConstants>(),
            AdaptiveFeeConstants::LEN
        );
    }
}
```

### Error


```bash
error[E0080]: evaluation of constant value failed
  --> programs/whirlpool/src/state/oracle.rs:85:1
   |
85 | #[zero_copy]
   | ^^^^^^^^^^^^ the evaluated program panicked at 'derive(Pod) was applied to a type with padding', programs/whirlpool/src/state/oracle.rs:85:1
```

programs/whirlpool/src/lib.rs
jshiohaha: ohh i totally missed these — thanks for enumerating. getting a quote on-chain and then deciding whether or not to take it or not is such a new way of thinking 😅 

based on your reasoning, i think it makes sense to exclude. thinking about future vulnerability reports, the only thing i might advocate for is a comment about this for documentation? or maybe this PR comment is sufficient since we are in the public repo?  i imagine it will be more convenient to link a previous conversation/context

legacy-sdk/whirlpool/src/instructions/v2/prepare-swap-ix.ts
jshiohaha: ah, it sounds like my original question was based on an incorrect assumption that the input shapes could diverge in the future. if the idea is that they should be same, then i think your original approach makes sense a lot of sense. the only other approaches are a lot more verbose and duplicate fields defined elsewhere 🙈

programs/whirlpool/src/instructions/v2/commit_swap.rs
jshiohaha: ok, great. it sounds like we're aligned here ⚡ and i can't think of any other issues here

i didn't originally think about it, but it seems like a non-starter for this design if we had to rely on SysvarInstructions and couldn't invoke via CPI 😅

jshiohaha (approved):
i don't have any concerns beyond what we've discussed below. i'm excited to see more feedback on this path and if this can become integrated into on-chain routing 🚀

programs/whirlpool/src/lib.rs
yugure-orca: Added `// Note: there is no other_amount_threshold parameter. .....` comment to the argument list. 🫡

Once a GitHub discussion comment gets "edited" mark, it becomes less useful as evidence in the context of an Immunefi report because the original wording is no longer readily visible. Adding a code comment instead provides a more durable record that the issue was known and intentionally considered at the time.

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: Since the maximum lifetime of a `PreparedSwap` is only one slot, any `PreparedSwap` created under an older version will expire almost immediately. As a result, I don't think we need to support multiple versions simultaneously.

Also, a `prepare -> upgrade -> commit` flow can only occur either in an attack scenario or through a non-recommended usage pattern. I don't think we need to protect against errors arising from those cases.

The fact that `prepare` and `commit` can be invoked from separate transactions within the same slot is an unavoidable consequence of not being able to enforce atomic CPI usage. `prepare` and `commit` are intended to be used atomically. As long as they are used atomically, there is no opportunity for an upgrade to occur in between. 🙂

programs/whirlpool/src/instructions/v2/prepare_swap.rs
yugure-orca: With `MAX_PREPARED_SWAP_NONCE` in place, I don't think there is any meaningful downside to using `u16`.

Also, as you noted, we do not store either the nonce or the bump in the `PreparedSwap` account. I don't really envision using this PDA as a signer, so there is no need to persist those values. 🙂

Since I don't see any significant drawback to using `u16`, I'll go ahead and change the nonce type to u16.

programs/whirlpool/src/lib.rs
yugure-orca: CU consumption for CLMM swaps is always a concern, so from the perspective of a prepare/commit user, reducing CU usage is definitely desirable.

The motivation here was that a straightforward quote + swap_v2 call effectively doubles the CU cost, which is hard to justify. What I wanted to evaluate was how much overhead prepare/commit introduces relative to a plain swap_v2.

The result is roughly a 15% overhead, which is significantly better than a 2x increase.

Since the staging pipeline already contains the Pinocchio-version swap, prepare/commit should benefit from those improvements as well when that version is eventually released.

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: Thanks for the investigation!
I think  you're right!

`PendingTickUpdate` only contains `Pod` and `Zeroable` fields, so I changed it to use `#[zero_copy]`.

I also expected `PreparedSwapPendingUpdates` to be eligible for `#[zero_copy]`, but it turns out `#[zero_copy(unsafe)]` does not generate the `unsafe impl`s for `Pod` and `Zeroable`. Those `unsafe impl`s are only generated for `#[account(zero_copy(unsafe))]`.

It also looks like the `unsafe` flag on `Oracle` and `AFConsts`/`AFVariables` could be removed. However, since that can be done in a separate PR, I'd prefer not to include those changes in this PR.

programs/whirlpool/src/state/prepared_swap.rs
yugure-orca: Thanks!
I agree with your point. After thinking it through myself, I came to the same conclusion. 😁
I simplified the code by introducing a shorter pending variable.

programs/whirlpool/src/instructions/v2/reposition_liquidity_v2.rs
```diff
@@ -7,7 +7,7 @@ use crate::state::*;
 #[derive(Accounts)]
 pub struct RepositionLiquidityV2<'info> {
     #[account(mut)]
-    pub whirlpool: Account<'info, Whirlpool>,
+    pub whirlpool: Box<Account<'info, Whirlpool>>,
```
yugure-orca, line 10: Just to remove the following warning.
```
Error: Function _ZN205_$LT$whirlpool..instructions..v2..reposition_liquidity_v2..RepositionLiquidityV2$u20$as$u20$anchor_lang..Accounts$LT$whirlpool..instructions..v2..reposition_liquidity_v2..RepositionLiquidityV2Bumps$GT$$GT$12try_accounts17hb283dde67bf6a7adE Stack offset of 4104 exceeded max offset of 4096 by 8 bytes, please minimize large stack variables. Estimated function frame size: 4224 bytes. Exceeding the maximum stack offset may cause undefined behavior during execution.
```

This warning is safe because reposition_liquidity_v2 is handled by pinocchio-based implementation. Anchor definition is used just for IDL generation. So this warning should be safe, but it is still nice to remove this warning.🙂

programs/whirlpool/src/state/prepared_swap.rs
```diff
@@ -895,9 +940,31 @@ mod data_layout_tests {
...
     fn test_len_constant() {
+        assert_eq!(PendingPostSwapUpdate::LEN, 177);
+        assert_eq!(PendingTickUpdate::LEN, 37);
+        assert_eq!(PreparedSwapPrecondition::LEN, 103);
+        assert_eq!(PreparedSwapPendingUpdates::LEN, 9947);
         assert_eq!(PreparedSwap::LEN, 1024 * 10); // 10KB
     }
 
+    #[test]
+    fn test_struct_size() {
```
jshiohaha, line 1254: nice 🔥

wjthieme (approved):

PR #1315, opened 2026-06-12
Refactor docs to only include typedocs and homepage

crnorthc:
**Title**
  Restructure the docs site into a reference-only landing page hosted on the default GitHub Pages domain

  **Details**
  The prose guides, tutorials, and conceptual documentation have moved to docs.orca.so. This PR strips the Docusaurus site down to its remaining responsibility — hosting the auto-generated SDK API references — and repoints it from the `dev.orca.so` custom domain to the default GitHub Pages domain.

  Changes:
  - **Single landing page.** Removed all prose pages under `docs/whirlpool/docs/` and replaced them with one homepage
  (`01-Welcome.mdx`) that links out to the three generated references: TypeScript SDK (Kit), TypeScript SDK (Legacy), and Rust SDK..
  - **Default GitHub Pages domain.** Switched `url`/`baseUrl` to `https://orca-so.github.io` + `/whirlpools/` (required for the project-page subpath so generated asset/link paths resolve), prefixed the SDK-reference nav `href`s with `/whirlpools/`, removed `static/CNAME`, and updated the version-check URL in `scripts/deploy-docs`.
  - **No sidebar.** Set `sidebarPath: false` so the single landing page renders full-width instead of auto-generating a sidebar.
  - **Cleanup.** Deleted the now-orphaned prose images under `static/img/`.
  - **Unchanged pipeline.** The existing build/deploy flow is preserved — the Docusaurus build still merges the TypeDoc (`../ts/dist`, `../legacy/dist`) and rustdoc (`../rust/dist`) output via `staticDirectories`, and `deploy-docs` still pushes to the `docs` branch via `gh-pages`.
  - **Changeset.** Added a `patch` changeset for `@orca-so/whirlpools-docs`; the version bump is what allows `deploy docs` to publish the new build.

### Preview of new page
Screenshot 2026-06-12 at 2 01 48 PM

PR #1318, opened 2026-06-16
Crnorthc/fix publish

crnorthc:
**Title**
Fix docs deploy step failing in the Publish workflow

**Details**
The `deploy` step of the Publish workflow was failing in the
`@orca-so/whirlpools-docs` workspace with exit code 1, which in turn broke the
topological run for every workspace that depends on it
(`can't satisfy the dependency graph`).

Two issues, both surfaced by #1315 (the docs site moving to
`orca-so.github.io/whirlpools` and away from `dev.orca.so`):

1. **`scripts/deploy-docs` aborted silently.** The script scrapes the published
   site for a `<meta itemprop="version">` tag to decide whether to redeploy.
   When `grep` found no match it exited with status 1, and under
   `set -eo pipefail` that killed the whole script before it printed anything or
   compared versions — hence the bare `exit code 1`.

2. **The version meta tag wasn't in the served HTML.** It was rendered in the
   React body of `Root.js`, so it never reached the static `<head>`, and
   react-helmet renders it as `<meta data-rh="true" itemprop="version" ...>`,
   which the script's anchored `<meta itemprop=...` pattern would not match
   anyway.

Changes:
- `scripts/deploy-docs`: add `|| true` so a no-match no longer aborts the
  script (an empty value is treated as version 0, falling through to publish),
  and loosen the grep pattern to `itemprop="version" content="..."` so it
  matches regardless of attribute ordering.
- `docs/whirlpool/src/theme/Root.js`: emit the version meta tag via
  `@docusaurus/Head` so it lands in the static HTML `<head>`.

### DIFF
--- a/.changeset/fix-docs-deploy-version-probe.md
+++ b/.changeset/fix-docs-deploy-version-probe.md
@@ -1,0 +1,5 @@
+---
+"@orca-so/whirlpools-docs": patch
+---
+
+Fix the docs deploy step failing in CI. The version probe in `scripts/deploy-docs` aborted under `set -eo pipefail` when grep found no match, and the `<meta itemprop="version">` tag was no longer reaching the served HTML. The tag is now emitted via `<Head>` and the probe tolerates a missing match.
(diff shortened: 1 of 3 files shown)


jshiohaha (approved):
let's try it 🚢

PR #1322, opened 2026-06-28
remove hidden cu margin minimum in transaction builder

jshiohaha:
## Motivation

We use TransactionBuilder in coralreef. In these transactions, it's often the case that we are grossly overestimating CU consumption due to this maximum. Maybe there's not as big of an impact on normal transactions, but I think there's a compounding effect when used with Jito bundles.

Unless there's a good reason to keep, I'd like to remove this in favor of only applying the requested margin. Without this, it's annoying for the caller to have to go back and manually manipulate the values in the CU budget instruction.

### DIFF
--- a/.changeset/three-cities-sing.md
+++ b/.changeset/three-cities-sing.md
@@ -1,0 +1,5 @@
+---
+"@orca-so/common-sdk": patch
+---
+
+Remove minimum CU margin when estimating compute units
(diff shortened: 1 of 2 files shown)


yugure-orca (approved):
nit: WDYT about "maximum" in the PR title ? Indeed "Math.max" is used, but I think this pattern is used to ensure "minimum" value (In this case, `100000` is minimum value).

In any case, I understand the motivation behind the change. The minimum margin can become non-negligible when multiple transactions are combined.

jshiohaha:
i used "minimum" initially, but i changed to "maximum" because i second guessed myself. i switched back to minimum since we are changing the minimum amount of margin applied to the CU budget — good catch 👍

PR #1323, opened 2026-06-30
update changelog dev deps

jshiohaha:
# Motivation

Publish action is consistently failing on changeset step, so trying to bump the relevant packages to see if that helps

CleanShot 2026-06-30 at 12 53 09@2x


defnotzed (approved):

yugure-orca (approved):
let's go!

PR #1324, opened 2026-07-05
docs(ts-sdk): fix JSDoc for calculateMinimumBalanceForRentExemption

### LINKED ISSUE #1319: JSDoc for calculateMinimumBalanceForRentExemption does not match its parameters
In `ts-sdk/whirlpool/src/sysvar.ts` 

The documentation describes the first parameter as:
```typescript
@param {Rpc} rpc - The Solana RPC client to fetch sysvar rent data.
```

But the actual signature is:
```typescript
export function calculateMinimumBalanceForRentExemption(
  rent: SysvarRent,
  dataSize: number,
): bigint
```

Update the JSDoc to:
```typescript
@param {SysvarRent} rent - The sysvar rent data used to compute the minimum balance.
```

Dhruv-mavani:
## Summary

This PR fixes the JSDoc for `calculateMinimumBalanceForRentExemption` so that it matches the actual function signature.

### Changes

- Updated the first parameter type from `Rpc` to `SysvarRent`
- Updated the parameter description accordingly

Closes #1319.

### DIFF
--- a/ts-sdk/whirlpool/src/sysvar.ts
+++ b/ts-sdk/whirlpool/src/sysvar.ts
@@ -8,7 +8,7 @@ const ACCOUNT_STORAGE_OVERHEAD = 128;
 /**
  * Calculates the minimum balance required for rent exemption for a given account size.
  *
- * @param {Rpc} rpc - The Solana RPC client to fetch sysvar rent data.
+ * @param {SysvarRent} rent - The connected cluster's Rent sysvar, obtained via `fetchSysvarRent(rpc)`.
  * @param {number} dataSize - The size of the account data in bytes.
  * @returns {bigint} The minimum balance required for rent exemption in lamports.
  */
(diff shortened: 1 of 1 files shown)


Dhruv-mavani:
@jshiohaha @wjthieme @yugure-orca....*Reminder*

ts-sdk/whirlpool/src/sysvar.ts
```diff
@@ -8,7 +8,7 @@ const ACCOUNT_STORAGE_OVERHEAD = 128;
 /**
  * Calculates the minimum balance required for rent exemption for a given account size.
  *
- * @param {Rpc} rpc - The Solana RPC client to fetch sysvar rent data.
+ * @param {SysvarRent} rent - The sysvar rent data used to compute the minimum balance.
```
jshiohaha: thanks for the proposed change! wdyt about smth like

```ts
@param {SysvarRent} rent - The connected cluster's Rent sysvar, obtained via `fetchSysvarRent(rpc)`.
```

Dhruv-mavani: yes i mean that makes sense.....I'll update the description to mention where it comes from.

PR #1326, opened 2026-07-08
Fix: u256 div edge case

yugure-orca:
Note: If this occurs, the program panics, so no data inconsistency can occur.

### DIFF
--- a/programs/whirlpool/src/math/u256_math.rs
+++ b/programs/whirlpool/src/math/u256_math.rs
@@ -573,12 +573,8 @@ fn div_loop(
             k = t >> U64_RESOLUTION;
         }
 
-        let new_carry = dividend
-            .get_word_u128(index + num_divisor_words)
-            .wrapping_add(k)
-            .lo();
         if use_carry {
-            *dividend_carry_space = new_carry
+            *dividend_carry_space = (*dividend_carry_space as u128).wrapping_add(k).lo();
         } else {
             dividend.update_word(
                 index + num_divisor_words,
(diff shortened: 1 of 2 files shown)


programs/whirlpool/src/math/u256_math.rs
```diff
@@ -1066,6 +1062,47 @@ mod test_div {
         let divisor = U256Muldiv::new(0, 0);
         let _ = dividend.div(divisor, true);
     }
+
+    #[test]
+    fn test_div_8_full_width_dividend_less_than_divisor_returns_zero() {
```
jshiohaha, line 1067: nice, confirmed these new test cases experience the expected oob panic on the old logic 👍 

```bash
index out of bounds: the len is 4 but the index is 4
```

programs/whirlpool/src/math/u256_math.rs
```diff
@@ -573,12 +573,8 @@ fn div_loop(
             k = t >> U64_RESOLUTION;
         }
 
-        let new_carry = dividend
-            .get_word_u128(index + num_divisor_words)
-            .wrapping_add(k)
-            .lo();
         if use_carry {
-            *dividend_carry_space = new_carry
+            *dividend_carry_space = (*dividend_carry_space as u128).wrapping_add(k).lo();
```
jshiohaha, line 577: looks like the update here aligns with the recommended fix — slightly different syntax, but logically equivalent

`dividend_carry_space` is only referenced when `use_carry=true`, which represents when we are at the 4th word boundary and prevents us from the previous oob issue

jshiohaha (approved):
🚢

wjthieme (approved):

PR #1362, opened 2026-08-10
Try to fix npm trusted publishing in the release workflow

jshiohaha:
Publishing has 404'd since #1332 switched to OIDC -- npm masks auth failures as 404 for scoped packages. Node 22 bundles npm 10.x, which has no OIDC support (needs >= 11.5.1), and setup-node's `registry-url` writes an empty `_authToken` that makes npm skip the OIDC exchange (actions/setup-node#1551).

Moves to Node 24, drops `registry-url`, and checks the npm version and OIDC token in deploy-npm so a regression names its cause instead of 404ing. The npm version is worth asserting because it is now implicit: nothing in the repo pins it, it is whatever the resolved Node 24.x happens to bundle.

Also repoints common-sdk's repository field at orca-so/whirlpools. It still referenced orca-so/orca-sdks, the package's old home, archived in March 2025 when the SDKs moved here -- the field was simply never updated. It matters now because trusted publishing generates provenance automatically, and that requires the repository field to match the publishing repo case-sensitively, so the stale value would fail this package on its next release.


.github/workflows/publish.yml
```diff
@@ -27,9 +27,10 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v7
         with:
-          node-version: 22
+          node-version: 24
```
jshiohaha, line 30: from claude. originally a code comment but pulled out to PR comment

```text
Node 24 bundles npm 11.x. Trusted publishing (OIDC) needs npm >= 11.5.1, and Node 22
ships npm 10.x, which has no OIDC support at all -- it falls back to publishing
anonymously, which the registry rejects as a misleading "404 Not Found".
scripts/deploy-npm enforces the minimum version before publishing.
https://docs.npmjs.com/trusted-publishers/
```

wjthieme (approved):

.github/workflows/publish.yml
```diff
@@ -27,9 +27,10 @@ jobs:
...
           cache: 'yarn'
-          registry-url: 'https://registry.npmjs.org'
```
jshiohaha, line 32: from claude. originally a code comment but pulled out to a PR comment.

```text
No `registry-url` on purpose. Setting it makes setup-node write
`//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}` into an .npmrc. We publish via
OIDC and set no NODE_AUTH_TOKEN, so that placeholder expands to an empty token, npm
treats auth as already configured, never performs the OIDC exchange, and the registry
rejects the anonymous PUT -- again as a 404.
https://github.com/actions/setup-node/issues/1551
Nothing here needs the input: npmjs.com is npm's default registry, we have no private
dependencies to authenticate, and Yarn reads .yarnrc.yml rather than .npmrc.
```

PR #1366, opened 2026-08-13
Prepare for rent reduction starting with Agave 4.2

yugure-orca:
The following behavior can cause issues during the phased Rent reduction or if a fallback to lamports_per_byte = 6960 occurs:

- DynamicTickArray is variable-length.
- Position transfers lamports to and from DynamicTickArray.

To address this, we add the following safeguards:

- Fix the amount of Rent reserved for one tick to the current amount.
- When initializing DynamicTickArray and Position accounts, always reserve Rent based on the current amount.

Note: The size of a Position account does not change, but when it sends lamports to a DynamicTickArray and its balance decreases, its rent-exempt status is re-evaluated using the latest Rent parameters. As long as the Position account itself was funded assuming lamports_per_byte = 6960, it will remain safe even if that re-evaluation occurs.

### References:
- [Agave v4.2 updates](https://www.helius.dev/blog/agave-v4-2)
- [SIMD-0437 Rent reduction](https://github.com/anza-xyz/agave/pull/10126)
- [SIMD-0438 Rent reduction fallback](https://github.com/anza-xyz/agave/pull/10127)
- [SIMD-0392 Rent increase adaptions](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0392-rent-increase-adaptations.md)
- [SIMD-0194 Deprecate Rent exemption threshold](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0194-deprecate-rent-exemption-threshold.md)

### DIFF
--- a/programs/whirlpool/src/manager/tick_array_manager.rs
+++ b/programs/whirlpool/src/manager/tick_array_manager.rs
@@ -3,7 +3,9 @@ use solana_program::{program::invoke, system_instruction};
 
 use crate::errors::ErrorCode;
 use crate::state::{DynamicTick, Position, PositionUpdate, Tick, TickUpdate};
-const TICK_INITIALIZATION_SIZE: usize =
+use crate::util::{get_position_minimum_rent_amount, get_tick_rent_amount};
+
+pub const TICK_INITIALIZATION_SIZE: usize =
     DynamicTick::INITIALIZED_LEN - DynamicTick::UNINITIALIZED_LEN;
 
 #[derive(Default, Debug, PartialEq)]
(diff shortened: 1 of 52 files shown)


programs/whirlpool/src/manager/tick_array_manager.rs
```diff
@@ -66,14 +66,31 @@ pub fn get_tick_rent_amount() -> Result<u64> {
...
         // - 3480u64 = 1_000_000_000 / 100 * 365 / (1024 * 1024)
         // - 0x40_00_00_00_00_00_00_00u64 = 2.0f64 in u64 bits representation
-        (112, 3480u64, 0x40_00_00_00_00_00_00_00u64) => Ok(779520u64), // Solana
+        (112, 3480u64, 0x40_00_00_00_00_00_00_00u64) => Ok(779520u64),
         _ => {
-            let amount = ((TICK_INITIALIZATION_SIZE as u64 * rent.lamports_per_byte_year) as f64
-                * rent.exemption_threshold)
-                .ceil() as u64;
-            Ok(amount)
+            unreachable!(
```
yugure-orca: This code is needed for fallback and Eclipse network support.
We've already deprecated Eclipse network support, so this fallback is no longer needed.
Also I believe that `lamports_per_byte_year` will not exceed the current value (6960) because the ecosystem is trying to reduce it...

programs/whirlpool/src/util/token_2022.rs
```diff
@@ -490,14 +491,20 @@ pub fn initialize_vault_token_account<'info>(
         },
     )?;
 
-    let lamports = Rent::get()?.minimum_balance(space as usize);
+    let lamports = if spl_token::native_mint::check_id(&vault_mint_account_info.key()) {
```
yugure-orca, line 494: Strictly speaking, WSOL vaults are also at risk.
In short, for an x/SOL pool, the vault could end up with an insufficient WSOL balance. (Although this can be easily fixed by sending a small amount of SOL to the vault.)

- Initialize the WSOL vault with the lower rent (lamports = Rs, WSOL amount = 0)
- Transfer 1 SOL & sync native (lamports = 1 SOL + Rs, WSOL amount = 1 SOL)
- Rent fallback occurs (rent increases from Rs to Rb)
- Sync native (lamports = 1 SOL + Rs, WSOL amount = 1 SOL - (Rb - Rs))

Other senario

- Initialize the WSOL vault with the lower rent (lamports = Rs, WSOL amount = 0)
- Transfer 1 SOL & sync native (lamports = 1 SOL + Rs, WSOL amount = 1 SOL)
- Rent fallback occurs (rent increases from Rs to Rb)
- Try to send 1 SOL from the vault (post lamports = Rs < Rb) -> failed (account size is not changed, but WSOL account update lamports value...)

yugure-orca: Note: Whirlpool program reject Token-2022's native mint, so it is sufficient to check Token's native mint here.

programs/whirlpool/src/util/rent.rs
```diff
@@ -0,0 +1,95 @@
...
+pub fn get_position_minimum_rent_amount() -> Result<u64> {
+    assert_expected_rent_config()?;
+    Ok(2_394_240)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const ACCOUNT_STORAGE_OVERHEAD: u64 = 128;
```
jshiohaha, line 68: nit: maybe overkill, but could consider permalink to original sdk source for context https://github.com/anza-xyz/solana-sdk/blob/5190ff456079d17b64669bcb5eeac48dd595b91e/rent/src/lib.rs#L86

programs/whirlpool/src/manager/tick_array_manager.rs
```diff
@@ -146,12 +110,21 @@ pub fn calculate_modify_tick_array(
...
     funder: &Signer<'info>,
     position: &Account<'info, Position>,
     system_program: &Program<'info, System>,
 ) -> Result<()> {
-    let rent_amount = get_tick_rent_amount()? * 2;
+    let rent_with_reduction_fallback_margin = get_position_minimum_rent_amount()?;
+    let reduction_fallback_margin =
+        rent_with_reduction_fallback_margin.saturating_sub(position.get_lamports());
+
+    let rent_for_ticks = get_tick_rent_amount()? * 2;
```
jshiohaha, line 124: extreme nit: i think it's impossible to overflow here, but a small departure from checked math operations in the rest of this file

programs/whirlpool/src/util/token_2022.rs
jshiohaha: nice catch!

jshiohaha (approved):
nice work!! changes look good to me 🚢 

i also thought about position (2022) mint since that is dynamically allocated, but there's no resize after initialization, so it should be immune to rent fluctuations.

the only other thought i had (and it might be overkill), but since we're using LiteSVM, it maybe possible to add integration test(s) demonstrating we correctly preserve lamports even after modeling a rent reduction -> fallback process

yugure-orca:
verifiable build point

wjthieme (approved):

PR #1392, opened 2026-09-09
Prepare for SIMD-0460 Virtual Address Space Adjustments

yugure-orca:
DynamicTickArray was accessing the +10 KB account data growth region, but no meaningful writes were performed there. Those writes were effectively ignored, so this did not cause any issues.

Once SIMD-0460 is enabled, accesses beyond the account's current data size will be detected more strictly and result in errors. Therefore, we are updating the logic to avoid accessing data beyond the current account size.

Note: This change is intended to prepare for the behavior change introduced by SIMD-0460. The existing behavior is not considered a bug prior to the activation of SIMD-0460.

### Testing
The feature ID is `7VgiehxNxu53KdxgLspGQY8myE6f7UokaWa4jsGcaSz`.

Enabling this feature in LiteSVM results in `Failed to create program execution environment`, so there may be a compatibility issue with LiteSVM or node-litesvm.

Therefore, I ran the tests using `solana-test-validator` instead. I confirmed that cases that result in `InvalidRealloc` with the existing program complete successfully with the updated program.

I also strengthened the unit tests to detect and fail if a write goes beyond the length of an account's data.


programs/whirlpool/src/pinocchio/errors.rs
```diff
@@ -5,6 +5,7 @@ use pinocchio::program_error::ProgramError as PinocchioProgramError;
 pub use crate::errors::ErrorCode as WhirlpoolErrorCode;
 pub use anchor_lang::error::ErrorCode as AnchorErrorCode;
 
+#[derive(Debug)]
```
yugure-orca, line 8: needed to use "unwrap" in **test** context.

programs/whirlpool/src/pinocchio/state/whirlpool/tick_array/dynamic_tick_array.rs
```diff
@@ -124,3 +141,556 @@ impl MemoryMappedDynamicTickArray {
...
+            let mut account_len =
+                STATIC_FIELD_LEN + DynamicTick::UNINITIALIZED_LEN * TICK_ARRAY_SIZE_USIZE;
+            for i in 0..TICK_ARRAY_SIZE_USIZE {
+                let offset = initialize_order[i];
+                let tick_index = offset_to_tick_index(offset, start_tick_index, tick_spacing);
+
+                account_len += crate::state::DynamicTickData::LEN;
+
+                // dirty write to non-allocated buf range
+                buf[account_len..].fill(rand_u8_for_initialize);
```
yugure-orca, line 453: "rotate" overwrites this  area.
If the updated logic overwrites this area, it will be detected by the following assertion.

programs/whirlpool/src/pinocchio/state/whirlpool/tick_array/dynamic_tick_array.rs
```diff
@@ -124,3 +141,556 @@ impl MemoryMappedDynamicTickArray {
...
+            let mut uninitialized = 0;
+            let mut bitmap = ALL_INITIALIZED_BITMAP;
+            let mut account_len =
+                STATIC_FIELD_LEN + DynamicTick::INITIALIZED_LEN * TICK_ARRAY_SIZE_USIZE;
+            for i in 0..TICK_ARRAY_SIZE_USIZE {
+                let offset = uninitialize_order[i];
+                let tick_index = offset_to_tick_index(offset, start_tick_index, tick_spacing);
+
+                // dirty write to non-allocated buf range
+                buf[account_len..].fill(rand_u8_for_uninitialize);
```
yugure-orca, line 533: "rotate" overwrites this  area.
If the updated logic overwrites this area, it will be detected by the following assertion.

wjthieme (approved):
Send it!

PR #1396, opened 2026-09-11
Improve checkpoint update

yugure-orca:


programs/whirlpool/src/manager/position_manager.rs
```diff
@@ -52,6 +66,45 @@ pub fn next_position_modify_liquidity_update(
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CheckpointUpdateMode {
+    // Advances the checkpoint only for growth converted into owed amount.
+    Partial,
+    // Advances the checkpoint fully to growth_inside.
+    // Required before changing the position liquidity.
+    Full,
+}
+
+pub fn next_owed_delta_and_checkpoint(
```
wjthieme, line 78: nit: since these two functions (for rewards and for fees) are pretty much identical. Maybe give them different names so that it is more explicit that they are separate

wjthieme (approved):
LGTM

programs/whirlpool/src/manager/position_manager.rs
yugure-orca: We agreed that it makes sense to keep it as a single function since the logic (for fee & reward) is identical.

